PR #102, opened 2026-06-24
Verify eject fidelity and harden partial-snapshot revert

troyjr4103:
## Summary

Adds an end-to-end conviction suite for the "leave anytime" contract and closes a
real fidelity gap it surfaced in `kin eject --revert-files`.

The new tests drive the real lifecycle through the public CLI and the daemon
runtime — `kin init` → edit through the reconcile path → `kin commit` →
`kin eject` / `kin git export` — and verify the result with **no Kin tooling
present** (stock `git` and `rustc` only).

## Fidelity gap found and fixed

`kin eject --revert-files` read the snapshot manifest's `file_count` only for a
status message. It then copied whatever files happened to be in `.kin/snapshot/`
and unconditionally deleted `.kin/`. A partial or corrupt snapshot therefore
restored fewer files than promised, reported success, and destroyed the graph —
silent data loss behind the core adoption-trust guarantee.

The revert path now verifies snapshot completeness **before any mutation**: if the
snapshot holds fewer files than the manifest declares, it fails loudly and changes
nothing — the working tree is untouched, `.kin/` is preserved, and no backup is
written.

## Tests (`crates/kin-cli/tests/eject_fidelity.rs`)

- `eject_revert_files_is_byte_faithful_and_leaves_git_intact` — after
  init/edit/commit, `kin eject --revert-files` restores every file byte-for-byte
  to its pre-init state; `.git` and Git HEAD are unchanged; the tracked tree is
  clean; the pre-eject edit is recoverable from the mandatory backup; and the
  restored tree builds and its tests pass under a plain toolchain.
- `git_export_round_trips_to_plain_git` — `kin git export` yields a plain Git
  repo; a stock `git clone` is fully usable and builds/tests green with no Kin
  present. The semantic graph (entities, relations, reviews, provenance, work,
  verification, sessions, intents, per-change spec/evidence/risk metadata) is
  documented as intentionally out of scope — Git has no representation for it.
- `eject_fails_loud_on_partial_snapshot` — a corrupt snapshot makes the revert
  exit non-zero with an integrity error and zero side effects.

A unit test (`revert_fails_loud_on_incomplete_snapshot`) covers the same guard
without the daemon.

## Docs

Corrected the README and quickstart, which claimed `kin eject [--force]` restores
files. The default eject leaves working files untouched; `--revert-files` is what
restores them. Documented exactly what eject restores and what it never touches
(Git history was never Kin's to restore).

## Verification

- `cargo test -p kin-cli --lib eject` — 29 passed (new + existing guards green).
- `cargo test -p kin-cli --test eject_fidelity` — 3 passed.
- `cargo fmt` clean; `cargo clippy` adds no warnings on the changed files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ In Kin, the graph is the source of authority. The filesystem is a derived projec
 - **Semantic VCS**: Version control at the level of entities and relations instead of lines and files.
 - **AI-native context packs**: Build structured context packs containing a target entity plus its caller/callee neighborhood in a single call (`kin context`).
 - **Trace & dataflow**: Trace call chains and dataflow dependencies in one step instead of looping over file reads (`kin trace`, `kin trace-data-flow`).
-- **Ejectable brownfield compatibility**: Works alongside Git and traditional toolchains. `kin eject` removes Kin and restores files to their pre-init state, so there is no data lock-in.
+- **Ejectable brownfield compatibility**: Works alongside Git and traditional toolchains. `kin eject` removes Kin and leaves your working files untouched; `kin eject --revert-files` additionally restores files to their byte-for-byte pre-init state. Git history is never touched, so there is no data lock-in.
 
 ---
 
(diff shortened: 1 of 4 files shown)


troyjr4103:
Superseded — FIR-892 landed via #103 (lane-daemon's eject_round_trip.rs conviction suite: real binary, byte-exact round-trip, fail-loud corruption paths, 8+5 tests, merged + verified). #102 is a parallel eject-conviction effort (eject_fidelity.rs) from another session — same property proven, so closing as redundant now that the ticket is Done. Nice convergence; #103 is canonical.

PR #109, opened 2026-06-24
Correct README eject wording and add git export round-trip test

troyjr4103:
## Summary

Closes two residual gaps in the "you can always leave" conviction that the
eject-revert work (already on main) did not cover.

### 1. README eject wording (correctness)

README's eject bullets claimed `kin eject [--force]` "restores files to their
pre-init state." That does not match the mechanism:

- default `kin eject` removes Kin and leaves working files **untouched**;
- `kin eject --revert-files` is what restores files to their byte-for-byte
  pre-init state;
- `--force` is a deprecated no-op.

Both the feature bullet and the everyday-command list are corrected so the
public claim matches behavior. (quickstart.md was already corrected separately.)

### 2. `kin git export` → plain-git round-trip (coverage)

Adds `crates/kin-cli/tests/git_export_roundtrip.rs`, an end-to-end test of the
migration-out path through the real CLI + daemon runtime: `kin init` → edit
through the reconcile path → `kin commit` → `kin git export`. It then proves,
with no Kin tooling present, that:

- the export is a plain Git repository (`git rev-list` / `git log` / `git show`
  work; no `.kin/` inside);
- the exact committed file bytes and the commit message round-trip;
- a stock `git clone` of the export builds and its tests pass under a plain
  `rustc` toolchain.

The semantic graph (entities, relations, reviews, provenance, work items,
annotations, verification links, sessions, intents, and per-change
spec/evidence/risk metadata) is documented as intentionally **out of scope** —
Git has no representation for it. Code and Git history travel; the semantic
layer is what Kin adds.

## Verification

- `cargo test -p kin-cli --test git_export_roundtrip` — 1 passed.
- `cargo fmt` clean; `cargo clippy` adds no warnings on the changed files.
- The export test was previously run 3/3 plus an 8x stability loop on the
  sibling eject-fidelity branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ In Kin, the graph is the source of authority. The filesystem is a derived projec
 - **Semantic VCS**: Version control at the level of entities and relations instead of lines and files.
 - **AI-native context packs**: Build structured context packs containing a target entity plus its caller/callee neighborhood in a single call (`kin context`).
 - **Trace & dataflow**: Trace call chains and dataflow dependencies in one step instead of looping over file reads (`kin trace`, `kin trace-data-flow`).
-- **Ejectable brownfield compatibility**: Works alongside Git and traditional toolchains. `kin eject` removes Kin and restores files to their pre-init state, so there is no data lock-in.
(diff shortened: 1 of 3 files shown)


troyjr4103:
Superseded by #112 — rebased onto clean main after #110 and #111 landed.

PR #185, opened 2026-07-02
fix(git): deterministic history-import selection and order at commit-time ties

troyjr4103:
import_full() walked history by commit time and truncated with take(max_commits). Equal-timestamp commits sit in a process-dependent order in that walk, so when a timestamp tie straddled the truncation boundary two preps of the same repository selected different commit SETS — and always fed a different commit ORDER into semantic enrichment, whose incremental diff assigns entity and relation deltas order-sensitively. The same entity then lands under a different imported change id across preps, producing disjoint revision-id sets in otherwise identical prepared graphs (the standing FIR-814/FIR-932 retrieval-variance root). The co-change miner had this exact bug fixed in #87; the history-import path did not.

Selection now collects the walk as (commit_time, oid) pairs and applies the existing total order — time descending, then oid — before truncating, reusing the cochange helper (promoted to crate visibility; it already carries a tie-boundary unit test). Behavior is identical whenever timestamps are distinct; only the equal-timestamp tie boundary becomes content-deterministic. The test-only serial reference implementation is updated in lockstep so the parallel/serial byte-identity gate still holds.

New regression test builds an equal-timestamp repository and asserts the truncated import selects the oid-deterministic subset with stable imported change ids across two preps; verified failing on the pre-fix truncation. kin-git suite 27/0; fmt and clippy clean.

End-to-end confirmation (two-prep freeze byte-identity on a real corpus) is queued behind the current GPU window and tracked on FIR-932.

### DIFF
--- a/crates/kin-cli/src/commands/bench_meta.rs
+++ b/crates/kin-cli/src/commands/bench_meta.rs
@@ -458,6 +458,8 @@ mod tests {
             .unwrap();
         Command::new("git")
             .args(["commit", "-m", "init"])
+            .env("GIT_AUTHOR_DATE", "1000000000 +0000")
+            .env("GIT_COMMITTER_DATE", "1000000000 +0000")
             .current_dir(repo.path())
             .output()
             .unwrap();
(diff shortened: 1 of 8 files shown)


troyjr4103:
CI exposed two follow-on issues, both fixed in the amended commit: the order-asserting fixtures relied on wall-clock commit seconds (equal timestamps on fast runners legitimately flip the tie order the fix makes deterministic — fixtures now set explicit dates, and the multi-commit builder deliberately pins ALL commits to one timestamp so tie handling is exercised on every run, not only on fast machines); and the parallel mapping hit the documented gitoxide transient-miss window for loose-object repos, which the pack-index pre-warm does not cover — object lookups now retry once before failing loud as genuinely absent.

PR #238, opened 2026-07-07
feat(review): revert-history evidence channel for shadow review

troyjr4103:
## Summary
The v7 run's checkpoint-80 concentrated K's recall gap into one mechanism: revert-shaped changes read structurally as ordinary additions/removals and carry no gating signal (7 pytest FNs, plus the 52f3/e2867 class — evidence on FIR-1341/FIR-1295). This adds the channel FIR-1341 names as the primary recall lever:

- Walks a bounded window (250 changes) of the **base** ref's ancestry — evidence available at review time only; never consults the future.
- Matches head-side **additions** against window **removals**: behavior-fingerprint match = strong (restores removed content verbatim); name+kind match = weak (same surface, modified content). Matches head-side **removals** against window **additions** (revert of a recent feature).
- Findings feed the gate as ordinary warning findings (same shape as the command-effect and toolchain-surface channels). Shallow history (<25 changes) produces an explicit `revert_history_shallow` evidence gap instead of silent certification.
- Review-time computation from the change DAG + revision store. The durable follow-on is persisting the same evidence as mined `Reverts`/`RegressedBy` relations (needs a kin-model registry bump — deliberately not in this PR); this computation is that miner's oracle.

## Validation
- 4 new end-to-end tests through `build_shadow_report`: strong reintroduction flips needs_attention; recent-addition removal flags; fresh additions stay silent; shallow history gaps honestly. Full workspace gate green (fmt, ci.yml clippy allowlist, build --all-targets, all suites).
- **18-row artifact backtest staged** (.kin-coord/merge-trust-fir1341-backtest.sh): 10 revert-class FN targets must flip, 2 already-caught reverts must stay flagged, 6 benign controls must stay pass — runs on the v7-hydrated graphs after v7 completes; results will be posted to FIR-1341 before any merge decision.

## Review notes
**HOLD FOR FOUNDER — do not merge**: gating strength (warning/attention) is intentionally the minimal form; the backtest decides whether it stays. Window/depth constants are documented and only adjustable via further backtests. Implements FIR-1341 channel B.

PR #249, opened 2026-07-07
experiment(review): leaf-coherence gating for revert-history evidence

troyjr4103:
**DO NOT MERGE** — this is a measurement branch for the revert-history channel gating decision. It builds on top of the revert-history channel PR (`fix/fir-1341-revert-history-channel`) and exists to be backtested, not shipped. Linear: https://linear.app/firelock-ai/issue/FIR-1341

## What

Restrict body-reversion **gating** to *coherent leaf* groups. A body-reversion finding drives the shadow verdict (`revert_history`, warning) only when **two or more LEAF entities** revert to bodies from the **same** prior semantic change. Everything else stays info (`revert_history_incidental`), reported but not verdict-driving. Add/remove reintroduction matching is unchanged.

Leaf-coherence counting, decided purely from `EntityKind` + span data already on the channel's inputs (no path re-derivation):

- **Aggregate** kinds — `Module`, `File`, `Package` — never count. Their body is a rollup of their members, so they co-revert *by construction* whenever any member reverts.
- **Container** kinds — `Class`, `Interface`, `TraitDef`, `EnumDef` — count only when **no member leaf nested in the container's span** (same `file_origin`, line span contained) already explains the reversion. A class reverting because one of its methods reverted is *one* logical reversion, not two.
- **Leaf** kinds — everything else (`Function`, `Method`, `Constant`, …) — always count.

Nesting must be positively shown; a missing origin/span reads as "not nested", so a container with no locatable member still counts (conservative toward the low-FP direction only where it matters — at the exact 1-leaf + 1-container margin).

## Why

The revert channel's two matchers sit at very different points on the recall/FP frontier. Body-reversion recovers most of the revert class but fires on benign textual reversions, so the shipped channel pins it to info. A plain "≥2 entities co-revert to the same change ⇒ gate" rule was measured and regressed ~a third of clean benigns — structurally, because a `Module` entity is a content aggregate of its file, so any single-function reversion makes the module co-revert to the same change *for free*. Two-entity coherence is nearly costless. Excluding aggregates (and member-explained containers) from the count is the fix under test.

### Measured frontier (prior arms measured on real backtests; this arm is a candidate)

| Arm | Revert-class recall | Benign cost |
|---|---|---|
| matcher-1 add/remove reintroduction (gates) | 2/11 | ~zero |
| matcher-2 body reversion, always-info (shipped in #238) | 9/11 detected, 0 gate | zero (never gates) |
| snapshot-coherence (≥2 *any* entities ⇒ gate) | 9/11 | **15/42 clean benigns regressed (~36% FP)**; 2/6 adversarial controls |
| **leaf-coherence (this branch)** | *to be measured* | *to be measured* |

### Predicted effects (offline read; the backtest is the authority)

- The module + single-function co-reversion class (the 15 benign regressions) **no longer gates** — the module is aggregate, so the group has one leaf, below threshold.
- Genuine ≥2-leaf reverts to the same change **still gate**.
- Single-leaf reverts (e.g. the `1ad971ea2b` Catch2 row) **drop to info** — an accepted, documented recall cost of the leaf restriction.
- `fb99b5c66e` stays flagged (caught by add/remove matcher-1, untouched).

## Tests

`cargo test -p kin-review` → **151 passed, 0 failed** (was 146; net +5 after replacing the old always-info coherence test with the new gating semantics). New/rewritten coverage:

- `leaf_coherent_body_reversion_group_gates` — two leaf functions, same change ⇒ gates (warning, needs_attention).
- `module_plus_member_body_reversion_stays_incidental` — module + one member function ⇒ info only (the measured FP class).
- `scattered_body_reversions_to_different_changes_stay_incidental` — two leaves, different changes ⇒ info only.
- `lone_leaf_body_reversion_is_documented_recall_cost` — single leaf ⇒ info (plus the still-green `body_reversion_to_older_revision_flags`).
- `class_reverting_alone_counts_toward_coherence` / `class_reversion_explained_by_member_stays_incidental` — the Class rule pinned both ways.

`cargo fmt --all --check` clean; `cargo clippy -p kin-review` introduces no new warnings.

Staged (unrun) backtest scripts for the 18-row flip/stay/control set and the 60-row benign sweep are prepared under the coordination directory for the captain to fire under resource locks; they reuse the existing hydrated corpus graphs (this is a review-layer-only change, so graph state is identical).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Revert-history review channel: shadow review flags revert-shaped changes — an added
+  entity that restores content removed in the recent past (behavior-fingerprint match),
+  a reintroduction of a recently-removed surface (same name and kind, modified content),
+  and removals of recently-introduced entities. The evidence is temporal, read from a
+  bounded window of the base ref's ancestry at review time, and feeds the gate as
+  ordinary warning findings; when the base has too little history to scan, the report
+  carries an explicit evidence gap instead of a silent pass.
+
 ## [0.2.12] - 2026-07-06
 
 First-touch review latency: ancestor..descendant review pairs hydrate git history once instead of twice.
--- a/crates/kin-cli/src/commands/review.rs
+++ b/crates/kin-cli/src/commands/review.rs
@@ -1120,7 +1120,8 @@ fn inline_comment_severity(kind: kin_review::InlineCommentKind) -> &'static str 
         | InlineCommentKind::ToolchainSurfaceChange
         | InlineCommentKind::ConsumerFanout
         | InlineCommentKind::Renamed
-        | InlineCommentKind::AgentUnreviewed => "warning",
+        | InlineCommentKind::AgentUnreviewed
+        | InlineCommentKind::RevertHistory => "warning",
         InlineCommentKind::Added | InlineCommentKind::Removed => "info",
     }
 }
--- a/crates/kin-cli/src/commands/review.rs
+++ b/crates/kin-cli/src/commands/review.rs
@@ -1122,7 +1122,9 @@ fn inline_comment_severity(kind: kin_review::InlineCommentKind) -> &'static str 
         | InlineCommentKind::Renamed
         | InlineCommentKind::AgentUnreviewed
         | InlineCommentKind::RevertHistory => "warning",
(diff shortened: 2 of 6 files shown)


troyjr4103:
Closing as concluded experiment: leaf-coherence gating for revert-history evidence was evaluated and rejected — module co-reversion grants free coherence on 15/42 benign rows, so the gate cannot fire safely (decision record on FIR-1341, 2026-07-08; revert channel ships always-info instead). Branch tip preserved in the 2026-07-09 ref snapshot; findings are in the FIR-1341 thread.

PR #288, opened 2026-07-10
Add deterministic ranked impact output

troyjr4103:
## Summary

- add deterministic graph-native downstream impact ranking to `kin-review`
- preserve complete relation-path evidence for every ranked candidate
- use stable semantic identity based on file, name, kind, and normalized signature rather than snapshot-local entity IDs or line numbers
- fail closed when structured symbol resolution is ambiguous
- expose ranked JSON alongside the existing human-readable impact output so legacy CLI behavior remains intact
- label the score as inspection priority, not break probability

## Why

Kin needs a reproducible impact surface before it can measure predicted blast radius against observed downstream regressions. The existing impact output exposed reachability but did not provide stable ranking, evidence paths, or claim-safe identity semantics.

## Impact

Agents and benchmark adapters can consume a deterministic, evidence-preserving impact list without replacing Kin's graph authority or falling back to raw file search.

## Verification

- `cargo test -p kin-review ranked_impact` (5 passed)
- `cargo test -p kin-cli impact` (6 passed)
- `git diff --check origin/main...HEAD`

## Claim boundary

This PR adds the product surface, not a benchmark result. No accuracy, probability, or investor claim should be made until a sealed impact-vs-observed-reality cohort is run against a release that contains this code.

## Integration

Draft only. This PR has no file overlap with the current Kin queue and should be enqueued behind the existing four PRs.


troyjr4103:
Consolidated review — HOLD. Three blocking findings, each verified against origin/main; determinism audit and freeze-wave check are clean.

**Blocking**

1. **Wire-schema break under CLI/daemon version skew** (`crates/kin-cli/src/commands/impact.rs`)
`ImpactResponse` adds required fields `schema_version`, `resolution`, `query` with no `#[serde(default)]`. `DaemonClient::impact` (daemon_client.rs:523) deserializes the daemon body directly into this struct, `ensure_daemon_running` reuses an already-running daemon with no build gate, and build mismatch is warn-only unless `KIN_STRICT_BUILD_MATCH`. An upgraded CLI against a still-running pre-PR daemon (a supported state; MCP-spawned daemons idle for 30 minutes) makes plain `kin impact NAME` fail with "parse daemon impact response" instead of printing lines. The repo's wire-evolution convention for additive response fields is `#[serde(default)]` (locate.rs, context.rs). Fix: serde defaults on the three new fields; if the `--json` path needs fail-closed capability detection against old daemons, treat an absent `schema_version` as the loud error there specifically.

2. **Legacy resolution semantics silently narrowed and mis-declared** (`crates/kin-cli/src/commands/impact.rs`, `resolve_entities`)
On main, name resolution goes through kin-db `by_name_pattern` (engine/index.rs:130): case-insensitive, `*` prefix/suffix wildcards, token match, substring fallback. The new `matches.retain(|entity| entity.name == request.entity)` applies case-sensitive exact equality to the legacy text path too, so queries that resolve today (`kin impact get*`, case variants, substring names) flip to not_found. Exact fail-closed matching is right for `--json`; on the human surface it is a breaking change, and the PR body's "legacy CLI behavior remains intact" is not accurate. Either restore loose resolution for the legacy text path, or declare the break explicitly, add tests covering it, and update the not-found guidance to state that matching is now exact.

3. **Ranked BFS runs and is serialized on every /impact request** (`crates/kin-cli/src/commands/impact.rs` + `crates/kin-review/src/ranked_impact.rs`)
`build_impact_response` unconditionally runs `rank_impact` — per-edge path Vec clones plus full `RelationEvidence` clones — and ships the complete uncapped report over the wire even for text-mode callers that discard it. The daemon POST /impact path (api.rs:3504) pays this on the shared serialized daemon for output nobody consumes. Gate ranked computation on a request flag; `require_unique` is already transported and is the natural gate.

**Non-blocking (fix alongside or ticket)**

- `PRIORITY_SCORE_FORMULA` documents `round(100 * clamp(confidence, 0, 1))` but the code computes `(round(10000*c) + 50) / 100` (double rounding). The PR's own test shows the divergence: 0.805f32 yields 81 in code vs 80 by the published formula. Make the string exact or compute `round(100*c)` directly.
- The UUID fast path in `resolve_entities` returns before applying `--file/--kind/--signature`, so contradictory qualifiers are silently ignored instead of failing closed.
- Under `--json`, ambiguous and not_found resolutions exit 0; automation must parse `resolution`. Return nonzero or document the exit contract.
- The two surfaces in one response can disagree: legacy `lines` use `get_downstream_impact`, which follows all inbound relation kinds with no depth clamp (kin-db traverse.rs:235 has no kind filter), while `ranked` clamps depth to 8 and filters through `is_impact_relation`. Entities can appear in the text list but be absent from `ranked.candidates` with no reported gap. Document the allowlist and clamp in the report, or align the sets.
- `RankedImpactCandidate.hop`/`path` reflect the best-scoring encounter, not minimum graph distance (`candidate_is_better` prefers score over hop), so `hop` can exceed the entity's true distance. Document the semantics on the struct.
- `ImpactResponse.schema_version` carries the ranked schema id even when `ranked` is absent (not_found/ambiguous).

**Clean**

Determinism holds: BTreeMap-keyed accumulation, total-order relation sort (proved by the storage-order test), floats become integer basis points before comparison, and the final sort tie-breaks end in `entity_id`. No benchmark-relevant serving default changes: kin-bench routes impact through `/spine/impact` and MCP `impact_analysis` routes through `kin_review::impact`, both untouched. The hold is solely for the three items above.

troyjr4103:
2026-07-20 merge-captain recheck: preserving as draft, not merge-ready. Remote head e6790eab is 212 commits behind current main and has no approval. The three blocking review findings remain unresolved: wire compatibility under CLI/daemon skew, legacy resolution regression, and unconditional ranked BFS/serialization. The clean local worktree contains old main-integration commits already present on current main, not fixes for these blockers. No merge was attempted.

troyjr4103:
Superseded by #376. The replacement is replayed onto current `main`, fixes all three blocking review findings plus the documented non-blockers, and passed the full local CI-equivalent gate. Closing this stale draft; its queue entry will be cancelled when it reaches the head.

PR #296, opened 2026-07-10
Resume legacy history hydration from semantic checkpoints

troyjr4103:
## Summary

- resume legacy Git history hydration from graph-owned semantic checkpoints
- validate checkpoint identity, source history, build identity, parser and linker frontiers, and parent/change topology before trusting reuse
- make checkpoint publication crash-safe, concurrency-safe, bounded, deterministic, and fail-closed
- recover older truncated and empty-final-HEAD repositories without silently accepting partial or incompatible state
- reject symlink and path-alias boundary violations while preserving supported macOS /tmp and /var aliases

## Correctness boundaries

- graph truth remains the authority; no raw filesystem search is introduced
- corrupt manifests, corrupt segments, invalid parent graphs, parser/linker drift, unknown builds, and unsafe paths cannot authorize checkpoint reuse
- dirty or unknown build identities fall back to a full replay without reading or writing checkpoints
- refusal paths preserve graph and checkpoint authority
- successful warm replay is semantically idempotent across graph root, repo-truth hash, entities, relations, changes, reverse indexes, revisions, file hashes, artifact index, and checkpoint digests

## Verification

- independent exact-head SHIP review on the hydration code tree with no blocking correctness findings
- signed v0.2.15 binary upgrade tests for source-present and empty-final-HEAD legacy repos
- warm resumption at 2/2 and 3/3 checkpoints with byte-identical checkpoint digest maps
- corruption, frontier, topology, symlink, concurrent publisher, and crash recovery fixtures
- full locked kin-cli verification: 841 library tests, 1 binary test, and 40 integration tests
- exact CI clippy allowlist with -D warnings, formatting, zero-file-search, and runtime-boundary guards
- 15/15 commits include DCO trailers
- final head 90fbc5b87b84163ebea30c7a59a3d9cbb15f9988 integrates main da2ede70c588293c9dd70bac25a19ed7cf939ff9; Cargo.lock and crates are byte-identical to the independently reviewed cfd0fd0 code tree, and the full kin-cli suite passed again after integration

## Known separate debt

Repeated successful saves can produce different raw graph.kndb bytes while retaining identical semantic graph state. That pre-existing unordered snapshot serialization issue is tracked in FIR-824 and is not introduced by this change.

One verification-status test flaked once, then passed isolated, 32 concurrent repetitions, and two full suites. It remains visible as separate test-stability debt.

## Coordination

This PR stays draft until fresh remote CI passes and it reaches the serialized Kin merge-queue head. It does not claim release or completion before merge verification.

Linear: FIR-1373


troyjr4103:
Superseded by #305. The replacement preserves this checkpoint implementation, integrates current main without force-pushing this shared draft, keeps the broader MCP environment-test serialization, and fixes the equal-timestamp child-before-parent ordering defect with deterministic parent-first topology. #305 remains draft pending independent review and fresh CI.

PR #301, opened 2026-07-10
VFS tree-cache and history-hydration hardening (rebuilt)

troyjr4103:
Successor to #280. Preserved by hygiene pass. Needs owner review + CI before queue entry.

FIR-1444

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3152,6 +3152,7 @@ dependencies = [
 name = "kin-daemon"
 version = "0.2.15"
 dependencies = [
+ "anyhow",
  "async-stream",
  "axum",
  "chrono",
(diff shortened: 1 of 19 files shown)


troyjr4103:
Closing as superseded after current-main source audit. The two core VFS snapshot cache/single-flight patches are already represented on `main`; the remaining history-hydration implementation targets the retired pre-preparation API. Current `main` has the later fail-closed replacement chain (`a27f5d63`, `975a1e6d`, `9a312edb`, `6aa9f14d`, plus counted hydration fixes `743e0f90`/`bd9534ff`) with checkpoint authority, graph-authority preservation, and prepare-before-render acknowledgement. Replaying the old branch would regress those newer boundaries. No unique current architecture remains to merge.

PR #302, opened 2026-07-10
Honor safe renames in shadow review

troyjr4103:
May be superseded by merged #295 — owner to confirm before review.

FIR-1440


troyjr4103:
Superseded by merged PR #295, which implements the same FIR-1440 shadow downgrade with the stronger demote-not-suppress evidence semantics and full parse→link→persist→impact→shadow coverage. This draft would regress that behavior by removing the downstream warning. The remaining default-value correctness gap is being handled as a separate current-main follow-up. Closing to keep one implementation authority; no release or pinned-row claim is made here.

PR #334, opened 2026-07-13
chore: remove ticket references from source comments

troyjr4103:
## What

Removes internal tracker ids from source comments and one user-facing runtime string, so the public source describes durable technical behavior only.

## Changes

- `crates/kin-cli/src/commands/health.rs` — drop the tracker id from the shim-reinstall-hint doc comment and the VFS-remediation test comment.
- `crates/kin-cli/src/commands/setup.rs` — drop the tracker id from the `doctor` VFS-repair comment and two shim-restore test comments.
- `crates/kin-review/src/shadow.rs` — drop the tracker id from four doc/line comments describing the deep-history impact ceiling, from two `downstream_risk` test comments, and from the user-facing `deep_history_impact_ceiling` evidence-detail string. The evidence-detail string is behavior-visible, so its emitted text changes (`...historical-substrate fidelity: across a range this deep...`); the meaning is unchanged.

All edits are comment/string-literal only; each sentence keeps its technical meaning.

## Verification

`cargo check -p kin-cli -p kin-review` passes.

Draft — held for post-freeze review. Do not merge during the launch freeze.

### DIFF
--- a/crates/kin-cli/src/commands/health.rs
+++ b/crates/kin-cli/src/commands/health.rs
@@ -286,8 +286,7 @@ fn check_vfs_projection() -> HealthCheck {
 /// The durable step that repairs a missing or corrupt shim when `kin doctor
 /// --fix` cannot source one locally. It must NEVER name `kin doctor --fix`
 /// itself: this text is reprinted in the post-`--fix` "still needs manual steps"
-/// list, where pointing back at the command that just ran is a dead loop
-/// (FIR-1409).
+/// list, where pointing back at the command that just ran is a dead loop.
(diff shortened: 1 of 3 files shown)


troyjr4103:
Superseded by #367, which carries the same source-comment cleanup on current `main` together with the graph-authority recovery work. #367 merged as `628e66945506bc503441b7771511d2d255d6a6f9` after the full local `kin-cli` suite (1,037/1,037) and all exact hosted checks passed. Closing this stale draft so it cannot be mistaken for remaining release work.

PR #336, opened 2026-07-13
Restore registry dependency waves

troyjr4103:
## Outcome

Restores the dependency-wave workflow that was removed in the unrelated npm-recovery commit. The current registry publisher sends `kin-registry-release`, but Kin has no matching workflow, so the successful kin-vfs-core 0.1.5 dispatch created no consumer PR.

This workflow:

- consumes the canonical `kin-registry-release` event;
- runs an hourly missed-callback reconciliation at minute 17;
- serializes events instead of cancelling in-flight updates;
- keeps the default GITHUB_TOKEN read-only;
- mints a one-hour selected-repository GitHub App token behind the main-only `release-followups` environment;
- calls the SHA-pinned kin-actions v0.1.9 dependency workflow and opens a signed-off PR only after the Kin smoke command passes.

## Verification

- actionlint over every workflow
- release authority policy test
- diff check
- DCO sign-off

## Activation gate

Do not merge until the selected-repository GitHub App is installed for `kin`, `homebrew-kin`, and `kin-infra`, with Actions read, Contents write, and Pull requests write, and its ID/private key are stored in the main-only `release-followups` environment. The same App then activates the existing immediate release follow-up callback.

### DIFF
--- a/.github/workflows/kin-dependency-wave.yml
+++ b/.github/workflows/kin-dependency-wave.yml
@@ -1,0 +1,54 @@
+name: Kin Dependency Wave
+
+on:
+  repository_dispatch:
+    types: [kin-registry-release]
+  schedule:
+    # Recover any missed registry callback by resolving every watched crate.
+    - cron: "17 * * * *"
+
+# The reusable workflow receives the dedicated PR token below. Keep the
+# per-run GITHUB_TOKEN read-only so a caller-controlled token cannot become a
+# second write path.
+permissions:
+  contents: read
+
+concurrency:
+  group: kin-registry-dependency-wave
+  cancel-in-progress: false
+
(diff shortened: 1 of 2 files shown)


troyjr4103:
Closing as superseded after the 2026-07-20 merge-captain audit. Current main deliberately retires `.github/workflows/kin-dependency-wave.yml`, and `scripts/test-release-workflow-authority.py` now fails if it reappears because `release.yml` owns protected public writes. This July 13 branch is 63 commits behind main and would restore an authority path the current release policy forbids. FIR-1270 remains open for the current-main automatic follow-up implementation and end-to-end outcome proof; closing this PR is queue cleanup, not a claim that automation is fixed.

PR #344, opened 2026-07-15
Fix optional VFS installer reporting

troyjr4103:
## What changed

- probe the installed `kin-vfs` binary before claiming filesystem projection is available
- remove the unusable optional VFS binary and shim when the host loader cannot execute them
- preserve the working `kin` and `kin-daemon` installation
- cover both runnable and non-runnable projection paths in CI

## Why

The v0.2.23 static core runs on Alpine, but the optional VFS payload is currently glibc-linked. Debian 12 can also reject the VFS binary because it requires a newer glibc. The installer previously reported the projection as installed whenever the files existed, even when the host could not load them.

This closes [FIR-1478](https://linear.app/firelock-ai/issue/FIR-1478/installer-falsely-reports-vfs-projection-installed-when-the-host) without weakening the core install. Portable VFS build targets remain separate in FIR-1480.

## Validation

- `shellcheck -e SC3043 scripts/install.sh`
- `python3 scripts/test_install_optional_vfs.py` — 2 passed
- `python3 scripts/test_installer_parity.py` — 7 passed
- `python3 scripts/test-release-workflow-authority.py` — passed
- exact public v0.2.23 Alpine 3.24 arm64 negative-path acceptance
- exact public v0.2.23 Ubuntu 24.04 arm64 positive-path acceptance

## Claim boundary

This PR does not make the VFS payload musl-compatible or lower its glibc requirement. It makes installer reporting and installed optional state truthful.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,6 +74,7 @@ jobs:
       - name: Test release ordering and npm provenance policy
         run: |
           python3 ./scripts/test-release-workflow-authority.py
+          python3 ./scripts/test_install_optional_vfs.py
           python3 ./scripts/test_installer_parity.py
           python3 ./scripts/test-npm-automatic-release.py
           python3 ./scripts/test-verify-npm-release.py
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -214,7 +214,19 @@ done
 ok "Binaries installed (kin, kin-daemon)"
 
 if [ "$HAVE_VFS" = "1" ] && [ "$HAVE_SHIM" = "1" ]; then
-    ok "Filesystem projection installed (kin-vfs + shim)"
+    # The release archive can contain a projection built for a different libc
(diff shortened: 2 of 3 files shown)


troyjr4103:
Independent review at exact head `da4fad61693ac709c581332d97af8a7d9e2c77f5`: conditional GO on code, with no P0/P1/P2 findings. The installer now probes `kin-vfs --help` before reporting projection support, preserves compatible-host files, removes only unusable optional VFS/shim files, and keeps core install/checksum/PATH/setup behavior intact.

Exact-head CI is not yet green. Code Coverage job 87413124466 failed because the GitHub runner exhausted disk while writing its diagnostic log (`System.IO.IOException: No space left on device`), not because a test assertion failed. macOS and Windows jobs are still active. This PR stays draft until the same SHA has a fully green retry.

Non-blocking review notes: the positive fixture could assert the shim explicitly, and synthetic archive CI is not a replacement for the retained public-byte platform evidence in FIR-1478. Neither requires a code change at this head.

PR #383, opened 2026-07-21
[VERIFICATION ONLY — DO NOT MERGE, will be closed] FIR-1501 regression assertion against unmodified main

troyjr4103:
Throwaway draft opened solely to run the FIR-1501 regression assertion against unmodified `main` in CI, because CI does not trigger on branch push. Expected to FAIL: on this commit the `_for_locate` entry point runs the full semantic replay, so the hydrated tip carries entity deltas.

Evidence for #382. This PR will be closed and its branch deleted as soon as the Check & Test job reports.

### DIFF
--- a/crates/kin-cli/src/commands/ref_lookup.rs
+++ b/crates/kin-cli/src/commands/ref_lookup.rs
@@ -879,6 +879,45 @@ mod tests {
         );
     }
 
+    /// Verification-only: this is the FIR-1501 regression assertion expressed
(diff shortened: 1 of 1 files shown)


troyjr4103:
Purpose served. Recorded in #382: on unmodified main this assertion FAILS with `the locate entry point must hydrate at artifact-only depth, but it produced 1 entity deltas` (job 88680007150). Closing and deleting the branch.

PR #390, opened 2026-07-21
[throwaway] Probe the artifact-only hydration before-state

troyjr4103:
Not for merge. FIR-1506 before-state evidence.

One assertion on top of unmodified `main` (4998a4bc): after `resolve_ref_importing_git_if_needed_for_locate` hydrates a Git ref at artifact-only depth, the semantic entry point resolves the same ref successfully, reports `hydrated_changes == 0` (nothing re-hydrates, nothing upgrades the depth), and lands on a change with no replayed entity deltas — the state from which `history`, `blame`, and `review` answer empty.

Expected result: `a_semantic_caller_resolves_into_unreplayed_history_after_a_locate_hydration` FAILS. The failing job id is the before-state evidence for #389, after which this PR is closed and the branch deleted.

### DIFF
--- a/crates/kin-cli/src/commands/ref_lookup.rs
+++ b/crates/kin-cli/src/commands/ref_lookup.rs
@@ -1030,6 +1030,33 @@ mod tests {
             .expect("hydration must insert the requested commit")
     }
 
+    /// Before-state probe, expected to FAIL on this commit. `locate --ref`
+    /// hydrates at artifact-only depth and persists it; a semantic entry point
+    /// then resolves the same ref successfully into history whose per-commit
+    /// deltas were never replayed, and reports an empty answer that cannot be
+    /// told apart from a real negative.
+    #[test]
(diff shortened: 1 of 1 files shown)


troyjr4103:
Before-state recorded; closing.

`Check & Test (ubuntu-latest)` FAIL — https://github.com/firelock-ai/kin/actions/runs/29850963976/job/88703428132
`Check & Test (macos-latest)` FAIL — https://github.com/firelock-ai/kin/actions/runs/29850963976/job/88703428145

`1046 passed; 1 failed`, the one failure being this probe:
`a semantic caller resolved into history with no replayed entity deltas: history, blame, and review answer empty here` (ref_lookup.rs:1054). The preceding `hydrated_changes == 0` assertion passed, which is the mechanism itself: the tip is present, so nothing re-hydrates and nothing upgrades its depth.

Fix and green run: #389.

PR #445, opened 2026-07-26
Narrow the whole-graph entity fallback and compact history output

troyjr4103:
## The bug

`kin history alwaysTrue` answered:

```
Multiple entities match 'alwaysTrue': AND_THEN, AND_WHEN, ANON_TEST_CASE, Approx, Approx::Approx
```

None of those contain the query. `resolve_entity_query` filtered candidates by name and, when that came back unusable, swept `list_all_entities()` **without reapplying the filter**. `choose_entity_match` then reported the first five entities alphabetically as the "matches".

`resolve_entity_query_at_ref` already filtered and has no such fallback, which is why the `--ref` forms of `history` and `blame` worked while the ref-less forms did not. One fix covers both commands.

The sweep now reapplies `entity_matches_query`, and when nothing matches it returns the first attempt's error rather than burying it under a wider, wrong one.

## History rendering

Two full 64-character identifiers per revision plus the entire commit body pushed every readable field past the right edge. A three-revision history rendered as ~20 lines of hex and prose.

Before:
```
  e384bddd96e96a545874b562fe683faa7d9215df2cdc95b4c79713df8877a4fe @ 770707931ec80918f542e6b7df7e703e6e03fa2945d21c25461ff5cff66faab8 - Big assertion capture refactoring.
- moved as much logic out of the macros as possible
...
```

After:
```
History for 'alwaysTrue' (Function, cpp) at eb45be5f9798:
  770707931ec8  2014-05-28  Phil Nash             Big assertion capture refactoring.
  d6d034db7cf6  2017-01-19  Mickey Rose           use sizeof(expr) for unevaluated syntax check
  eb45be5f9798  2017-01-30  Martin Hořeňovský     Revert "use sizeof(expr) for unevaluated syntax check"
```

Identifiers abbreviate to 12 characters, messages reduce to their subject line, timestamps render as calendar dates, authors drop the email. Matches `git log --oneline`.

## Verification

- Regression test added, and **falsified**: reverted the fallback, the test fails with the exact production signature, restored, it passes.
- End to end against a live daemon: the misleading list is gone and `history --ref` returns the full timeline.
- `cargo test -p kin-cli --lib`: 1147 passed, 4 failed. All 4 fail identically on `main` at 570e4b8e and are unrelated (MCP config repair, transaction manifests).

## Note for reviewers

The daemon renders history server-side, so a rebuild alone does not change behavior. `kin daemon stop` in the target repo first.

Separately and not addressed here: entities are not resolvable without `--ref` on a graph whose default head predates them. That is a staging/hydration question, not this fallback.


troyjr4103:
Second commit: the MCP delegate misread busy daemons as dead. 5s request timeout + pooled sockets + revival-on-first-transport-error meant every tool call failed for the first minute after  while the daemon ran enrichment, and revival could never succeed because the live daemon holds the repo lock. Now: 60s request timeout (500ms connect timeout still catches dead daemons fast), pooling off, one same-URL retry before revival. Regression test included. Diagnosed live: the daemon answered /health throughout while tool calls timed out; the earlier structured error was the tell.

troyjr4103:
Added f455e89b: impact output now groups impacted entities by traversal hop with the direct-caller group labeled and file:line locations, prefers exact-name matches over substring cousins in the human path (an exact ask no longer trips the ambiguity note), and refs counts referencing entities per file+name instead of one row per file. kin-bench PR 155 keeps its live-agent summary parser compatible with both header spellings.

PR #471, opened 2026-07-28
Classify authority bind refusals on namespace identity alone

troyjr4103:
Revalidating a startup-pinned repository namespace previously rode a full authority load,
and `open_bound` mapped every error that load produced onto one refusal arm. A truncated
snapshot, a missing or non-regular `.lock`, and a quarantined state on a namespace this
process still reached were therefore diagnosed as a replaced namespace and returned as
`409 CONFLICT`.

The narrower identity probe corrected that diagnosis, but two later reviews found reopen
boundaries that also had to remain fail closed. First, an intact pinned namespace with its
persisted `authority.json` removed must not be reopened as a fresh in-memory
generation-zero repository by the startup/payload path. Second, local repository-transfer
status and receive endpoints must use that same persisted-record contract rather than
opening the retained backend directly. The current head handles all three conditions.

## The change

Revalidation now goes through kin-db's typed `probe_pinned_repository_namespace`, which
answers the identity question and only the identity question.
`revalidate_pinned_local_namespace` maps its probe variants onto
`PinnedNamespaceRefusal`:

- `Identity` for a namespace replaced at the same ambient path, detached, occupied by
  something that is not a directory, displaced during creation, a replaced or detached
  storage root, or a store that does not hold the repository at all
- `Unavailable` for faults that say nothing about identity, including IO, permissions, or
  an unreadable entry

`open_bound` maps those two arms onto the two existing
`RepositoryAuthorityBindRefusal` arms, so `409` is reserved for an actual identity loss
and other probe failures remain internal errors. Both matches are total with no catch-all.

The downstream authority open goes through
`open_persisted_local_repository_authority`. It consumes the payload receipt returned by
the same KinDB recovery and refuses when that receipt is absent. That preserves KinDB's
ability to deliberately construct an unpersisted generation-zero repository while making
startup/reopen of an initialized local repository require a persisted authority record.

The helper is now generic over `StorageBackend`, allowing every local repository-transfer
route to use the identical persisted-record boundary through the daemon's retained local
backend capability. This covers transfer status/read operations and the durable
`/transfer/receive` write decision. Hosted daemon storage keeps the raw open because a
genuinely new hosted repository may legally begin at generation zero; only the initialized
local startup binding carries the stricter reopen contract.

## One authority recovery and one recovery lock

The old revalidation called `LocalFileBackend::load_snapshot_authority`, which took the
repository's exclusive recovery lock and decoded authority before answering. An exposed
checkout or branch bind therefore paid once in revalidation and again in `open_manager`.

The identity probe does neither. Its steady-state path revalidates the pinned storage root
and compares the repository directory identity against the pinned one. The subsequent
authority open calls `open_with_payload_stats` exactly once and uses that recovery's
receipt to enforce the persisted-record boundary. There is no second authority recovery
or recovery lock just to distinguish a missing record.

KinDB may separately persist a reusable history-validation proof under its own lock after
a full replay. That does not reload authority or decide whether the persisted authority
record exists. The identity probe also takes brief in-process mutexes over the backend's
pinned-namespace and poisoned-namespace maps; those are not repository recovery locks and
guard no IO.

## Coverage

Core and daemon startup coverage includes:

- `revalidation_passes_a_corrupt_payload_on_an_intact_namespace_to_the_authority_open`
- `revalidation_classifies_a_replaced_namespace_as_an_identity_refusal`
- `revalidation_classifies_a_detached_namespace_as_an_identity_refusal`
- `unavailable_probe_is_a_non_identity_refusal`
- `authority_open_refuses_an_intact_namespace_missing_its_persisted_record`
- `open_fails_when_persisted_authority_record_is_missing`

The transfer repair adds two load-bearing API regressions that start from a valid local
daemon, remove only `authority.json` after startup while retaining snapshot material, and
then exercise the real routes:

- `repository_transfer_status_refuses_authority_record_lost_after_startup` proves the
  read/status path returns `424 FAILED_DEPENDENCY` and does not recreate authority
- `repository_transfer_receive_cannot_recreate_authority_lost_after_startup` builds a
  valid transfer pack for an unborn destination, proves the durable receive path returns
  `424 FAILED_DEPENDENCY`, and proves no authority record is written

Those tests fail against the prior raw-open transfer path: status accepts generation zero,
and receive can durably republish authority.

One CLI unit fixture had modeled an absent authority while testing exact ref-alias behavior.
The stricter reopen contract correctly exposed that stale setup. The fixture now initializes
a real persisted empty authority.

## Why this does not wait on a kin-db release

The probe (`probe_pinned_repository_namespace` and `LocalNamespaceProbe`) is present in
kin-db 0.7.8, which `main` already pins. The repair consumes the payload receipt already
returned by KinDB's published `open_with_payload_stats` API. No kin-db change, dependency
version bump, or publish train is required.

## Scope

FIR-1611 lists three fixes and this carries all three: the typed identity probe with
classification, the single-recovery bind, and the bind documentation checked against the
final shape. The kin-db half of the ticket shipped in 0.7.8. The transfer fix closes the
local API authority bypass discovered while independently reviewing that final shape.

## Local gates

Current head `3c930aad2d71cfa263bc3721cd096633a0bc0770` passed:

- the exact CI clippy allowlist over all targets and all features with
  `RUSTFLAGS=-D warnings`
- `cargo build --all-targets --locked` with `RUSTFLAGS=-D warnings`
- full `cargo test -p kin-daemon --locked`: 618 library tests plus all daemon binary and
  integration suites passed
- both new regressions exactly and the four-test `repository_transfer` filter
- `cargo test -p kin-core repository_authority --locked`: 12 passed
- `cargo test -p kin-cli --no-default-features --locked`: 1,035 library tests plus all
  binary, integration, and doctest suites passed
- `scripts/verify-zero-file-search.py`
- `scripts/zero_file_search_guard.sh`
- `scripts/check_runtime_boundaries.sh`
- `scripts/check_private_repo_coupling.sh`
- `scripts/test-private-repo-coupling.py`: 4 passed

The immediately preceding head also passed the full workspace nextest/doctest gate (4,693
nextest cases passed, 10 skipped), but that earlier run is not being used as proof of the
new transfer delta.

## Not claiming

- Fresh hosted CI on `3c930aad` is still required, including the Windows authority leg.
  No Windows leg was run locally.
- This is source and local-test evidence, not released-byte, external-client, install, or
  production proof.
- This PR has not been submitted to the merge queue and must receive a fresh independent
  review of the repaired head before admission.

Closes FIR-1611

### DIFF
--- a/crates/kin-cli/src/commands/ref_lookup.rs
+++ b/crates/kin-cli/src/commands/ref_lookup.rs
@@ -647,8 +647,10 @@ mod tests {
     #[test]
     fn full_git_oid_is_not_converted_into_a_synthetic_change_id() {
         let graph = kin_db::InMemoryGraph::new();
-        let layout = kin_core::KinLayout::new(std::path::PathBuf::from("/absent/.kin"));
-        let binding = absent_binding(&layout);
+        let directory = tempfile::tempdir().unwrap();
+        let initialized = kin_core::init(directory.path()).unwrap();
+        let binding = kin_core::LocalRepositoryAuthorityBinding::from_layout(&initialized.layout)
+            .expect("fixture must carry persisted empty repository authority");
         let error = resolve_ref(
             &graph,
             &binding,
--- a/crates/kin-core/src/lib.rs
+++ b/crates/kin-core/src/lib.rs
@@ -96,6 +96,7 @@ pub use ranking::{
 };
 pub use ref_view::{build_graph_at_ref, collect_changes_at_ref, filter_vector_results_to_scope};
 pub use repository_authority::{
-    durable_semantic_enrichment_summary, revalidate_pinned_local_namespace,
-    DurableSemanticEnrichmentSummary, LocalRepositoryAuthorityBinding, PinnedNamespaceRefusal,
+    durable_semantic_enrichment_summary, open_persisted_local_repository_authority,
+    revalidate_pinned_local_namespace, DurableSemanticEnrichmentSummary,
+    LocalRepositoryAuthorityBinding, PinnedNamespaceRefusal,
 };
--- a/crates/kin-core/src/lib.rs
+++ b/crates/kin-core/src/lib.rs
@@ -97,5 +97,5 @@ pub use ranking::{
 pub use ref_view::{build_graph_at_ref, collect_changes_at_ref, filter_vector_results_to_scope};
 pub use repository_authority::{
     durable_semantic_enrichment_summary, revalidate_pinned_local_namespace,
-    DurableSemanticEnrichmentSummary, LocalRepositoryAuthorityBinding,
+    DurableSemanticEnrichmentSummary, LocalRepositoryAuthorityBinding, PinnedNamespaceRefusal,
 };
--- a/crates/kin-core/src/repository_authority.rs
+++ b/crates/kin-core/src/repository_authority.rs
@@ -228,7 +228,9 @@ impl LocalRepositoryAuthorityBinding {
     /// Establish a binding once for a fresh one-shot CLI/offline process.
     ///
     /// Request handlers in a long-lived daemon must receive an existing
-    /// binding instead of calling this constructor.
+    /// binding instead of calling this constructor. This pins namespace
+    /// identity only; the first authority open below also proves that the
+    /// namespace still carries a persisted authority record.
     pub fn from_layout(layout: &KinLayout) -> Result<Self> {
         layout.check_version()?;
         let manifest = KinManifest::load(&layout.manifest_path())?;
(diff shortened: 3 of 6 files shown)


troyjr4103:
Updated gating, with a dependency finding that widens what this PR waits on.

`Cargo.toml` now requires `kin-db = "0.6.6"`, not 0.6.5. Two lanes bumped kin-db to the same 0.6.5 with byte-identical edits, so the rebase merged them silently; the identity probe lands in 0.6.6 (kin-db #136, green on every check).

**This PR is now blocked on more than the kin-db mint.** kin-db #137 changed kin-db's dependency from `kin-model = "0.6.1"` (caret) to `= "0.6.4"` (exact). That exact pin propagates: kin pins `kin-model = "=0.6.1"`, and the two cannot co-resolve.

```
error: failed to select a version for `kin-model`.
    ... required by package `kin-db v0.6.6`
versions that meet the requirements `=0.6.4` are: 0.6.4
  previously selected package `kin-model v0.6.1 (registry `kin`)`
    ... which satisfies dependency `kin-model = "=0.6.1"`
```

So consuming kin-db 0.6.6 forces kin to move to kin-model `=0.6.4`, which forces the kin-side FIR-1621 consumption. I measured it: roughly twenty `RepositoryTransaction` struct literals across `kin-git`, `kin-core` (init.rs, tree.rs), `kin-remote`, `kin-mcp`, and `kin-cli` tests need the new `merge_transaction_delta` field. I prototyped two of them (`kin-git/src/admission_history.rs:143`, `kin-remote/src/repository_transfer.rs:1111`, both `None`, both generation-zero or fast-forward paths where a merge delta cannot apply) purely to measure the shape, then reverted. This PR carries none of it.

I did not absorb that work. It is FIR-1621's kin-side scope, it is not a mechanical one-liner, and some of the `kin-core/src/tree.rs` sites are on merge paths where `None` may not be the right value. Silently folding twenty edits across five crates into an authority-classification PR would be the wrong shape for review.

Ordering this now needs:

1. kin-db #136 merges, 0.6.6 mints and publishes
2. kin consumes kin-model `=0.6.4` (FIR-1621 kin side, unassigned as far as I can tell — no open kin PR is doing it)
3. this PR re-locks patch-off, re-gates, and enqueues

Happy to take step 2 if it is not already owned; it just should not ride in here unannounced.

The classification work itself is unchanged and verified: `cargo test -p kin-core -p kin-daemon` exits 0 with 447 + 459 passing, clippy with the full ci.yml allowlist exits 0. I attempted `cargo test --workspace` to close that gap and it is the run that surfaced the resolution conflict above, so the workspace suite remains unverified until step 2 lands.

PR #474, opened 2026-07-28
Incremental historical enrichment and cheap admission proofs

troyjr4103:
# Incremental historical enrichment and cheap admission proofs

Targets the `O(history x tree)` admission term measured in FIR-1601 and the hoistable-preflight and dead-env-pin items from FIR-1629. Four workstreams, one dependency chain.

> **Takeover correction (2026-07-28):** correctness repairs are pushed at `ba9b004` + `8e4cb41`, have an independent GO review, and the exact hosted head is fully green across macOS, Ubuntu, Windows, daemon/MCP, Docker, DCO, dependency, secret, and launcher checks. This draft is still not merge-ready: the fresh full-corpus before/after pair is open, the synthetic `commits * files` coefficient remains a regression in the last valid fit, and the prior local full-workspace wrapper did not produce an accepted gate. The evidence below remains conservative until those substantive gates pass.

## 1. Post-parse enrichment is incremental per commit (kin-index)

Parsing was already memoized against parent state; everything after the parse was recomputed from scratch at every commit: whole-tree clones of every file's entities/relations/imports, fresh completeness and identity maps, a full sort, a whole-tree cross-file link, and full-tree entity/relation diffs.

Now:

- Parse payloads are `Arc`-shared per file version; carrying an unchanged file forward is a pointer clone.
- The tree entity and relation maps are patched from the first parent; the whole-tree duplicate-identity refusal is preserved exactly (kept-vs-kept pairs were validated by the parent, everything else re-checks on insert).
- First-parent deltas are emitted from the patch. An identity untouched by the patch carries the same shared value on both sides, and a patched identity value-compares old against new, so the output is identical to the full-map diff by construction.
- Only invalidated files relink. The invalidation set is conservative by design, per rule:
  - **Names**: a kept file's resolution probes the global indices only at keys derivable from its own parse payload (destination names, bare leaves, `::`/`.` segments, type-qualified suffix pairs). The indices change only at keys named by the changed files' old and new entity sets. Intersection relinks.
  - **Module targets**: import/pin/namespace/default-export resolution reaches other files through `resolve_module_path`, which is a pure function of the path set; per-file resolved targets are cached per path-set era (eras compared by the sorted path list itself, never a digest) and a changed file that is a target relinks its importers.
  - **Include topology**: a changed include-target list relinks everything that can reach the changed file through the old or new include graph (reverse reachability, unbounded, a superset of the bounded forward closure).
  - **Hierarchy**: inheritance walks discover class names mid-walk that no static token set can enumerate. Any change to a file's hierarchy projection (Extends pairs, class-like entity names/kinds, imports when it declares bases) or to any entity sharing a name with a declared base relinks every walk-capable file (only owner-dotted and `::`-scoped receiver calls can enter a walk).
  - **Structural commits** (any add/remove/rename/classification flip of a semantic file) relink the whole tree, in parallel through the same pure per-file resolution the batch linker uses.
- The batch linker's resolution code is untouched; context construction is factored over borrowed refs so both paths share one code path (and batch callers stop deep-cloning the entity universe per call).

Two fail-closed audits convert any invalidation gap into a refusal instead of silently wrong history:

- a carried-forward link output that references a net-removed entity, or whose dropped external placeholder collides with a net-added entity, fails admission loudly;
- `KIN_HISTORY_LINK_VERIFY=1` recomputes every commit through the batch path and requires identical entity maps, relation maps, and emitted deltas.

## 2. The migration proof re-observes instead of re-capturing (kin-git)

`observe` ran a full `capture_lossless_git_repository` per observation: with two observations per proof and three proofs per init, six of the seven full object-closure captures per admission were TOCTOU stability re-checks re-reading and re-hashing every blob body into the CAS.

`verify_lossless_snapshot_unchanged` now re-proves the boundary against the held snapshot: shallow boundary, object format, byte-exact refs and HEAD (using the same conversion code as capture), and a structural walk of the reachable closure. Every reachable commit, tree, tag, and blob body is re-read byte-exactly and fully re-verified for kind, declared length, body hash, and recomputed OID. This closes same-kind/same-length blob substitutions in the source ODB instead of trusting the initial capture. The two-observation structure and every mutable-boundary proof (index, worktree, hooks, filters, remotes) are unchanged.

## 3. Plan validation runs once, not six times (kin-git)

`SemanticGitImportPlan::validate` is a full raw-object re-derivation plus deep compare. It ran at bind (inside `with_historical_semantics`), three times inside the migration proofs (`validate_plan_binding`), at admission, and inside the admitted-plan audit. Now:

- `with_historical_semantics` validates that the plan is unenriched and that every change identity recomputes from its payload (which pins the whole DAG transitively through parent identities) instead of re-deriving the plan; the full rebuild-and-compare audit still runs exactly once at `admit_semantic_git_import`, so a plan that does not match its raw-object derivation still fails closed before admission.
- `validate_plan_binding` checks binding (repository id, format, objects, refs, head) only.
- The admitted-plan audit in `into_generation_zero_repository_transaction` is unchanged (one more full rebuild, deliberately kept).

## 4. Preflight refusals in milliseconds; dead env pins refuse loudly

- `preflight_git_source_compatibility` (new, public) runs the snapshot-free refusal subset (shallow, in-progress operations, materialized-worktree identity, sibling registered worktrees, hook/filter blockers) and is called at the top of `init_from_git`, before any capture. The same helper backs the full proof, so the two can never drift. A refusable source now fails in milliseconds instead of after a ~28-minute capture. Note `kin clone` still deletes the clone destination on admission failure; with the preflight the failure now happens before capture/enrichment, but the destination-deletion UX itself is untouched here.
- Removed `KIN_*` variables are now a first-class registry class: setting `KIN_INIT_WARM_CACHE` hard-fails startup in every validation mode with a message naming the variable and why it is dead, instead of the generic possible-typo warning. The freeze protocol's contamination pin can no longer silently no-op.

## Replay semantics and the manifest

Output identity remains required, but the corrected divergent-head cache identity changes persisted replay semantics, so `HYDRATION_SEMANTICS_VERSION` is intentionally bumped from 5 to 6. Digests for the edited pinned functions are regenerated, and the new derivation core (`derive_tree_semantics`, `link_single_file`, `invalidated_files`, the analyzers, patch-delta emitters, and the carried-output audit) is added to the guarded set with owners and reasons. All 21 guarded functions match the canonical v6 manifest.

Evidence of output identity:

- Six new equivalence fixtures run every commit through both paths with the batch result as the oracle (content edits, ambiguity shifts, structural commits, Python hierarchy edits, TS import rewiring, merges). The oracle caught one real bug during development (structural-mode relation deltas) before it ever shipped.
- `real_repository_enrichment` with `KIN_HISTORY_LINK_VERIFY=1` completed on fd: 1,999 changes across 8 heads cross-checked batch-vs-incremental in 144s wall.
- The corresponding ripgrep verification did not complete; no ripgrep after result exists.
- On corrected head `8e4cb41`, the same-length blob-corruption and divergent-leaf replay regressions pass; canonical manifest verification, formatting, and `-D warnings` clippy for kin-git + kin-index are green. An independent source rereview returned GO with no findings. Hosted checks on the exact head are fully green. The prior local full-workspace wrapper omitted `pipefail`, so its terminal marker is still not accepted as proof of a green workspace suite; that local gate and the full-corpus performance pair remain open.

## Measurements

Same-machine fd pair using `/usr/bin/time -l kin init --json` on a fresh copy:

| corpus | before | after | observed change |
| --- | ---: | ---: | ---: |
| fd wall | 1550.71s | 1149.32s | -25.9% |
| fd user CPU | 702.34s | 641.06s | -8.7% |
| fd peak RSS | 1.964 GB | 1.964 GB | essentially flat |

The synthetic refit does **not** demonstrate the target scaling improvement: the `commits * files` user-time coefficient moved from `0.00041376` to `0.00049871` (about +20.5%), while only the fixed per-commit term improved. The ripgrep BEFORE run stopped during linking with an empty init result, so there is no valid before/after pair. No several-fold or broad performance claim is being made.

## Remaining, deliberately out of scope

- `commit_trees` is still materialized twice (plan + bind maps), so peak RSS keeps its `O(history x tree)` memory term (FIR-1601 finding 4).
- Link-context map construction is still rebuilt per commit; it is pure map-building over borrowed data (no deep clones, no resolution), so it is a small constant, but it keeps a residual `O(tree)` per-commit term.
- The FIR-1629 shallow-corpus decision (bounded admission mode vs corpus rebuild) is a policy call, not taken here.

### DIFF
--- a/crates/kin-core/src/env_registry.rs
+++ b/crates/kin-core/src/env_registry.rs
@@ -162,6 +162,7 @@ pub const OPERATIONAL: &[EnvVarSpec] = &[
     EnvVarSpec { name: "KIN_WRITE_VETO", kind: Kind::OneOf(&["warn", "enforce", "off"]), default: "warn", sensitivity: Sensitivity::Correctness, summary: "write-veto mode: 'warn' (default) annotates would-be vetoes into foreign-held scopes, 'enforce' rejects them with a 409, 'off' disables" },
     EnvVarSpec { name: "KIN_DAEMON_LOCATE_ONLY", kind: Kind::Bool, default: "false", sensitivity: Sensitivity::Correctness, summary: "daemon serves locate-only from a snapshot, changing what it answers" },
     EnvVarSpec { name: "KIN_DAEMON_DISABLE_LSP", kind: Kind::Bool, default: "false", sensitivity: Sensitivity::Correctness, summary: "disable LSP enrichment in the daemon, reducing relation coverage" },
+    EnvVarSpec { name: "KIN_HISTORY_LINK_VERIFY", kind: Kind::Bool, default: "false", sensitivity: Sensitivity::Operational, summary: "cross-check incremental historical enrichment against the full batch derivation at every commit during Git admission (self-audit; much slower, output unchanged)" },
     EnvVarSpec { name: "KIN_COCHANGE_MAX_FAN_OUT", kind: Kind::Usize, default: "15", sensitivity: Sensitivity::Correctness, summary: "cap on co-change fan-out per entity (co-change ranking signal)" },
     EnvVarSpec { name: "KIN_COCHANGE_MAX_FILES_PER_COMMIT", kind: Kind::Usize, default: "20", sensitivity: Sensitivity::Correctness, summary: "cap on files per commit considered for the co-change signal" },
 
@@ -590,6 +591,18 @@ impl AuditReport {
 /// invocation inside a kin-vfs projection would emit a spurious typo warning.
 const EXTERNAL_PREFIXES: &[&str] = &["KIN_VFS_"];
 
+/// `KIN_*` variables that existed in an earlier kin and were deliberately
+/// removed. Setting one is not a possible typo: the caller is pinning a
+/// guarantee this binary no longer implements, and silently ignoring the pin
+/// would let a protocol fail-safe rot into a no-op. These fail startup loudly
+/// in every validation mode so the caller learns the pin is dead.
+const REMOVED: &[(&str, &str)] = &[(
+    "KIN_INIT_WARM_CACHE",
+    "this variable was removed with the exact Git admission cutover: init derives all \
+     semantic truth from the captured object closure and consults no warm cache, so the \
+     contamination this pin guarded against cannot occur. Unset it",
+)];
+
 /// Audit an environment (any iterator of name/value pairs). Pure: it never reads
 /// or mutates the process environment, so it is deterministic and testable.
 ///
@@ -608,6 +621,14 @@ where
         }
         if EXTERNAL_PREFIXES.iter().any(|p| name.starts_with(p)) {
             continue; // owned by a sibling component (e.g. kin-vfs); not our surface
+        }
+        if let Some((_, reason)) = REMOVED.iter().find(|(removed, _)| *removed == name) {
+            report.invalid.push(Finding {
+                var: name.clone(),
+                severity: Severity::Error,
+                message: (*reason).to_string(),
+            });
+            continue;
         }
         match spec(&name) {
             None => report.unknown.push(name),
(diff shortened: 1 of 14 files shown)


troyjr4103:
This branch's admission-history and enrichment work has been superseded. Kin #1590 and #1600 rebuilt the same territory in crates/kin-git/src/admission_history.rs and crates/kin-index/src/history.rs, both merged after this PR opened. Closing as superseded; no action needed here.

PR #477, opened 2026-07-28
Automate protected releases and publish Kin v0.4.0

troyjr4103:
## Summary

- coalesce reviewed main drift into one protected release PR and auto-merge it after required checks
- mint the exact release tag automatically with the repository-scoped release App
- retry transient release failures twice and alert only after the bounded retries are exhausted
- publish an attested terminal release marker after GitHub Latest, install proof, npm, and GHCR promotion settle
- stage v0.4.0 for the breaking repository-v6 authority cutover

## Safety and liveness

- the automation branch is protected and accepts updates only from kin-release-bot or the founder
- main required checks are strict and up to date
- the release train regenerates only seven allowlisted files from trusted main
- no later tag can be admitted until the previous stable release is terminally proven
- v0.3.6 is the sole explicit pre-marker migration fallback; markerless v0.4.0+ releases fail closed

## Migration to v0.4.0

A Kin binary accepts exactly one on-disk graph snapshot format. v0.3.6 wrote
format 12 and v0.4.0 accepts only format 13, so no repository initialized by
v0.3.6 opens under v0.4.0; each must be re-initialized against its source of
truth. This is the standing pre-release storage posture rather than something
this release introduces: v0.3.6 accepted only the format it wrote, and so does
this one. The accepted range widens when the format is declared stable.

## Verification

- actionlint on CI, release train, tag, recovery, and release workflows
- release authority policy test
- 13 release generator/version tests
- npm automatic-release and public-verification policy suites
- @kinlab/kin: 33 tests and lint
- @kinlab/kin-mcp: 13 tests and lint
- cargo check --locked --workspace
- cargo fmt --all -- --check
- cargo metadata --locked --no-deps
- coherent v0.4.0 release-intent and origin/main minor-version gate
- release-tag admission replayed against live merge-queue landings on main
- release-workflow authority suite, including the executable gate fixtures
- cargo fmt, clippy on the CI allowlist, build --all-targets, test --workspace,
  and doc tests on the merged tree

The merge is intended to trigger the first automatic tag and release. Public and hosted convergence will be verified separately; this PR does not claim those external surfaces before they complete.

Part-of FIR-1015.

PR #480, opened 2026-07-28
Move the Windows release leg off the pull-request path and scope CI caches to trusted refs

troyjr4103:
## What

Two changes, both aimed at the wall clock a pull-request round costs the whole fleet.

**1. The Windows installer leg leaves the pull-request path.** `Windows installer + vector-free release build` cold-compiles the entire dependency graph in the release profile for `x86_64-pc-windows-msvc`. Measured at 20.5 and 24.6 min on runs 30390278470 and 30390270908, longer than any other job, so it set the floor for every land-and-iterate cycle. It now runs on push to main and on the `dependency-updated` dispatch, and reports skipped on a pull request.

This job already ran on every push to main before this PR, so main-push wall clock is unchanged. The entire saving is on the pull-request path.

Release gating is pinned in two places. `release-tag.yml` names this exact check in `REQUIRED_CHECKS` and refuses to mint a tag unless it is present and green on the release sha. The load-bearing fact underneath that is the `changes` classifier computing `docs_only` only inside the structurally matched `pull_request` shell guard, so `docs_only` is false on main and the installer cannot skip there. The authority test now finds the matching `fi`, requires the only truthy assignment inside it, and self-falsifies by placing a second truthy assignment after it.

**2. Cargo and Docker caches restore anywhere, save only from main.** Measured on 2026-07-28 the repository held 12.1 GB across 10 entries, eight of them 1.2 to 1.9 GB pull-request entries, and not one cargo cache on `refs/heads/main`.

A cache entry written from `refs/pull/N/merge` is readable by later runs on that same ref, so the old configuration did warm pushes 2..N of a single pull request. It also spent the shared repository cache budget on entries no other branch could read, evicting the main entry every pull request can restore from its first push. This is a tradeoff: give up the intra-branch warm start in exchange for a shared one. Because strict required-status-check policy forces a re-push before merge, the intra-branch restore was load-bearing at land time, which is the leading explanation for the `Check & Test` regression below.

`Check & Test` (both platforms), `Windows authority tests`, and `Linux Daemon Smoke` use `Swatinem/rust-cache` pinned by commit SHA with `save-if: ${{ github.ref == 'refs/heads/main' }}`. The smoke job keeps registry-only scope via `cache-targets: false`. `Docker Image Build (no push)` keeps `cache-from` everywhere and exports `cache-to` only from main.

The justification for save scoping is budget, not a new trust boundary. GitHub already prevents a pull-request merge-ref cache from being restored by main or a sibling pull request. The structural authority check now binds exactly one main-only `save-if` to the `with:` mapping of every live rust-cache step; comments, unrelated fields, and an `env.save-if` in the same action step cannot compensate for a missing input.

## Scope boundary

This is not a repository-wide “no PR cache writes” invariant.

- All `Swatinem/rust-cache` uses are covered structurally.
- Docker buildx import/export is covered.
- Three `actions/cache` uses remain outside that invariant. Windows installer and coverage are main-only by job admission. `fuzz.yml` can still write a cargo/fuzz cache on a qualifying parser pull request.

FIR-1644 therefore stays open through the post-merge evidence below and until its broader acceptance wording is reconciled with the deliberately retained fuzz cache.

## Files

- `.github/workflows/ci.yml`: installer admission and cargo cache steps
- `.github/workflows/docker.yml`: conditional `cache-to`
- `.github/workflows/daemon-smoke.yml`: main-only rust-cache save and no stale restore fallback
- `scripts/test-release-workflow-authority.py`: release/check linkage plus structural shell/YAML authority and self-falsification

## Authority test

The existing 15-probe sandbox matrix still covers installer admission, release-tag required-check presence, action pinning, Docker export, main trigger presence, daemon cache scope, and the original classifier mutations.

Four additional counterexamples now execute in memory on every authority-test run and must fail for the intended reason:

| counterexample | caught |
| --- | --- |
| `docs_only=true` inserted after the pull-request guard's matching `fi` | yes |
| missing rust-cache `save-if` compensated by a YAML comment | yes |
| missing rust-cache `save-if` compensated by an unrelated active field | yes |
| rust-cache `save-if` moved from `with:` to `env:` in the same step | yes |

`actionlint` is clean on every workflow touched by this PR. Whole-tree `actionlint` still reports the pre-existing `SC2034` warning in `registry-index-migrate.yml`; that file's blob is identical at this PR's base and head and is outside this focused repair.

## Measured PR path

Before: runs 30390278470 / 30390270908 (cold). Docker from #467. After: runs 30393521508 / 30393521700 / 30393521545, also cold because no main-side entry exists until this merges.

| job | before | after | delta |
| --- | --- | --- | --- |
| Windows installer + vector-free release build | 20.5 / 24.6 min | **skipped, 0 min** | off the PR path |
| Check & Test (macos-latest) | 16.0 / 16.2 min | 18.6 min | **+2.5 min** |
| Check & Test (ubuntu-latest) | 13.5 / 13.0 min | 14.8 min | **+1.6 min** |
| Windows authority tests | 13.5 / 14.4 min | 11.2 min | -2.7 min |
| Docker Image Build (no push) | 14.1 min | 9.7 min | -4.4 min |
| Linux Daemon Smoke | 3.0 min | 2.5 min | -0.5 min |

**Critical path: about 20.5 to 24.6 min before, 18.6 min after.**

The Windows and Docker cold-run gains are consistent with removing cache export work billed inside job wall clock. `Check & Test` regressed on both platforms. That regression is real, n=1 per arm, and not explained away here.

## Post-merge obligation — FIR-1644 remains open

Once main has run and written entries:

1. Inspect `gh cache list` first. The repository's configured eviction limit is live-verified at 10 GB, but usage can temporarily exceed it before LRU eviction. A main push creates four runtime rust-cache entries (two `check` matrix arms, Windows authority, and daemon registry-only) plus Docker blobs and the existing main-only caches, so main-entry stability must be observed rather than assumed.
2. Confirm main-scope entries exist and qualifying pull requests restore them.
3. Take the median warm `Check & Test (macos-latest)` across five qualifying same-arm pull-request runs, excluding `dependabot/*` because they change `Cargo.lock`.
4. Roll back the `check` job's rust-cache step if that median is `>= 16.0 min`, the best old-configuration duration.

Rollback is narrow: delete only the rust-cache step from the `check` job and leave the other jobs alone. The authority test enumerates actual rust-cache steps rather than enforcing a hardcoded count, so the remaining steps still have to bind their own main-only save policy.

## Merge order with #477

Land #480 before #477. The two heads are each mergeable against current main but not mutually compatible without a rebase. After #480 lands, #477 must preserve the structural helpers, resolve the shared `RELEASE_TAG` constant once, update the required-check error needle to its renamed `missing required GitHub Actions check` text, and rerun the full authority gate. The detailed bidirectional collision proof remains in the PR discussion.

## Exact-head verification

At `1702be7ab5be27d1ca73db7064d17eeab988d78b` after a signed merge of main `40a39029e683e5fbc86e81aee16af3210dacfe58` (PR #479):

- embedded classifier/cache falsifications: green
- external rerun of the same adversarial mutations: rejected for the intended structural errors
- `python3 scripts/test-release-workflow-authority.py`: green
- relevant-workflow `actionlint`: green
- `ruff check` and `ruff format --check`: green
- Python compile: green
- both npm launcher test/lint suites: 46 tests green
- install/VFS/parity/npm release-policy script suite: green
- `git diff --check`: green
- net diff against current main remains exactly the four intended CI/authority files

## Not claimed

This PR's own timings do not demonstrate a warm-cache win. FIR-1573's acceptance is a Windows job p50 under 10 minutes on a warm cache; this relocates that job rather than meeting the duration target, so FIR-1573 also stays open.

Refs FIR-1644

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,12 +108,14 @@ jobs:
           targets: x86_64-pc-windows-msvc
 
       # Restore on every event, save only from main. A cache entry written from
-      # a pull request ref is readable by that one pull request and nothing
-      # else, so saving from pull requests bought no warm start while consuming
-      # the shared repository cache budget and evicting main's entry, which is
-      # why no reusable cargo cache existed on refs/heads/main at all. Saving
-      # only from main keeps one warm entry per job that every pull request
-      # restores, and leaves fork pull requests with no path to write one.
+      # a pull request ref is readable by later runs on that SAME ref and by
+      # nothing else, so the old configuration warmed pushes 2..N of one pull
+      # request while spending the shared repository cache budget and evicting
+      # main's entry, which is why no reusable cargo cache existed on
+      # refs/heads/main at all. Saving only from main trades that intra-branch
+      # warm start for one entry every pull request restores from push 1.
+      # The tradeoff is real and measured: see the trusted-ref note in
+      # scripts/test-release-workflow-authority.py.
       - name: Cache cargo registry and build
         uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -325,14 +325,14 @@ jobs:
     outputs:
       docs_only: ${{ steps.classify.outputs.docs_only }}
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 0
       - name: Classify changed paths
         id: classify
         # This step gates release-critical jobs. Override workflow defaults and
         # shell startup injection rather than inheriting mutable semantics.
-        shell: /usr/bin/bash --noprofile --norc -e -u -o pipefail {0}
+        shell: /usr/bin/bash --noprofile --norc -p -e -u -o pipefail {0}
         env:
           BASH_ENV: ""
           EVENT_NAME: ${{ github.event_name }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,18 +107,17 @@ jobs:
         with:
           targets: x86_64-pc-windows-msvc
 
+      # Restore on every event, save only from main. A cache entry written from
+      # a pull request ref is readable by that one pull request and nothing
+      # else, so saving from pull requests bought no warm start while consuming
+      # the shared repository cache budget and evicting main's entry, which is
+      # why no reusable cargo cache existed on refs/heads/main at all. Saving
+      # only from main keeps one warm entry per job that every pull request
+      # restores, and leaves fork pull requests with no path to write one.
       - name: Cache cargo registry and build
-        uses: actions/cache@v6
+        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            target
-          # A key distinct from the installer job: this builds the debug profile,
-          # and one shared key would have the two jobs overwrite each other.
-          key: windows-authority-cargo-${{ hashFiles('Cargo.lock') }}
-          restore-keys: |
-            windows-authority-cargo-
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Compile and run native Windows authority tests
         shell: bash
@@ -146,7 +145,19 @@ jobs:
   windows-installer:
     name: Windows installer + vector-free release build
     needs: changes
-    if: ${{ !cancelled() && needs.changes.outputs.docs_only != 'true' }}
+    # Deliberately off the pull-request path. This leg cold-builds the whole
+    # dependency graph in the release profile for x86_64-pc-windows-msvc and
+    # measured 20 to 25 minutes per round, longer than any other job, so it set
+    # the floor for every land-and-iterate cycle. It now runs on each push to
+    # main, which is exactly the commit release proof reads: release-tag.yml
+    # names this check in REQUIRED_CHECKS and refuses to mint a tag unless it is
+    # present and green on the release sha. On a pull request the job reports
+    # skipped, which satisfies the required context the same way the
+    # documentation-only path already relies on.
+    if: >-
+      ${{ !cancelled()
+      && github.event_name != 'pull_request'
+      && needs.changes.outputs.docs_only != 'true' }}
     runs-on: windows-latest
     timeout-minutes: 60
     steps:
@@ -398,16 +409,13 @@ jobs:
           grep -q '^\[registries\.kin\]' .cargo/config.toml
           if grep -q '^\[patch\.kin\]' .cargo/config.toml; then echo "::error::[patch.kin] git pins must be removed — kin-* resolve from the Kin registry"; exit 1; fi
 
+      # Same trusted-ref scoping as the Windows authority job: restore always,
+      # save only from main. This is the pull-request critical path, so a warm
+      # dependency graph here is what a round actually costs.
       - name: Cache cargo registry and build
-        uses: actions/cache@v6
+        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            target
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-cargo-
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Check formatting
         run: cargo fmt -- --check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -330,11 +330,17 @@ jobs:
           fetch-depth: 0
       - name: Classify changed paths
         id: classify
-        # This step gates release-critical jobs. Override workflow defaults and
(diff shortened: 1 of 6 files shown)


troyjr4103:
## Measured, all checks green on `c370fdd6`

Before: dependabot runs 30390278470 / 30390270908 (cold). Docker from PR #467.
After: runs 30393521508 / 30393521700 / 30393521545, **also cold** (no main-side cache exists until this merges).

| job | before | after | delta |
| --- | --- | --- | --- |
| Windows installer + vector-free release build | 20.5 / 24.6 min | **skipped, 0 min** | off the PR path |
| Check & Test (macos-latest) | 16.0 / 16.2 min | 18.6 min | **+2.5 min** |
| Check & Test (ubuntu-latest) | 13.5 / 13.0 min | 14.8 min | **+1.6 min** |
| Windows authority tests | 13.5 / 14.4 min | 11.2 min | -2.7 min |
| Docker Image Build (no push) | 14.1 min | 9.7 min | -4.4 min |
| Linux Daemon Smoke | 3.0 min | 2.5 min | -0.5 min |

**Critical path: ~20.5 to 24.6 min before, 18.6 min after.** Real, but smaller than the installer's own duration, because Check & Test macOS was already close behind it and becomes the new pole.

### Why Windows authority and Docker got faster on a cold run

Not a cache hit. It is the removal of the **export**. The old config uploaded a 1.3 GB target cache (Windows) and a `mode=max` layer export (Docker) in the post step of every round, and that upload is billed inside the job's wall clock. Deleting a write nothing could ever read is a pure subtraction, available immediately.

### The regression I am not explaining away

Check & Test went **up** on both platforms. I cannot attribute it with confidence: n=1 per arm on hosted runners, the before-runs are a different commit (a dependabot lock bump), and rust-cache's restore does a `cargo metadata` + lookup that buys nothing cold. Cutting the other way, these jobs also stopped uploading ~1.9 GB, which by the Windows/Docker reasoning should have made them faster.

So it is a real number and I am not discarding it. **This needs the post-merge measurement**, and the rollback lever is narrow: revert only the two rust-cache steps in the `check` job. The installer move, the Docker export change, and the Windows/smoke scoping stand on their own.

### Mechanism verified in the live run, not assumed

From the daemon-smoke log: `save-if` correctly resolved to `false` on the PR and the post step saved nothing; key namespace is per job (`v0-rust-daemon-smoke-Linux-x64-c15cfb4b`), so the four cached jobs cannot overwrite each other; `cache-targets: false` was honored (no target dir in Cache Paths); and `cargo metadata` against the kin sparse registry works inside rust-cache's restore step, which was the one integration risk in the swap.

### For the captain

PR #477 (queue head) also touches `ci.yml`, `release-tag.yml`, and `scripts/test-release-workflow-authority.py`. Its ci.yml hunks are in `npm-launchers` and an inserted step in `check`; mine are the two cache steps and the installer `if`. Different regions, so a clean auto-merge is likely but not guaranteed. **If 477 lands first and this goes non-mergeable, it needs a rebase, and the authority test must be re-run after any rebase** since 477 edits that same file.

troyjr4103:
## For the merge captain: PR 480 and PR 477 break each other in BOTH orders

Supersedes my earlier comment on this thread, which documented only the "477 first" direction. The hazard is bidirectional. Whoever lands second goes red on the same assertion, and a clean textual merge does not surface it, because the conflict and the breaking assertion are in different regions of different files.

**Reviewer recommendation: land 480 first, let 477 absorb the rebase.** 480 is 4 files with a one-constant conflict surface. 477 is 20 files and rewrites `release-tag.yml` wholesale. Moving the rebase cost onto 477 is the cheaper direction, and 477 has to re-run its full gate after any rebase regardless. The merge-order call is the captain's; this comment only makes both paths safe.

Note `mergeable: MERGEABLE` on either PR means "against main", not "against each other". Do not read it as compatibility.

### What collides

`git merge-tree --write-tree` on the two heads: `.github/workflows/ci.yml` **auto-merges clean** (nearest hunk gap is 8 lines, above git's 3-line context). Exactly one file conflicts, `scripts/test-release-workflow-authority.py`, at the module constants, because both PRs add a `RELEASE_TAG = WORKFLOWS / "release-tag.yml"` line at the same anchor.

The conflict is trivial. The **semantic** break is the real hazard: 477 rewrites release-tag.yml's verification loop and renames the error string that 480 pins.

| assertion 480 pins | against 477's release-tag.yml |
| --- | --- |
| `REQUIRED_CHECKS: \|` | present |
| all six required context names | present |
| `required check not green: {name}` | present |
| `missing required check: {name}` | **ABSENT, renamed** |

### If 480 lands first (recommended)

**477 rebases** onto the new main and must:

1. Resolve the constants block by keeping **both** sides' constants. 477's own set is a superset except that 480 contributes nothing extra, so 477's block stands as-is; just do not drop it.
2. Update **480's** assertion to match 477's own renamed string, since 477 is the PR changing it:
   ```python
   -        "missing required check: {name}",
   +        "missing required GitHub Actions check: {name}",
   ```
3. Not introduce a duplicate `RELEASE_TAG` constant or a second `release_tag = RELEASE_TAG.read_text(...)` binding. Both are legal Python and would shadow silently rather than error.
4. Re-run the full kin CI gate, including `python3 scripts/test-release-workflow-authority.py`.

### If 477 lands first

**480 rebases** and applies the same two edits: take 477's constants block verbatim (strict superset, it already defines `RELEASE_TAG`), then the identical one-line needle update above. Verified end to end: merged tree extracted, both edits applied in a sandbox, authority test run to a pass printing 477's updated summary line.

### Compatibility, in the good direction

477 tightens required non-DCO checks to conclude exactly `success`, where `skipped` was previously acceptable. That is strictly stronger than what 480 relies on and it adds defense in depth to the installer relocation: 480's `skipped` conclusion only ever appears on `pull_request` events, and `release-tag.yml` only ever reads the release sha, which is a main commit where the job genuinely runs. The two PRs are semantically complementary. Only the assertion text collides.

PR #482, opened 2026-07-29
Read entity body indentation absolutely and surface session expiry

troyjr4103:
FIR-1598 measured an unscripted agent losing 2/2 to a grep agent at 3.4x the cost, and root-caused it to four defects on the transaction surface. Four of the six items on the ticket already had landed fixes on `main`; this verifies each of those in the source and locks it behind a FIR-1598-shaped test, and fixes the two that were still open.

## Per defect: mechanism and evidence

**1. Transaction poisoning. Already fixed on `main`; regression-locked here.**
`unstage_failed_attempt` clears the whole staged set on any commit failure landing before the publication fence, names every operation it dropped, and says corrected operations can be staged on the same transaction. Every pre-fence failure path routes through it. Added `three_file_signature_change_survives_a_mid_flight_staging_error` and `ambiguous_name_target_lists_candidates_and_the_id_retry_commits`, which assert the FIR-1598 shape (ambiguous bare name mid-flight, corrected id retry on the same transaction) rather than the unresolvable-name shape the existing test covers.

**2. Session TTL. Half fixed on `main`; the other half fixed here.**
Already landed: the idle TTL is 1800s (it was two 30-second heartbeat intervals, which is the measured ~150s), and `kin_session_heartbeat` is in `agent_default_tool_names`. Still open: nothing surfaced the deadline, so an agent could not decide to heartbeat before a read phase outlasted it, and a reaped session answered with a bare not-found at HTTP 500. Session start and heartbeat now report `idle_timeout_secs` and `expires_at` computed from the session's own `last_heartbeat` plus the coordinator's live TTL; a call on an ended or idle-reaped session answers 404 with a message naming the expiry, the env var that sets the window, and `kin_session_start` as the recovery. `SessionRegistry::begin_transaction` says the same. Both tool descriptions now point at the fields.

**3. Ambiguous name targets. Mostly fixed on `main`; sharpened here.**
The candidate list already carried entity ids and `file:line`. Each candidate now also carries its one-line declaration, bounded at 160 chars: a location says where a candidate lives, a declaration says what it is, and that is what the caller is choosing between. The acceptance half with no test, that an unambiguous id retry works on the transaction the caller already holds, is now covered.

**4. impl-method indent. The real remaining defect, fixed here.**
An entity span is the tree-sitter node range, so it opens at the entity's first token. A nested entity therefore carries its indentation in the file *ahead* of the span while every line of its body after the first carries indentation *inside* it. Splicing a caller-authored body verbatim read line 1 as relative to the entity's column and lines 2..n as absolute, so a body written back as the file renders it landed at twice the indentation: it compiled and failed `cargo fmt`, which is worse than a refusal.

`kin_projection::entity_body_splice` reads line 1 absolutely too. When the span is preceded on its line by nothing but whitespace and the body opens with exactly that whitespace, the splice extends back over the file's copy so the body's own indentation lands at column 0 of the entity's line. A body that does not open with the entity's indentation is spliced verbatim, which covers both the exact span slice the read surface serves and a deliberate re-indentation. The exact MCP commit planner and `splice_entity` both route through it; `reconstruct_file` deliberately does not, because it runs the graph-to-file direction where a body is the exact bytes of its own region.

**5 and 6. Already fixed on `main`, verified in source.** `get_context_pack`'s description names `focal_entity.body` (with a test pinning the string), and `ACCEPTED_OPERATION_SHAPES` replaces the raw serde field error on both the non-array and the deserialization refusal.

## Falsification

The indent fix was falsified, not just tested. With `entity_body_splice` reverted to a verbatim splice and nothing else changed:

- `kin-projection --lib`: 4 failures — `impl_method_body_carrying_its_indentation_is_not_double_indented`, `nested_module_function_body_carrying_its_indentation_is_not_double_indented`, `deliberate_reindentation_is_preserved`, `tab_indentation_is_matched_exactly`.
- `kin-daemon --lib mcp_commit`: `nested_entity_bodies_commit_at_their_own_indentation` fails, producing `impl Builder {\n        pub fn set(...)` where the intended source is `impl Builder {\n    pub fn set(...)`. That is byte-for-byte the reported symptom, an impl method re-indented from 4 spaces to 8.
- `span_slice_body_for_a_nested_entity_still_commits_exactly` passes with and without the fix, which is the evidence the normalization did not trade one indent bug for another.

All pass with the fix restored.

## Acceptance rehearsal

`three_file_signature_change_survives_a_mid_flight_staging_error` runs the measured shape through the real repository-authority commit path: session, transaction begin, a three-file signature change with an impl-nested caller and a module-nested caller, one deliberate ambiguous bare-name operation mid-flight, a refusal carrying the candidate ids and saying the operations were cleared, repository authority unmoved, then the corrected id retry on the SAME transaction committing with `ops_applied: 3` and all three files asserted byte-for-byte against the intended source.

## Not claimed

- The agent-arm comparison has **not** been rerun. Nothing here claims the 2/2 loss or the 3.4x cost has flipped.
- The rehearsal drives `commit_exact_transaction`, the same entry point MCP `kin_transaction_commit` reaches after `mcp_session_registry_snapshot`, but it does not exercise the `kin mcp start` stdio JSON-RPC transport or `semantic_locate`. The session-expiry test drives the axum router directly rather than a spawned daemon.

## API change for review

`kin_projection::splice_entity` gained a leading `original: &[u8]` parameter — the replaced region cannot be decided without the file bytes, so the signature now forces callers to supply them. `entity_body_splice` is newly exported.

---

## Follow-up commit: a way out of a transaction

`agent_default_tool_names` carried `kin_transaction_begin`, `_stage`, and `_commit` but not `kin_transaction_abort`, so an agent on that profile could only push a transaction forward and had no clean exit from one it had decided against. That is the other half of "nothing half-applies". Added, with the profile test asserting it stays.

The write-path tool descriptions now carry the contract the surface actually enforces, because those descriptions are what an unscripted agent reads:

- `kin_transaction_stage` steers to the entity id the read tools already returned, says a bare name resolves only when unique and that an ambiguous one comes back with the candidate ids, and states the body contract: the entity exactly as its file renders it, first-line indentation read as the entity's own line indentation rather than as indentation on top of it.
- `kin_transaction_commit` says a refusal before publication leaves the transaction usable with its operations cleared and named, so corrected operations go on the SAME transaction.
- `kin_transaction_abort` says when to reach for it and that recovering from a refused commit is not one of those times.
- `kin_session_start` and `kin_session_heartbeat` point at `idle_timeout_secs` / `expires_at` and at the expiry recovery.

`project_to_bytes` splices caller-authored bodies, so it now reads first-line indentation the same way the rest of the projection surface does.

Closes FIR-1598

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -326,6 +326,14 @@ struct SessionStartResponse {
     started_at: kin_model::timestamp::Timestamp,
     capabilities: SessionCapabilities,
     status: String,
+    /// How long the session may go idle before the sweeper may reap it.
+    idle_timeout_secs: u64,
+    /// When that window closes if nothing refreshes it. Every session-bound
+    /// daemon call refreshes the heartbeat, so this only arrives on a genuinely
+    /// silent agent — but an agent that cannot see the deadline cannot decide to
+    /// send `kin_session_heartbeat` before it, which is how a long read phase
+    /// silently stranded a transaction.
+    expires_at: kin_model::timestamp::Timestamp,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -333,6 +341,10 @@ struct SessionHeartbeatResponse {
     session_id: String,
     status: String,
     heartbeat_at: kin_model::timestamp::Timestamp,
+    /// Same contract as on [`SessionStartResponse`]: the refreshed window, and
+    /// the deadline it now runs to.
+    idle_timeout_secs: u64,
+    expires_at: kin_model::timestamp::Timestamp,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -1823,6 +1835,8 @@ async fn start_session(
             )
         })?;
 
+    let idle_ttl = state.coordinator.session_idle_ttl();
+    let expires_at = session_expiry(&session.last_heartbeat, idle_ttl);
     Ok(Json(SessionStartResponse {
         session_id: session.session_id.to_string(),
         vendor: session.vendor,
@@ -1831,7 +1845,46 @@ async fn start_session(
         started_at: session.started_at,
         capabilities: session.capabilities,
         status: "active".to_string(),
+        idle_timeout_secs: idle_ttl.as_secs(),
+        expires_at,
     }))
+}
+
+/// What to tell a caller whose session id no longer names a live session.
+///
+/// The id is well-formed, so this is an ended or idle-reaped session rather
+/// than a typo, and the only move is a fresh `kin_session_start`. Saying so is
+/// the difference between an agent restarting in one call and an agent
+/// retrying the same dead id until it times out.
+fn expired_session_message(session_id: &SessionId) -> String {
+    format!(
+        "session not found: {session_id}. It was ended or expired after its idle timeout \
+         (KIN_SESSION_IDLE_TTL_SECS). Call kin_session_start for a new session id, then \
+         re-open any transaction on it; every session-bound call refreshes the idle window, \
+         and kin_session_heartbeat refreshes it during a long read phase."
+    )
+}
+
+/// The instant an idle session stops being refreshable, from the heartbeat it
+/// last recorded.
+///
+/// Saturates rather than wrapping: a TTL large enough to overflow the calendar
+/// means "not reachable from here", and reporting the heartbeat itself as the
+/// deadline would tell an agent its live session had already expired.
+fn session_expiry(
+    last_heartbeat: &kin_model::timestamp::Timestamp,
+    idle_ttl: Duration,
+) -> kin_model::timestamp::Timestamp {
+    let window = i64::try_from(idle_ttl.as_secs())
+        .ok()
+        .and_then(chrono::TimeDelta::try_seconds)
+        .unwrap_or(chrono::TimeDelta::MAX);
+    kin_model::timestamp::Timestamp(
+        last_heartbeat
+            .0
+            .checked_add_signed(window)
+            .unwrap_or(chrono::DateTime::<chrono::Utc>::MAX_UTC),
+    )
 }
 
 /// GET /session/{session_id} — fetch a single active session.
@@ -1859,6 +1912,18 @@ async fn session_heartbeat(
     State(state): State<Arc<DaemonState>>,
 ) -> Result<impl IntoResponse, (StatusCode, String)> {
     let session_id = parse_session_id(&session_id)?;
+    // A heartbeat against a session the sweeper already reaped is the one call
+    // guaranteed to be in flight when a long-lived agent loses its session. It
+    // must come back as an actionable 404 naming the restart, not as a 500 that
+    // reads like a daemon fault.
+    if state
+        .coordinator
+        .get_session(&session_id)
+        .map_err(internal_error)?
+        .is_none()
+    {
+        return Err((StatusCode::NOT_FOUND, expired_session_message(&session_id)));
+    }
     state
         .coordinator
         .heartbeat(&session_id)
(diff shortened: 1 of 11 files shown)


troyjr4103:
HOLD FOR ONE FOLLOW-UP COMMIT before merge (independent review verdict: no P0, splice fix verified under 12 edge shapes, but one blocking P1): abort_transaction has no state guard and this PR makes it newly reachable from the agent profile; aborting a transaction in the committing state strands it past the durable publication fence, defeating the documented crash-recovery resume. Incoming commit: guard abort on active or validated states with a test, make abort actually discard staged operations as its description claims, scope the session-expiry description sentences to the daemon-backed path, and caveat the stage-description read-tool steer for the 40-line render cap. Fix lane dispatched; do not merge until it lands and checks re-run.

troyjr4103:
HOLD LIFTED, pending green checks. Follow-up commit `329b53f8` "Guard transaction abort behind its editable states" is pushed; CI re-triggered on that head.

**Every item on the hold list is addressed.**

`abort_transaction` (`crates/kin-mcp/src/session.rs`) now guards on `matches!(tx.state.as_str(), "active" | "validated")`, mirroring `clear_staged_operations`. A `committing` transaction refuses with a message naming the state and the call that owns it, saying its commit is already fenced so authority may have moved and to re-send `kin_transaction_commit`, which resumes the fenced payload idempotently. A terminal transaction refuses separately, because relabelling a `committed` receipt `aborted` records provenance that never happened. Abort now also clears `staged_operations`, so the "discard all staged mutations" claim is true.

**Tests: four in `kin-mcp`, two in `kin-daemon`, none existed before.**

`abort_discards_the_staged_mutations_it_says_it_discards`, `abort_is_refused_once_a_commit_is_fenced_and_the_commit_stays_resumable`, `abort_is_refused_on_a_terminal_transaction`, `abort_of_an_unknown_transaction_names_the_id` in `session.rs`. The fenced test does not stop at the refusal; it then re-enters `prepare_transaction_commit` with the same digest, asserts the resume returns `committing` with the fenced payload intact, and commits, which is the invariant the guard exists to protect.

Through the real repository-authority path in `mcp_commit.rs`: `aborting_a_staged_transaction_moves_neither_authority_nor_the_working_tree` (authority `roots.generation` unmoved, `src/lib.rs` byte-identical, staged set empty, session still usable) and `a_transaction_whose_commit_was_refused_can_still_be_aborted`, which drives a real refused commit so the guard is proven to accept the state `unstage_failed_attempt` leaves behind.

**Falsified, not just written.** With `abort_transaction` reverted to its pre-fix body and nothing else changed: three of the four `kin-mcp` tests fail (discard, fenced-refusal, terminal-refusal) and the daemon abort test fails on "abort must discard the staged mutations it says it discards". `abort_of_an_unknown_transaction_names_the_id` and `a_transaction_whose_commit_was_refused_can_still_be_aborted` pass either way, which is the evidence the guard did not narrow the recovery path it was supposed to leave open.

**Descriptions.** `SESSION_START_DESC` and `SESSION_HEARTBEAT_DESC` now scope `idle_timeout_secs` / `expires_at` and the expiry-recovery sentence to the daemon-backed response and state what the in-process branch returns instead. `TRANSACTION_STAGE_DESC` gains the P2-4 precondition (the indentation read is a byte-exact prefix match against the file's own indent run, so copy the file's whitespace) and the P1-3 caveat (source rendered by `get_entity_source` / `get_context_pack.focal_entity.body` is capped at 40 lines or 2400 characters and marked `... [truncated]`, so a truncated body must not be staged as the entity's full source).

**One thing found while landing this, worth the reviewer's attention.** All five `TRANSACTION_*_DESC` constants in `handlers/sessions.rs` are referenced nowhere else: `tool_definitions()` supplies its own inline literals for the five transaction tools, while the seven session tools do reference their constants. They have been unwired since `895d1503`, and `pub` in a `pub` module suppresses any dead-code warning. So the transaction-description rewrites in this PR, and the review's own P1-3 and P2-4 description fixes, currently do not reach an agent. I updated the live `kin_transaction_abort` literal in `tools.rs` in place, because my state guard changed the behavior that string describes and leaving it would have shipped a fresh doc-versus-code mismatch. I deliberately did not wire the other four, to keep this commit narrow on a PR that is already reviewed.

**Follow-ups: FIR-1660** carries the review's seven P2s (tabs de-indent overclaim, `expires_at` never firing for PID-carrying sessions, the self-referential TTL test, the heartbeat TOCTOU 500, the inverted-range splice guard asymmetry, `project_to_bytes` vs `reconstruct_file` with kin-vfs unchecked, `LayoutNotFound` misuse) plus the unwired-descriptions defect above. Milestone M8, priority Medium.

**Gates**, exit codes captured explicitly: `cargo fmt --all -- --check` exit 0. ci.yml-mirror clippy (`--all-targets --all-features` with the ci.yml allowlist, `RUSTFLAGS=-D warnings`) on `kin-mcp`, `kin-daemon`, `kin-projection`: exit 0. `cargo test -p kin-mcp -p kin-daemon -p kin-projection`: exit 0, 247 + 473 + 96 + 13 passed, 0 failed. `cargo test --workspace` was still running when this comment was posted and is NOT recorded here as green. All runs under `GIT_CONFIG_GLOBAL=/dev/null` in the lane's own `CARGO_TARGET_DIR`. CI on this PR is the enforcing gate.

PR #483, opened 2026-07-29
Report graph health and admission enrichment from graph truth

troyjr4103:
Two first-run truth defects shared one underlying mistake: user-facing status surfaces were not reporting from the graph authority that repository-v6 admission actually populated.

- `kin graph status` treated an empty derived facet layer as a critical defect even when the admitted entity graph was healthy.
- `kin init` and `kin status` reported enrichment as absent because they read raw authority tables, while admission persists semantic deltas that must be materialized at the selected workspace generation.

This PR closes FIR-1631 and FIR-1592 by making those reports graph-backed, generation-bound, and explicit about what they do and do not attest.

## Durable `init` and `status` truth

`status::semantic_enrichment_from_authority` now derives enrichment by replaying first-parent semantic authority through the repository generation and applying the workspace overlay. Entity and relation counts are therefore workspace-scoped; `semantic_change_count` remains repository-wide.

The durable status payload is now `kin.status.v2`. Its parser rejects:

- unknown authority identifiers
- `completion_attested: true`
- inconsistent present/absent count tuples
- repository, root-bundle, enrichment-authority, or workspace generation mismatches
- unsupported root schema versions
- unknown top-level or semantic-enrichment fields

`kin init --json` now emits `kin.init-result.v5` and carries the same `semantic_enrichment` object as `kin status --json`, instead of the constant `not-run`.

The README and quickstart now match the landed product behavior: admission derives semantic entities but not vectors, safely representable Git coexistence settings are sealed during admission, and embedding coverage belongs to `kin graph status` rather than durable `kin status`.

## Healthy admission no longer fails graph health

Repository-v6 admission binds entities and relations, while file-layout facets are derived later by reconcile/projection paths. Missing facets are therefore reported as pending enrichment, with counts, rather than as a critical defect.

Actual contradictions still fail closed:

- diverged tree identity
- conflicting or stale facets
- content-identity mismatch
- orphan entities
- contamination

Content identity is also checked before publishing a file-layout facet, preventing a changed artifact from being described by stale layout metadata.

## MCP graph status now reports one selected graph observation

The integrated temporal-scope model is preserved: `kin_graph_status` reports the selected HEAD graph or the exact temporal session graph, rather than silently substituting an unscoped HEAD view.

The daemon captures entity count, relation count, and embedding counters as one point-in-time selected-graph observation:

1. sample the graph-authority epoch
2. acquire the embedding-work fence without blocking behind active mutation
3. capture all counters from the selected graph
4. release the fence before any await
5. revalidate the epoch and temporal graph identity
6. retry bounded graph churn, or fail retryably

The MCP handler only formats those captured counters; it cannot reread the graph and create a mixed observation.

`kin.graph-status.v1` is strict on deserialization. It rejects unsupported schema/view/scope/authority/sampling/completion values, impossible embedding tuples, and selected-report/envelope disagreement. The stdio boundary strips any daemon-supplied `_kin` metadata and attaches an envelope derived from the selected report itself; it deliberately omits unscoped HEAD generation, reconciliation, and health metadata. Direct daemon responses omit that stdio-only envelope.

The selected-graph route is composed with the current daemon spawn/revival contract: a connection-loss retry preserves the repository/session arguments and re-enters the same strict response validation.

## Hermetic test fixtures

Exact-head testing exposed several test-only Git subprocesses that inherited machine-global hook and configuration policy. Fixture Git commands now disable system configuration and bind global configuration to the platform null device. Production Git behavior is unchanged: the two production helpers reached indirectly by unit tests apply this isolation only under `cfg(test)`.

## Verification

Exact local head: `b6e6d72b4a7c61163afb46eaa00398b7589224c0`

- CI-allowlisted `cargo clippy --all-targets --all-features` with warnings denied
- `cargo build --all-targets --locked` with warnings denied
- installer-checksum suite: 8 passed
- Homebrew release-gate suite: 49 passed
- hostile system/global/XDG Git-policy fixture checks: 7 passed
- bench-metadata integration tests under hostile parent Git policy: 2 passed
- `kin-git`: 40 passed
- `kin-daemon`: 517 passed
- graph-authority checks pass:
  - both zero raw-file-search authority guards
  - hydration semantics v6
  - runtime boundaries
  - private coupling
  - registry config has no local patches
- adversarial guards catch:
  - every injected probe in all 24 answer modules at every enforced site
  - a stale allowlist entry
  - all 11 hydration mutations
- `cargo nextest run --locked --no-fail-fast` in a fresh command-scoped `TMPDIR`: 4,054 passed, 9 skipped
- `cargo test --doc --locked` passes

The exact head includes current `main` through `4c239984b909121cb1bad7928cbc67fae0d0907c`, including PR491's hydration-semantics v6 and history-role changes.

## Explicit nonclaims and remaining debt

- Counts do not attest that semantic enrichment is complete.
- Durable status and live daemon readiness are distinct reports over related authority.
- `authority_epoch` is a process-local observation fence, not a durable repository generation.
- Both init and status currently replay first-parent history and rebuild identity sets for this summary; that uncached O(H) scalability debt remains.
- MCP graph status returns a retryable error while the coarse `embedding_work` lock is held; a foreground embedding pass can hold that lock for its duration.
- Embedding coverage sampling is O(N), and v1 distinguishes HEAD from temporal scope without emitting a durable selected-graph identity.
- HEAD vector fallback remains specific to `semantic_locate`; generic offline MCP graph status still fails rather than substituting a different authority.
- This is code and exact local verification, not released-byte, production, acceptance, or investor proof.
- This PR does not close FIR-1633 or the broader release/client/dogfood readiness rails.

Closes FIR-1631
Closes FIR-1592


troyjr4103:
Independent review verdict: MERGE after the four pending CI jobs go green. No P0. The v5 schema and status-JSON consumer sweep found nothing silently broken, the fail-closed decision matrix held for every two-things-disagree case, and one candidate P1 (MCP/locate gating on artifact coverage) was actively refuted. One information-loss P1 (kin graph validate drops the pending-enrichment note; the verdict itself is correct) plus three P2s are ticketed as a follow-up rather than held, per the reviewer's recommendation. Review record preserved in the session archive.

PR #484, opened 2026-07-29
Serve repository CAS bodies at the content address the VFS tree advertises

troyjr4103:
## Root cause: the route does not exist

The daemon advertises content addresses it has no route to answer.

`api_routes()` served `/vfs/version`, `/vfs/tree`, `/vfs/stat/{*path}`, `/vfs/read/{*path}`, `/vfs/readdir/{*path}`, `/vfs/subscribe` and nothing else. `git log -S"/vfs/blob" --all` over this repo returns nothing: the route was never added and never removed, and there is no fallback or nest that could serve it. So `GET /vfs/blob/<hash>` hit axum's unmatched-route handler and returned **404 with an empty body**.

The consuming shim maps a 404 on that route to `graph blob {hash} missing for {path}` (kin-vfs `kin_provider.rs:264`, whose comment reads "The validated tree references this blob; its absence is a graph gap, never a path-not-found"). A routing 404 and a store miss are the same status code, so the symptom read as a missing blob when the real cause was a missing route.

kin-vfs moved reads from path-addressed `GET /vfs/read/<path>` to content-addressed `GET /vfs/blob/<hash>`. This side never gained the counterpart.

### This is not a hydration gap and not a hash-space mismatch

Both alternatives are ruled out by `RepositoryAuthorityManager::workspace_tree_snapshot` in kin-db. Building the tree reads **every** artifact's size with `backend.source_blob_len(repository_id, digest)` and fails the whole tree with `workspace projection artifact {path} body {digest} is absent from immutable source CAS` when any body is missing.

**A `/vfs/tree` that returns 200 has already proven every hash it advertises has a body in repository-owned immutable source CAS, in the same hash space, under the same repository id.** There is nothing to hydrate and nothing to translate.

### Impact is wider than the report

The linked report says the released pairing is fine and only main-to-main is broken. That is not what the tags show. Released kin `v0.3.6` has no `/vfs/blob` route (no kin release ever has), and released kin-vfs `v0.2.0` already uses `/vfs/blob` exclusively with zero `/vfs/read` references. **Every kin plus kin-vfs pairing from kin-vfs v0.2.0 onward is broken, not only main-to-main.**

## Store map

| Route | Hash source | Body store | Before | After |
|---|---|---|---|---|
| `/vfs/tree` | repository-v6 authority workspace tree | sizes via `backend.source_blob_len` on repository authority CAS | 200 | 200, unchanged |
| `/vfs/read/{*path}` | tree artifact for that path | `authority.manager.load_source_blob(digest)` | 200 | 200, unchanged |
| `/vfs/blob/{hash}` | client-supplied content address | no route | **404, empty body** | `authority.manager.load_source_blob(digest)` |

The new route reads the **same** store as `/vfs/read`, which is the same store the tree's hashes and sizes come from. One authority, one hash space, no second store introduced.

## The fix

`vfs_blob` parses the address with `Hash256::from_hex`, opens the same one-lease `ActiveApiRepositoryAuthority` every other VFS route opens, and reads `manager.load_source_blob(digest)`. That manager call bounds the read and re-verifies the digest of the returned bytes before they leave it, so the response is provably the content named by the address. The response carries `Content-Type: application/octet-stream` and `X-Kin-Blob-Hash`, matching `/vfs/read`.

The whole body is always returned and `Range` is ignored. A client verifies the body against the hash the tree advertised, which a partial response cannot satisfy. Verified against the consumer: kin-vfs `read_range` fetches the whole blob through `fetch_verified_blob`, verifies it, and slices locally. It never sends a `Range` header here.

### Fail-closed and typed

Every failure returns a typed JSON body carrying `error`, `hash`, `authority: "repository_v6"`, and a `detail`, following the existing `filesystem_ingest_disabled_response` pattern in this file. No path reads the working copy or a Git object database.

| Condition | Status | `error` |
|---|---|---|
| address is not a sha256 | 400 | `malformed_content_address`, store never consulted |
| authority unavailable | 424 | existing `repository-v6 authority unavailable` |
| CAS read or digest verification fails | 424 | `repository_cas_read_failed` |
| repository owns no body at that address | 404 | `graph_blob_missing` |

404 is deliberate for the genuine gap. It is the status the shim already reads as a graph gap, so that message becomes accurate rather than a misreading of a routing 404, and the typed body distinguishes the two for any other observer.

## Tests, falsified against base

1. `vfs_blob_serves_every_content_address_the_tree_advertises` commits a file through the real repository authority, `GET /vfs/tree`, deserializes the wire `WorkspaceTreeSnapshot`, takes `entry.blob_identity()`, formats the URL, and `GET /vfs/blob/<hash>`. This is the identical round trip the shim performs. On base it fails: the tree advertises `6dc6785ed72375b18c83ed7354d7a4b62df4ebe75b35bae4dac78cc2e280d471` for `src/probe.rs`, then the blob read returns `left: 404, right: 200`.
2. `vfs_blob_names_the_graph_gap_for_an_unowned_content_address` asserts the typed 404. On base it fails with `EOF while parsing a value, line 1, column 0`, which is positive proof the base 404 body is empty, that is, a routing 404 rather than a handler response.
3. `vfs_blob_rejects_a_content_address_that_is_not_a_sha256` covers a non-hex address, a short one, and a 66-character one. On base all three return 404 instead of 400.

## Coordination with the ingest-CAS work

FIR-1654's hydration fix would neither have masked nor fixed this. That ticket is about `state.blobs`, the daemon's ingest CAS, a derived cache hydrated from authority and consumed by `rebuild_projection` and `refresh_projection`. None of `/vfs/tree`, `/vfs/read`, or the new `/vfs/blob` touches `state.blobs` at all; they go through `ActiveApiRepositoryAuthority::manager`. Hydrating the ingest CAS on the cloud open path changes nothing about whether a route exists. The two land coherently in either order with no shared file region, since FIR-1654's fix is in `state.rs` and the `kin-daemon.rs` shutdown paths.

FIR-1639 is a kin-vfs-side stale-binding defect and is orthogonal, but worth noting for sequencing: this defect blocked the kin-vfs smoke end to end, so FIR-1639 could not be validated against a real daemon until this lands.

Closes FIR-1642

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -956,6 +956,7 @@ fn api_routes() -> Router<Arc<DaemonState>> {
         .route("/vfs/tree", get(vfs_tree))
         .route("/vfs/stat/{*path}", get(vfs_stat))
         .route("/vfs/read/{*path}", get(vfs_read))
+        .route("/vfs/blob/{hash}", get(vfs_blob))
         .route("/vfs/readdir/{*path}", get(vfs_readdir))
         .route("/vfs/subscribe", get(vfs_subscribe))
         // Spine endpoints — cross-repo federation queries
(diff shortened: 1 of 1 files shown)


troyjr4103:
Post-merge adversarial review (independent lane): verdict was MERGE, no P0. Consumer contract verified fail-closed: kin-vfs branches on exact-404 then a not-success catch-all, so the new 400/424 statuses terminate at EIO with no disk bypass (traced through kin_provider.rs -> server.rs -> client.rs -> intercept.rs at every syscall entry point).

Two corrections to this PR's stated release implication:

1. Merging and releasing kin alone does NOT restore the released pairing. kin v0.3.6 serves /vfs/tree with no ETag and no WorkspaceTreeSnapshot, so kin-vfs v0.2.x fails at the tree handshake before /vfs/blob is ever requested. The missing route explains main-to-main only. For a working pairing, kin-vfs must also bump kin-model (0.5.0 -> 0.6.x+, schema 3 -> 4 with deny_unknown_fields plus a new required WorkspaceSnapshotBinding field) and cut its own release. Additionally kin main removed /vfs/write-notify and /vfs/file-changed, which the v0.2.1 shim still POSTs with a silent soft-fallback to the file watcher; that needs an explicit restore-or-retire decision.

2. Blast radius on real users is smaller than the body claims: the kin-vfs v0.2.0/v0.2.1 releases carry zero binary assets and only kin-vfs-core is registry-published; kin-vfs-daemon (the crate with the blob client) is not published anywhere, so only from-source builds ever hit the broken pairing.

One performance finding filed separately: this route (like its VFS siblings) calls ActiveApiRepositoryAuthority::open per request, which takes an exclusive fs2 flock and reads + SHA-256s + deserializes the full graph snapshot (129 MB on this repo) on a tokio worker with no spawn_blocking. Since /vfs/blob is now the only upstream read path the projection uses, every projected file body pays that cost, serialized. Not a regression introduced here (siblings share the shape) but it hard-bars any VFS throughput claim until the authority manager is cached across requests.

PR #486, opened 2026-07-29
Gate test daemon reuse on build identity, not snapshot version alone

troyjr4103:
## Summary

`crates/kin-cli/tests/common/mod.rs` validated a candidate `kin-daemon` by `graph_snapshot_version` only. A daemon left over from an earlier build passes that check while carrying a different `kin-buildinfo` stamp, and that stamp is the embedder identity the daemon writes into and demands from persisted vector sidecars (`kin-daemon/src/state.rs:1262`). Prepared state a test seeds is then rejected on reopen, `indexed_embedding_count` reads 0, and the suite reports a product defect that exists only because two binaries in one target directory were built at different times.

`--compat-json` already emits `build.sha` and `build.dirty`. This compares them against the test binary's own `kin_buildinfo::sha_with_dirty`, keeping the version check ahead of it. A mismatch stops the reuse so the existing rebuild runs; a mismatch that survives the rebuild fails naming both stamps and the remedy (`cargo build --all-targets`).

The two fields are joined the way `sha_with_dirty` joins them rather than compared as a raw pair. With no commit id the dirty flag is not part of the identity, so a pair comparison would reject a daemon whose embedder identity actually matches and that no rebuild could ever clear.

## Verification

Skew forced with the `KIN_BUILD_*_OVERRIDE` build inputs, daemon stamped `00000000000000000000000000000000deadbeef` against a test binary at `e73ce4b1...-dirty`, both on `graph_snapshot_version` 12.

| case | result |
| --- | --- |
| pre-fix, skewed daemon | **accepted in 0.04s** (the defect) |
| post-fix, matched pair | accepted in 0.04s, no rebuild |
| post-fix, skew a rebuild can heal | rejected, rebuilt (56.7s), daemon returns at the matching stamp |
| post-fix, skew a rebuild cannot heal | fails loud naming both stamps and `cargo build --all-targets` |

Seven unit tests in `tests/daemon_compat_gate.rs` cover the comparison, including a drift pin asserting the harness's join still agrees with `kin_buildinfo::sha_with_dirty`.

`cargo fmt --all --check` clean. `cargo clippy -p kin-cli --all-targets --all-features` clean under `RUSTFLAGS=-D warnings` with the ci.yml allowlist.

The daemon-backed `prepared_state_json` suite was not run: lane `corpus-prep` holds the `daemon` lock (FIR-1560). The gate fires in `fresh_daemon_bin()` before any daemon spawns, so it was verified without that runtime.

Closes FIR-1607

### DIFF
--- a/crates/kin-cli/tests/common/mod.rs
+++ b/crates/kin-cli/tests/common/mod.rs
@@ -159,27 +159,109 @@ fn read_capture(mut file: std::fs::File) -> std::io::Result<Vec<u8>> {
     Ok(bytes)
 }
 
-fn daemon_compat_ok(path: &Path) -> bool {
+/// The build identity this test binary carries, in the exact form the daemon
+/// stamps persisted vector sidecars with.
+///
+/// A test that seeds prepared state writes this stamp; the daemon that later
+/// reopens the repository demands its own. The two only agree when both
+/// binaries came out of one build of `kin-buildinfo`.
+pub fn expected_build_stamp() -> String {
+    kin_buildinfo::sha_with_dirty(kin_buildinfo::get())
+}
+
+/// Recompose a `--compat-json` `build.sha` / `build.dirty` pair into the stamp
+/// `kin_buildinfo::sha_with_dirty` would produce for the same build.
+///
+/// The daemon reports the two fields separately, so the harness has to join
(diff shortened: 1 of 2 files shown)


troyjr4103:

PR #488, opened 2026-07-29
Make the locate result order self-describing

troyjr4103:
`kin locate` printed a per-row score beside a list that is not sorted by that score, so a row scored 85.10 could render below one scored 45.79 and read as a sorting bug.

The order is composed, not score-sorted:

- `adaptive_cap` emits the score-ordered cluster prefix, then appends re-admission groups. The last group ranks by *uncompressed* (pre-compression) evidence while each row still carries its compressed live score, so a re-admitted row can outscore a corroborated row emitted above it.
- `fuse_locate_results` orders by fused RRF score while each row keeps the composite score of its best-ranked variant.

A row's `score` is therefore an artifact of the stage that admitted it and is not comparable against a row admitted by a different stage.

## Approach

Sorting the human output by the printed score was considered and rejected: it would discard the composed ranking (including the no-silent-elimination re-admissions) and split the human surface from JSON, which serializes the same array order. Keeping one ranking across both surfaces means the display has to stop implying a sort key it does not use.

So the ranking stays exactly as it is and the presentation becomes honest:

- the emitted rank is now explicit and 1-based, making the array order self-describing
- the per-row score is held back for `--explain`, where it sits alongside the per-signal and per-stage breakdowns as one diagnostic rather than an implied ordering, with a one-line stderr note that scores are stage-composed and not comparable between rows
- row rendering moves into `locate_text_lines` so the emitted order and row shape are directly testable

No ranking or scoring logic is touched.

## Tests

Five new tests in `crates/kin-cli/src/commands/locate.rs` pin the behavior against a deliberately non-monotonic ranking (the shape `adaptive_cap` produces):

- rows follow the composed ranking, not the printed score, with explicit 1-based ranks
- default rows carry no `score:`
- `--explain` rows do carry the score
- human row order equals JSON `files[]` order, so the two surfaces present one ranking
- `--explain` reasons stay indented under their own row

## Verification

- `cargo test -p kin-cli` green: 914 lib tests plus every integration target, 0 failures
- `cargo fmt --all -- --check` clean
- `cargo clippy -p kin-cli --all-targets --all-features` under the ci.yml allowlist with `RUSTFLAGS=-Dwarnings` clean

Note: the branch was rebased onto current `origin/main` (5a284959) partway through. On the stale base, `command_capabilities::capability_json_keeps_the_bounded_dogfood_bar_explicit` failed with `ready_commands` 17 vs a pinned 16; that reproduced with this change stashed and resolved on rebase, so it was base staleness and not related to this work.

Part of FIR-1582 (the dead-code over-report half remains open).

### DIFF
--- a/crates/kin-cli/src/commands/locate.rs
+++ b/crates/kin-cli/src/commands/locate.rs
@@ -1272,7 +1272,7 @@ pub async fn run(
     // Persist the paging cursor so `kin locate --next` can fetch the next page
     // without the caller re-supplying the query. Best-effort; never fatal.
     super::locate_cursor::persist_locate_cursor(result.next_cursor.as_deref());
-    output_result(&result, json);
+    output_result(&result, json, explain);
     Ok(())
 }
 
@@ -1428,7 +1428,7 @@ pub fn run_with_graph(
     max_files_explicit: bool,
 ) -> Result<()> {
     let result = run_with_graph_capture(graph, text, explain, max_files, max_files_explicit)?;
-    output_result(&result, json);
+    output_result(&result, json, explain);
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


troyjr4103:

PR #489, opened 2026-07-29
Order context-pack candidates by a total order before budget admission

troyjr4103:
Context-pack entity selection was nondeterministic under a tight token budget: the same repository state could yield a different surviving set on every invocation.

## Root cause

`build_context_pack` collected subgraph entities out of `SubGraph.entities`, a `HashMap`, and ordered them by relation weight alone. Relation weight comes from a small set of per-kind constants, so equal weights are the common case rather than an edge case, and tied candidates kept whatever order the hash map produced. The greedy budget fold then admitted candidates in that order, so which entities survived truncation varied run to run.

## Changes

All three sites are in `crates/kin-context/src/builder.rs`.

- The weight sort now compares `wb.total_cmp(&wa).then_with(|| a_id.cmp(b_id))`. Entity id makes the comparator a total order over distinct candidates, and `total_cmp` replaces `partial_cmp().unwrap_or(Equal)` so a NaN weight cannot collapse the comparison into a non-total ordering and reintroduce the same defect.
- The same-file neighbor fallback sorts on `(same_file_neighbor_rank(...), entity.id)`. The rank is three coarse buckets and only the first six neighbors are admitted, so ties were routine, and the surviving set depended on the order `query_entities` happened to return.
- The supporting-file list is deduped in encounter order over the already-path-sorted `supporting_artifacts` rather than by draining a `HashSet`. `append_artifact_scoped_metadata` admits work items and annotations under the same token budget through `push_work_item` and `push_annotation`, both of which drop entries once the budget is exhausted, so set iteration order was deciding which metadata survived.

The pre-sort iteration source is deliberately left as-is. `sort_by` is stable and the comparator is now a total order over distinct entity ids, so the resulting permutation is unique regardless of collect order; an additional id sort would be dead work in the context-pack path. The new test guards the invariant instead.

`query_entities` needed no change at intake. kin-db 0.6.4, the version this workspace resolves, already sorts results with an `a.id.cmp(&b.id)` total tie-break, and `InMemoryGraph` is that same implementation. The entity-id term added to the same-file sort also removes this crate's implicit dependence on that upstream guarantee.

## Test

`tight_budget_truncation_is_deterministic_across_invocations` builds the same pack repeatedly under a budget sized off an unconstrained pack so the cut lands mid-list, using equal-weight candidates with fixed-width names so projection cost cannot break the tie. It asserts truncation actually occurred, then asserts the admitted set is identical across repeated invocations and against a second graph populated in reverse insertion order.

The test was written before the fix and fails against the unfixed builder: two consecutive invocations in one process admitted different sets of 43 entities out of 88.

## Verification

`cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features` under the ci.yml allowlist, `cargo build --all-targets`, and `cargo test --workspace` all pass. 3969 tests passed with zero failures.

Closes FIR-1518
Closes FIR-1546

### DIFF
--- a/crates/kin-context/src/builder.rs
+++ b/crates/kin-context/src/builder.rs
@@ -445,7 +445,11 @@ where
                         )
                 })
                 .collect::<Vec<_>>();
-            entities.sort_by_key(|entity| same_file_neighbor_rank(&focal, entity));
+            // Same total-order requirement as the weight sort below: the rank is
+            // three coarse buckets, so neighbours tie routinely, and only the
+            // first six survive. Ranking on entity id last keeps the survivors
+            // from depending on the order `query_entities` happened to return.
+            entities.sort_by_key(|entity| (same_file_neighbor_rank(&focal, entity), entity.id));
             entities
         } else {
             Vec::new()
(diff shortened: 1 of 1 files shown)


troyjr4103:

PR #495, opened 2026-07-29
Hold one repository authority for the daemon projection routes

troyjr4103:
## What

Every daemon VFS route (`/vfs/version`, `/vfs/tree`, `/vfs/stat`, `/vfs/read`,
`/vfs/blob`, `/vfs/readdir`) opened repository-v6 authority per request through
`ActiveApiRepositoryAuthority::open`. That call chains
`RepositoryAuthorityManager::open` → `load_recovered_snapshot` →
`LocalFileBackend::load_recovery_state`, which takes an exclusive `fs2` flock on
`.kin/kindb/<repo>/.lock`, reads the complete graph snapshot, SHA-256s all of
it, and fully deserializes it. On this repository that snapshot is ~129 MB, so
the cost of one projected read was proportional to whole-repository size rather
than to the bytes requested.

The same routes performed that lock, IO, and CPU directly on tokio worker
threads, while the file's established pattern for blocking storage work is
`spawn_blocking` (ten call sites on base, one of them wrapping this same
authority open).

## How

1. `DaemonState` now holds one `ProjectionAuthorityCache`. The projection routes
   resolve authority through it, so at one publication they share a single
   loaded `RepositoryAuthorityManager` instead of each opening their own. A
   `load_gate` serializes misses, so a burst of concurrent cold requests pays
   for one load rather than one per request.
2. All remaining blocking authority work in those routes moved into
   `spawn_blocking`, matching the file's existing pattern.

`ActiveApiRepositoryAuthority` now holds its manager behind an `Arc` so one open
can be shared; nothing else about it changed.

## Freshness mechanism

The per-request reload this replaces bought per-request freshness, and none of
that may be given up — including for publications made by another process, such
as a CLI ingest or commit running beside the daemon.

Every request revalidates before reusing the held authority, with two cheap
checks and no snapshot decode:

- **Namespace identity** — `LocalFileBackend::probe_pinned_repository_namespace`.
  This is the check `RepositoryAuthorityManager::open` performs itself, so
  reusing a held authority has to perform it too; otherwise a `.kin/kindb`
  namespace replaced under the daemon would keep being answered from the store
  it replaced. The probe takes no repository lock and decodes no snapshot.
- **Publication identity** — the SHA-256 of KinDB's own commit record,
  `{kindb}/{repo_id}/authority.json`. KinDB stages snapshot and delta bodies,
  fsyncs them, and then atomically renames that record into place; the rename is
  the commit point. The record names the base and head generations and the
  digest of the snapshot bytes they describe, so its exact contents identify one
  publication. Generations only advance, so a record that reads identically
  twice cannot have moved and returned in between. It is read through a single
  file handle under a 1 MiB bound, and that bound is enforced by reading one
  byte past it rather than by a separate size probe, so nothing measures a
  record it did not then read and a commit landing mid-check cannot bound
  one record and return the bytes of another.

Ordering is load-bearing and is commented as such: the label is read **before**
the load it describes, never after. A label taken afterwards could name a
publication that landed during the load, marking older bytes as current and
serving them past the commit that replaced them. Taken before, a matching label
proves no publication landed across the whole load-and-reuse window.

One mechanism covers both directions, which is why there is no invalidation
hook to keep in sync: the daemon's own write paths open their own authority and
publish through the same record, so a commit made here and a commit made by a
separate process are both picked up by the same comparison.

Everything ambiguous reloads or fails closed: an unreadable record, an
oversized record, a probe that reaches no verdict, and a failed reload all
refuse to serve the held authority. A failed reload leaves the stale label in
place, so the next request mismatches and retries rather than serving what it
could not refresh.

This reads storage-commit metadata, never repository content. It decides only
whether an already-loaded authority may be reused; every answer served still
comes from repository-v6 graph truth, and no answer is ever filled or repaired
from the filesystem.

There is no cheap generation/digest accessor on the pinned `kin-db 0.6.7` public
surface — `load_snapshot_authority` and `load_recovery_state` both return the
full snapshot bytes, and `LocalNamespaceProbe` answers identity only and is
documented as unchanged by a truncated snapshot. The record read is the
sanctioned fallback, strengthened from stat-based (mtime/len) to exact contents.

## Tradeoff

The daemon now keeps authority snapshot state resident instead of allocating
and freeing it per projected read. Steady state is one loaded authority for the
single pinned repository, with no eviction; during a reload, in-flight requests
can briefly hold the outgoing generation alongside the incoming one, so peak
memory is a small multiple (roughly 2-3x) of one snapshot, not one. That is the
exchange for removing a whole-repository read, hash, and deserialize from every
request.

## Verification

Both new tests drive the real `router(state)` stack, and both were confirmed to
fail when the fix is reverted:

- `concurrent_projected_reads_share_one_authority_load` — eight concurrent
  `/vfs/blob` requests must load authority once. Reverting `vfs_blob` to a
  per-request `ActiveApiRepositoryAuthority::open` fails it with `left: 8, right: 1`.
- `projected_reads_serve_a_publication_committed_while_serving` — reads through
  `/vfs/tree` and `/vfs/read`, commits a second file through the ordinary
  repository write path (which opens its own authority, exactly as a separate
  process does), and requires the next read to serve it. Disabling the freshness
  comparison fails it with `got ["src/first.rs"]`.

The PR 484 projection tests (tree-advertised hash round trip, typed 404 graph
gap, 400 malformed address) are unchanged and pass.

## Follow-up not taken here

`workspace_tree_snapshot` still performs one `source_blob_len` metadata read per
artifact on every `/vfs/tree`, `/vfs/stat`, and `/vfs/readdir` request. That is
a separate per-request cost from the authority open this PR addresses, and is
left for its own change.

Closes FIR-1664


troyjr4103:
Independent adversarial review complete: no P0, no P1 code defect. Merge-clear.

Confirmed the designated P0 risk does not exist: RepositoryAuthorityManager holds no lock field, so caching it retains no flock; every production acquire_existing_lock is a scoped local, and concurrent CLI ingest/commit is unblocked by an idle daemon. During daemon-side write freezes the PR strictly improves behavior (one request contends on the flock, the rest park on the load gate, none on raw tokio workers).

authority.json verified stronger than the body argues: exactly three writers, all under the repository lock, atomic temp+fsync+rename, deterministic serialization, and the record content-binds the snapshot digest plus every acknowledged delta digest, so identical bytes imply identical durable state by construction. Cold-start and steady-state TOCTOU ordering sound; install monotonicity holds because the load gate spans label-read through install.

Applied from the review before merge: ca862f69 corrects the publication-record doc comment (the record can be re-minted once by opening a legacy store, so its bytes do not identify "one publication and nothing else"; with label-read-before-load that costs one spurious reload, never a stale serve). The body's spawn_blocking count (ten sites on base, not twelve) and the memory tradeoff (peak is a small multiple during reload, no eviction, single pinned repository) are corrected above. Head is ca862f69: comment-only delta from the reviewed head.

Two review findings routed outside this PR: the zero-file-search boundary script scans 1.5 percent of api.rs and blanket-exempts kin-daemon (now tracked to fix), and the freshness-ordering trap (digest-after-open would reintroduce stale-marked-current) is pinned on the kin-db accessor follow-up so it cannot be built that way.

PR #504, opened 2026-07-29
Read entity bodies from workspace head truth, one snippet field, 1-based lines

troyjr4103:
The MCP read-path defect family, all filed from real agent dogfooding. Together these
closed every in-profile way an agent could read source, which is why an agent on the
agent-default profile could hold the write surface and still be unable to safely use it.

## `EntitySourceScope::WorkspaceHead` did not read the workspace head

The root cause behind the context-pack graph-miss and the stale-position class. The scope
resolved through `workspace.base_target`, the change a workspace was **created from**, then
required the live entity to be byte-identical to its active committed revision there.

kin-model labels the two fields unambiguously: `tree` is "exact graph-owned working tree,
including uncommitted state", `base_tree_hash` is "tree at `base_target`". So any workspace
past its base failed the gate: a fresh clone whose admission populated the live graph, a
workspace holding a semantic overlay, or any repository with a commit since.
`get_entity_source` never had the bug because it reads `workspace.tree` and the live span.
The reported "three sibling surfaces serve the body, context pack reports graph-miss"
asymmetry was never a projection difference; it was two different definitions of now.

A head read now resolves the artifact from the workspace tree and slices the live span, the
same pair `get_entity_source` uses, so parity is structural. A historical read keeps the full
committed replay and uses the revision's own span, because at an older change the entity may
have occupied different bytes.

The artifact-identity binding is kept on head reads: a path later reused by a different
artifact must not feed an old entity's span another artifact's bytes, and since the bytes can
be identical, identity is the only signal. Only committed history knows the introducing
artifact, so the check runs when history records a revision and is skipped when it does not,
which is an ordinary state for a live entity rather than a gap. The existing rename and
path-reuse test still passes. What was removed is only the live-vs-base revision equality
requirement: the guard that protects a read and the gate that was blocking it are separable.

## Context pack served a synthesized stub as `body`

`kin-context`'s `project_full_body` does not read a body. It emits `// name (Kind, language)`
plus the signature for token accounting, and the pack labels that entry
`ProjectionLevel::FullBody`. Both the focal and dependency projections fell back to it
whenever the graph read returned no body.

That is the worst failure mode for a writing agent: source-shaped text that is not the
entity's source, with nothing marking it. An agent restating it as a body update deletes the
implementation. Bodies are now the graph-owned projection or null with a `body_unavailable`
reason naming the missing coordinate.

## `include_snippet` filled a field nobody could name, and its opt-out deleted the results

Two defects. The cosine arm emitted `snippet`, the fused arm emitted `body`, and the schema
named neither, so an agent that set `include_snippet: true` and read `snippet` found no
snippet key at all under whichever arm its profile served. The fused arm now mirrors the
excerpt as `snippet`, keeping `body` for the `kin locate --json` schema parity consumers
already parse, and the schema names the field.

The more serious one: `SnippetOptions::enabled` gated both the agent/JSON surface and the
body text, and `include_snippet: false` passed the default options, which switch the surface
off. So opting out of snippets returned **zero entity hits** rather than snippet-less ones. A
caller economizing on tokens lost the results. `enabled` now selects the surface and a new
`bodies` selects whether hits carry source. The human path keeps both off and is unchanged.

The end-to-end test covers the fused arm, including the declined case. The cosine arm's
`snippet` emission is pre-existing and unchanged, and cannot be exercised in the daemon fixture
because the cosine ranking needs a populated vector index that fixture does not have.

## One line-number convention at the boundary

`SourceSpan::start_line` carries tree-sitter `Point::row`, which is 0-based, while every
editor, `file:line` reference, diff hunk, and stack trace is 1-based. `entity_span_pair` in
`locate.rs` already shifted on emit, so the rule existed without being enforced in one place,
which is how some surfaces were off by one and others were not.

Conversion now happens at one seam per surface through `presentation_line` and the entity
accessors: entity responses, entity source and body reads, batched source rows,
`GraphSourceRecord`, search results, context-pack focal and dependency rows, reference rows,
`semantic_locate` cosine hits, impact rows, trace text, `kin graph` spans, and the session
ambiguity candidate list.

Deliberately unchanged: sort keys stay in the graph domain, `kin_lsp::EntityRef` stays 0-based
because the LSP protocol is, and the nested `span` object stays a faithful serialization of
graph truth whose byte offsets are read as offsets. The established convention holds: `span`
is graph truth, the sibling `start_line`/`end_line` are presentation. Impact rows now follow
it too. Span fixtures were corrected to state 0-based rows, so the assertions test the
conversion instead of agreeing with themselves.

Fixed along the way: `excerpt_from_line_range` counts 1-based and bails on `start_line == 0`,
but callers passed raw 0-based rows. As the fallback when the byte-range excerpt is empty it
clipped a line short and returned nothing at all for any entity starting on a file's first
line.

## `find_references` stopped making agents derive call sites

Rows carried only the caller entity's `start_line`, so an agent had to count forward through
a body it might not be able to see, compounding any staleness in the definition's own span.
`RelationEvidence::source_span` already records the syntax that produced each edge, so the
site is a graph fact and is served as one: `reference_lines`, 1-based, ascending, deduplicated,
restricted to the caller's own file. Empty when the parser recorded no span, which is honest
absence rather than a derived guess.

## The agent-default profile shipped writes without a body read

It carried the full transaction surface but no `get_entity_source` or `get_entity_body`, while
the context-bench profile carried both. Staging a body update means restating the entity's
current source, so an agent without a body read can only guess it, and a guessed body
overwrites the real one on commit. `get_entity_source` is added, with a structural test that
any profile carrying `kin_transaction_*` must also carry a direct body read, and a functional
test that drives the in-profile tools against a graph-backed entity.

## Verification

Gates run in order after the final commit: `cargo fmt --all --check`, `cargo clippy
--all-targets --all-features` under the ci.yml allowlist (zero warnings on workspace code),
`cargo build --all-targets`, `cargo test --workspace`. Also
`scripts/check_runtime_boundaries.sh`, `scripts/zero_file_search_guard.sh` (answer paths
graph-clean, the constraint this work is bound by), and
`scripts/check_private_repo_coupling.sh`.

Note for a red CI: `kin-cli --test locate_json_stdout` spawns real daemons and each test runs
over 60 seconds, so it is contention-sensitive. During this work it failed 5 of 6 under
cross-lane load with `Connection refused` on a daemon it spawns itself, and passed 6 of 6 in
isolation. That file asserts nothing about `entities`, `snippet`, or `start_line`. Re-run it
alone before believing a failure there.

Not claimed: FIR-1625 is addressed on the read side. Reference sites are served as graph facts
and head reads resolve against the workspace tree rather than the base change, so a
post-commit read reports post-commit positions, proven by a test that commits a real
span-shifting transaction. Whether reconcile updates every live entity span after every commit
shape belongs to the transaction path. No benchmark or retrieval-quality claim is made.

## Fix round after adversarial review (head 44ef1575)

A review of `9a1d6a62` returned NO-GO on the core change. Three things were true: the head read
took its bytes and its span from two stores nothing tied together, it labelled the result with a
committed change id that did not contain those bytes, and no fixture could tell a head read from
a read at base, so the headline change had no discriminating test. All are addressed.

**One coherent authority sample.** Authority reads return an `Arc` snapshot of the whole
published state, so one lease is one instant. The read took two: `workspace()` for the tree and
`current_source_change_id()`, which called `workspace()` again, for the change id. A concurrent
admission between them paired one generation's bytes with another's provenance.
`workspace_sample()` takes one lease and derives the tree, its hash, its generation, and its
resolved base from it; `current_source_change_id()` is deleted, since its only caller was the
site being fixed.

**The span is now bound to the bytes it cuts.** Blobs are content-addressed and reloading one
re-verifies its digest, so the bytes were never the loose half; the span was. The admission path
publishes the exact tree in one transaction (`publish_workspace_tree`, with an empty semantic
delta) and re-derives entity spans in a later one, so in between the graph holds a path's new
blob and the old offsets into it. Stale offsets normally still land inside the file, so the
bounds check passes and the caller receives a fragment of the wrong region as the body. The
reconciler already records the digest each span was derived from, and says why: span and blob
provenance must advance together. That digest is now compared against the artifact being sliced.
A mismatch refuses and names the retry. An entity with no recorded digest is served and reported
`span_coherence: "unverified"` rather than implied coherent, because entities admitted by paths
that do not stamp provenance, and entities rebuilt from history, legitimately lack one.

**Response shape, and why this shape.** `source_change_id` was the workspace's base change while
the bytes came from a tree routinely ahead of it, so after any admitted editor save every read
attested committed provenance for state no commit contained. `SourceProvenance` now splits the
two cases behind a `source_state` discriminator. Committed reads carry `source_change_id`.
Workspace reads carry `workspace_tree_hash`, `workspace_generation`, and `base_change_id`, and
**omit `source_change_id` entirely** rather than pairing it with a caveat: a consumer that reads
only that key cannot then be misled, which a marker alongside a populated field would not
achieve. Safe because no consumer of this path reads it. Outside it, `kin_artifact_read` and
`kin_release_check` use `source_change_id` through their own independent resolution, and none of
the 11 `WorkspaceHead` callers is a review, gate, provenance, verification, proof, or bench
path. One emission seam renders all of it, so the body-serving surfaces cannot drift apart the
way the snippet field name did between the two `semantic_locate` arms. The `stale` field is
removed: nothing in the tree ever set it true, so it asserted a freshness the read had never
established, and a proven-incoherent read now refuses rather than being flagged.

**Paging.** A cached locate ranking IS the paged answer, so the body mode it was built in decided
what later pages returned: declining bodies still got source, and requesting them got hits with
no snippet, which is the `include_snippet` defect relocated to page 1. The mode now joins the
cursor key AND is verified on lookup, because the cursor token carries the key and a caller can
present the other mode's. The fused arm also resolved its snippet options after the cursor fast
path; it now resolves them before. The cosine arm had the same flaw and is fixed the same way.

**Sibling surface.** `kin locate --json --no-snippets` returned zero entities because one flag
gated both the entity projection and the body text there too. `LocateRequest` gains an additive
`entity_surface` field so the two are separable; `locate_debug` and ContextBench deliberately
stay on the old projection so no benchmark number moves.

**Tests.** `admit_test_workspace_tree` builds the state no fixture could: tree and generation
advanced, base pinned, nothing committed. Each P0 has a test that fails on revert, verified by
reverting: with the digest check disabled the read returns `"export function alpha() {\n  const
extr"` instead of erroring, with the old stamp it reports `Committed{..}` for uncommitted bytes,
and pointing the head arm back at the base tree fails all three new tests. The impact handler's
presentation wiring and the agent-profile body read are now asserted through the production path
instead of around it, and the identity-binding skip path for a live-only entity is covered.

Two committed lane scratch logs are removed from the tree and the pattern is gitignored.

Closes FIR-1684
Closes FIR-1681
Closes FIR-1680
Closes FIR-1662
Port of FIR-1625

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@
 
 # Logs / local artifacts
 firebase-debug.log
+# Gate/check output redirected to the repo root by local multi-lane tooling.
+# Tracking these lands one lane's transient run output on every other lane.
+.kin-coord-*.log
 node_modules/
 **/node_modules/
 __pycache__/
--- a/.kin-coord-boundaries.log
+++ b/.kin-coord-boundaries.log
@@ -1,1 +1,0 @@
-Runtime guardrails check passed.
--- a/.kin-coord-boundaries.log
+++ b/.kin-coord-boundaries.log
@@ -1,0 +1,1 @@
+Runtime guardrails check passed.
--- a/.kin-coord-zfs.log
+++ b/.kin-coord-zfs.log
@@ -1,0 +1,1 @@
+Zero File-Search guard passed: answer paths are graph-clean.
--- a/.kin-coord-zfs.log
+++ b/.kin-coord-zfs.log
@@ -1,1 +1,0 @@
-Zero File-Search guard passed: answer paths are graph-clean.
--- a/crates/kin-cli/src/commands/contextbench_locate.rs
+++ b/crates/kin-cli/src/commands/contextbench_locate.rs
@@ -155,6 +155,10 @@ pub async fn run(task_file: PathBuf, json: bool, debug: bool) -> Result<()> {
         max_files_explicit,
         None,
         false,
+        // ContextBench scores `files[]` and `symbols[].name`, not `entities[]`.
+        // Held at the pre-existing projection so this change cannot move a
+        // benchmark number.
+        false,
         crate::commands::locate::LocatePaging::default(),
     )
     .await?;
--- a/crates/kin-cli/src/commands/graph.rs
+++ b/crates/kin-cli/src/commands/graph.rs
@@ -48,6 +48,13 @@ pub struct GraphSourceRecord {
     pub end_byte: usize,
     pub signature: String,
     pub body: String,
+    /// Whether the span that cut `body` was proven to describe these exact bytes
+    /// (`kin_mcp::handlers::common::SpanCoherence::label`).
+    ///
+    /// A provable mismatch never reaches here, because it fails the read. What this
+    /// distinguishes is a verified pair from one the graph recorded no digest for,
+    /// so a caller about to restate this body as an edit can tell which it has.
+    pub span_coherence: String,
 }
 
 /// The three distinguishable results of resolving an entity's source for
@@ -870,7 +877,18 @@ fn graph_source_record(
         .span
         .as_ref()
         .ok_or_else(|| anyhow::anyhow!("entity '{}' has no source span", entity.name))?;
-    let bytes = read_entity_file_bytes_from_graph(binding, graph, entity)?;
+    let (bytes, blob) = read_entity_file_bytes_with_digest(binding, graph, entity)?;
+    // Bind the span to the bytes it is about to cut, through the SAME rule the MCP
+    // resolver uses.
+    //
+    // This arm resolves its own bytes rather than routing through
+    // `read_entity_source_excerpt_detailed`, and it is the arm the daemon serves
+    // `get_entity_source` and `get_entity_sources` from, so in product mode it is
+    // the arm agents actually read through. Leaving the check on the offline
+    // resolver only would have protected the path used least. The bounds checks
+    // below cannot substitute: a stale span normally still lands inside the file.
+    let span_coherence =
+        kin_mcp::handlers::common::span_source_coherence(entity, &blob, &file_origin.0)?;
     if span.start_byte >= span.end_byte {
         anyhow::bail!(
             "entity '{}' has an empty or invalid source span ({}..{})",
@@ -911,14 +929,30 @@ fn graph_source_record(
         end_byte: span.end_byte,
         signature: entity.signature.clone(),
         body,
+        span_coherence: span_coherence.label().to_string(),
     })
 }
 
 pub(crate) fn read_entity_file_bytes_from_graph(
     binding: &kin_core::LocalRepositoryAuthorityBinding,
-    _graph: &impl GraphStore,
+    graph: &impl GraphStore,
     entity: &Entity,
 ) -> Result<Vec<u8>> {
+    read_entity_file_bytes_with_digest(binding, graph, entity).map(|(bytes, _)| bytes)
+}
+
+/// The bytes at an entity's path in the exact workspace tree, plus the digest
+/// they were loaded by.
+///
+/// Callers that go on to SLICE these bytes with the live entity's span need the
+/// digest, because the span and the bytes come from two independently updated
+/// stores and the digest is what binds them. Returning it here rather than
+/// re-resolving the artifact keeps the pair from being sampled twice.
+pub(crate) fn read_entity_file_bytes_with_digest(
+    binding: &kin_core::LocalRepositoryAuthorityBinding,
+    _graph: &impl GraphStore,
+    entity: &Entity,
+) -> Result<(Vec<u8>, kin_model::Hash256)> {
     let file_id = entity
         .file_origin
         .as_ref()
@@ -948,12 +982,13 @@ pub(crate) fn read_entity_file_bytes_from_graph(
             workspace.workspace_id
         );
     };
-    authority.load_source_blob(hash).with_context(|| {
+    let bytes = authority.load_source_blob(hash).with_context(|| {
         format!(
             "repository-v6 source body for artifact {:?} at '{}' is unavailable",
             artifact.artifact_id, file_id.0
         )
-    })
+    })?;
+    Ok((bytes, hash))
 }
 
 #[cfg(test)]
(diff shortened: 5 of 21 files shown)


troyjr4103 (commented):
Exact-head independent review: **NO-GO** at `1636447ed2633c2fa9978e06db811a3a5edce446` despite terminal-green hosted checks.

Blocking P1 findings:

1. Historical/session-scoped `get_entity_source` and batch body reads resolve the historical graph but unconditionally read workspace HEAD bytes (`crates/kin-daemon/src/api.rs:6231-6289`).
2. Context-pack fitting counts header/signature stubs, then appends up to 2,400 characters of body per non-focal entry without final re-budgeting, so `tokens_used` can under-report and the response can exceed `token_budget` (`entities.rs:620-717`).
3. Locate overload matching compares 1-based presentation lines with 0-based graph rows, causing the exact match to miss and potentially binding the wrong same-name entity/body (`locate.rs:15309-15322`).
4. Workspace-head body projection can replay full history, then resolve a tree, once per entity; default multi-hit locate/context/reference surfaces amplify history × hits (`common.rs:1314-1328,1488-1496`).
5. FIR-1625 acceptance remains open; the PR directive has been corrected from `Closes` to `Port of`.

P2 findings: the real daemon body-read arm drops the new committed/workspace provenance contract; newline-terminated tree-sitter spans overstate inclusive end lines; and history replay errors are collapsed into ordinary absence, silently disabling artifact-identity binding.

Required before merge: fix each code path, add historical-session/budget-tier/overload/large-history/trailing-newline/fail-loud regressions, rerun the complete matrix, and obtain another fresh exact-head review. The coordination gate now records this head as `NO_GO`.

troyjr4103:
🚫 **Merge gate changed after the earlier review brief.** Exact head `1636447ed2633c2fa9978e06db811a3a5edce446` is now recorded `NO_GO` in `kin_pr504_exact_head` with **5 P1 + 3 P2** findings (review 4817689948).

A verifier scoped only to the three older “expected-GO” items cannot authorize merge: it does not examine the newly found historical-session/current-HEAD body mismatch, final context-pack token undercount, overload line-base misbinding, per-entity full-history replay, daemon provenance loss, exclusive end-line error, or swallowed replay failures.

Stop on any scoped GO. Repair all eight findings, rerun the full exact-head matrix, and obtain a fresh full review.

troyjr4103:
CAPTAIN HANDOFF: this PR is fully cleared to merge. Adversarial review chain complete: original NO-GO (2 P0, both CONFIRMED) at 9a1d6a62; remediation closed both P0s at 44ef1575 (re-review: both CLOSED, blob_hash coherence design adjudicated better than the reviewer's own suggestion); short round closed P1-A (whole-projection cache key) and P1-B (per-row provenance) at 1636447e; final verification GO, 3 of 3 CONFIRMED, 0 P0, 0 P1, 2 non-blocking P2s (reports: .kin-coord/reports/review-pr504-20260729.md and verify-pr504-final-20260730.md). Head provenance: the fix lane's commits plus one captain rebase with a single union-import conflict resolution, compile-verified, disclosed and verified in the final report. CI fully green at 1636447e. If main moved since (502 landed), the relay is a plain rebase; zero file overlap expected with 502 (test-infra) but re-run the scanner pair if any guard changed. Merging closes FIR-1684, FIR-1681, FIR-1680, FIR-1662, FIR-1625 via the body's Closes lines; please add the squash sha as an evidence comment on each after landing.

PR #505, opened 2026-07-29
Report admission progress by default for init and clone

troyjr4103:
## Summary

`kin-cli` installed `EnvFilter::new("warn")` as the default for every command, so nothing emitted at `info` reached the user unless they happened to set `RUST_LOG`. That discards admission output wholesale, including the periodic replay progress KinDB is gaining.

- make `default_env_filter` command-aware, and split the directive choice into a pure `default_filter_directives` function
- raise only admission targets, and only for `init` and `clone`
- leave every other command on the quiet default, since a short command prints its own result and has nothing to report in between

`RUST_LOG` still wins outright. Profiling is unaffected because that path installs its own subscriber with no `EnvFilter` and never reaches this function.

## What this delivers, and what it does not yet

Being precise about this, because the failure mode it guards against is someone reading silence as a wedged process:

- The two `kin_core` targets carry three callsites between them, and all three are **terminal**: two report a completed admission (`init.rs:426`, `git_init.rs:315`) and one reports recovered stale initialization stages (`init.rs:2088`). They add at most three lines, and **none of them fires mid-flight**.
- `kin_db::storage::history_replay` is the only periodic emitter. It does not exist in the KinDB this workspace pins, and a directive naming an absent target never matches, so it is **pre-wired and inert**.

**So a long admission does not report mid-flight progress at this version.** It begins to the moment this workspace consumes a KinDB shipping `storage::history_replay`, with no further change here. Until then, silence during a long `kin init` is not evidence that instrumentation ran. The doc comment on `ADMISSION_PROGRESS_TARGETS` says the same thing at the code.

What the change buys now is that the admission surfaces its own outcome, and that the wiring is in place so the periodic emitter needs no second kin-cli change to become visible.

## Why this shape

A crate-wide `kin_core=info,kin_db=info` was rejected: it surfaces every unrelated event those crates emit, and grows silently as either crate adds logging. A dedicated progress target that emitters opt into is the better long-term shape, but it needs edits in the emitting crates rather than here, so it is left as a follow-up.

## Verification

Four tests in the existing `kin-cli` test module, two of which exist because this feature can fail silently:

- `progress_reporting_commands_are_real_subcommands` pins each configured name against clap's own command tree. The filter is keyed on the subcommand name, so a rename would stop reporting with every other test green, and this CLI does rename subcommands (`#[command(name = "locate-debug")]`).
- `every_default_directive_parses_and_is_retained` parses the exact string the binary installs and requires every directive to survive, because `EnvFilter` drops a malformed directive silently. It asserts on the parsed filter's rendering, not on the input.
- `progress_reporting_commands_raise_admission_targets_to_info` and `ordinary_commands_keep_the_quiet_default` cover the two live branches.

Known limits, recorded rather than hidden. No test can verify that a target string names a module that exists; `kin_core::innit=info` parses, survives, and matches nothing. That is unverifiable from `kin-cli`, which cannot introspect another crate's module paths, and it is the strongest argument for the dedicated-target follow-up. The three current targets were verified by reading the emitting files. Nothing asserts that `main()` passes the real command name either; guarding that needs a subprocess stderr assertion, which is heavier than the risk warrants.

`RUST_LOG` precedence is covered by construction rather than by a test, deliberately: asserting it requires mutating process environment, which races every other test in the binary.

MCP stdio hygiene is sound on two independent grounds: the single `fmt::layer()` in the binary writes to stderr (`main.rs:1970`), and the MCP stdio server is `Command::Mcp`, which is not in the raised set and keeps `warn` unchanged. `kin init --json` also stays pure JSON on stdout, since every added line goes to stderr.

Gates, re-run after the final edit:

- `cargo fmt --check`: clean
- `cargo clippy --all-targets --all-features` with the ci.yml allowlist under `RUSTFLAGS=-D warnings`: clean
- `cargo build --all-targets`: ok
- `cargo test --workspace`: 40 test binaries ok, 0 failures

## Scope

One file, `crates/kin-cli/src/main.rs`. No MCP handler, daemon lifecycle, or CI file touched.

Progresses FIR-1694 (ticket stays open per its acceptance-boundary comment: mid-flight progress requires the kin-db storage::history_replay consumption)

### DIFF
--- a/crates/kin-cli/src/main.rs
+++ b/crates/kin-cli/src/main.rs
@@ -1966,7 +1966,7 @@ fn main() -> Result<()> {
             .init();
     } else {
         tracing_subscriber::registry()
-            .with(default_env_filter(false))
+            .with(default_env_filter(false, &command_name))
             .with(tracing_subscriber::fmt::layer().with_writer(std::io::stderr))
             .init();
     }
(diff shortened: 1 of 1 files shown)


troyjr4103:
Pre-verified for the reviewer, since a verbose filter sharing a process with an MCP stdio JSON-RPC stream is the one shape worth probing here. It is sound on two independent grounds, either of which suffices on its own.

**1. The only fmt layer writes to stderr.** `crates/kin-cli/src/main.rs:1970` is the single `fmt::layer()` in the binary, and it is `.with_writer(std::io::stderr)`. No log level can reach stdout, so no filter change can corrupt a stdout protocol stream. That held before this PR and is untouched by it.

**2. MCP is not in the raised set.** The MCP stdio server is `Command::Mcp` (`main.rs:1526` "Start the MCP stdio server", dispatched at `main.rs:2538`), whose top-level subcommand name is `mcp`. That is not in `PROGRESS_REPORTING_COMMANDS`, so it keeps the `warn` default, byte-identical to before this change.

Nesting cannot defeat it either: `current_command_name` reads `matches.subcommand_name()`, the top-level name only, so a nested `init` or `clone` under any parent could never select the verbose branch.

The adjacent hygiene case is also clean. `kin init --json` writes its payload to stdout (`commands/init.rs:188`) while every line this PR adds goes to stderr, so machine-readable output stays pure JSON.

Happy to add a regression test pinning either property if the review wants one asserted rather than argued.

PR #506, opened 2026-07-29
Preserve merge bases and fail closed on ambiguous acknowledgements

troyjr4103:
## What changed

- records the true base-side relation value when a dangling-endpoint conflict is created
- sends `merge` and `resolve` exactly once with caller-stable operation IDs
- disables both redirects and reqwest retries for non-idempotent mutations
- returns a typed indeterminate outcome for every failure after dispatch
- accepts success only when the acknowledgement echoes the exact operation ID and carries its authoritative report
- keeps pre-dispatch serialization failures direct and preserves existing retries for caller-idempotent commands

## Why

The durable merge record landed with two remaining correctness gaps: a dangling relation could lose a present base value, and the client could replay a mutation after an ambiguous acknowledgement even though the daemon might already have committed it. Both paths now fail closed without inventing state or silently redispatching work.

HTTP status alone is deliberately not treated as rejection proof because repository authority can move before daemon-side finalization reports an error. Command-level receipt recovery remains separate protocol work.

## Exact-head validation

- head `c0432bdaf0cbe981890cf4f80f51d524d064aefc`
- tree `ffc7bf9a898b721c5a2003c188480a91c7808ab9`
- current `origin/main` `91d6626c3a1ff7f9c9ad093436ab6dc875eee508` integrated
- `git diff --check`
- non-idempotent transport regressions: 8/8 passed
- authoritative merge/resolve acknowledgement regressions: 2/2 passed
- repository merge and resolution acceptance: 17/17 passed
- final independent exact-head review: GO, zero findings

The 17-test nextest run reported two base-harness leak annotations because this branch does not yet contain PR 502; the exact orphan test processes were identified and cleaned. PR 502's current exact-head 4,503-test workspace run is green without leaked target processes.

## Still required

- terminal hosted CI at this exact head
- merge-queue turn after all earlier Kin PRs
- released-bytes and production acceptance remain separate gates

Relates to FIR-1621.


troyjr4103 (commented):
Exact-head operational HOLD at relayed head `c98f73b188f4570256ad6148dc3f4fbc47f174cd`.

The relay is on current main `1d99e30c...`, but the PR body still attests the pre-relay head `c0432bda...` and old base `91d6626c...`; GitHub has zero fresh reviews on `c98f73b1`; the queue-required full exact-head gate plus strict containment is not evidenced; and macOS is still running at this snapshot. Terminal hosted CI is necessary but not sufficient under the recorded queue note.

Do not merge on CI alone. Refresh the body with actual post-relay evidence, run/record the full gate and strict daemon containment against `c98f73b1`, obtain a fresh independent exact-head review, then recheck queue/head/base/cleanliness before merge. This is an operational evidence hold, not a newly asserted source-code defect.

troyjr4103 (commented):
Hosted CI has now reached terminal green on the relayed head, which closes only the hosted-check portion of the recorded hold. The PR body still attests `c0432bda` over `91d6626c`, not `c98f73b1` over `1d99e30c`; no fresh independent review exists for `c98f73b1`; and the queue-required full exact-head gate plus strict daemon containment is still not recorded. Keep this head unmerged until those remaining conditions are satisfied and the body is corrected. CI green is necessary, not sufficient.

PR #515, opened 2026-07-30
Prove process-group containment against exit state, not group membership

troyjr4103:
The post-reap containment check asked `kill(-pgid, 0)` whether the process group still had members and treated any answer but ESRCH as a failure. That is not the question the guarantee is about. A process that has exited but has not yet been waited keeps its slot in the process table and stays a member of its group, so an ESRCH-only test cannot tell a group that still runs code from a group holding nothing but corpses.

Waiting owned children first, which the old error text instructed the caller to do, cannot close this. A grandchild that outlives its parent is reparented to init, and its slot is cleared when init gets around to it, which is a window no caller participates in. The failing case in practice is exactly that shape: a fake git spawns a background sleep, so the group holds a grandchild whose corpse nobody local can collect. Under load the window widens and a cleanup that had genuinely killed everything reports failure.

Classify the members instead. An exited process holds no address space, executes no instructions, and cannot fork, so it cannot leave the group or outlive the call in any sense containment is about, and reporting it as contained states what the guarantee already meant rather than relaxing it. A group that is non-empty and cannot be classified stays indeterminate rather than being folded into either answer, so an unreadable group can never silently pass as contained.

Reading exit state needs care on macOS. The full BSD info record is assembled from a process's task, which an exited process has already released, so that query fails with ESRCH for precisely the processes this check must recognize. Pairing it with a signal probe resolves the ambiguity: info unavailable together with still signallable is a positive identification of a process that has exited, not an absence of evidence. The struct that reports the state directly is not in the libc version this workspace pins, and hand-declaring a kernel struct layout to save a syscall is the worse trade. Linux reads the one place the kernel publishes a process's group, parsing the stat line after the final parenthesis so a process whose name contains spaces or parentheses is not mis-parsed.

One existing test required finalization to fail on an uncollected direct child. That expectation cannot survive admitting the grandchild case, since a reparented process reaches the same state with nobody at fault, and both repair directions proposed for this defect invalidate it identically. The test keeps the assertion that carries the guarantee, that the member was killed by the barrier, and drops the one that made a corpse count as a containment failure.

Three new tests drive the states directly rather than racing for them, and each asserts what the old membership probe would have answered alongside the new verdict, so the arrangement that used to fail is pinned rather than described. Reverting the classification makes the end-to-end case fail with the reported error string verbatim, which turns an intermittent load-dependent failure into a deterministic one.

Reading the Linux process table is raw filesystem access as far as the zero-file-search guard is concerned, and the spawn crate's whole-file exemption that would once have covered it was replaced by per-primitive pins. The one directory enumeration and the two per-process stat reads are registered as counted pins under the existing entry rather than widening it back to a whole-file exemption, so every other line of the crate stays scanned and a fourth process-table read fails the guard until someone justifies it.

Enumerating members on macOS needed the same care as reading their state. `proc_listpids` does not use the -1/errno convention the rest of the module reads, because libproc turns a failed `__proc_info` into a return of zero and leaves the reason in errno. Both negative tests were therefore dead code, and a denied or failing query reached the caller as an empty member list, which classifies as contained and makes finalization succeed. Clearing errno before each call separates the two readings, so a failure now leaves containment indeterminate while a group that genuinely empties between the sizing call and the read still reads as empty.

The durable reaper held a second copy of the retired emptiness probe and logged an observable group as an error, which is the state this change establishes as contained. It now asks the same containment question the launcher asks, keeping the error for a member that can still execute and reporting an unclassifiable group as a warning, and the public reap documentation describes what the code requires instead of the retired check.

Closes FIR-1719


troyjr4103:
Delta review (r3) at exact head `3d2f580c6bd53ff7b1a96389b3027cbec3eed851`, against the r2 NO-GO spec. Full report: `.kin-coord/reports/review-kin-pr515-20260730-r3.md`

**VERDICT: GO.** All four r2 items satisfied, no new findings, no code change requested. **Conditional only on `Windows authority tests`**, a required check still `pending` at review time; everything else is terminal (11 SUCCESS, 3 SKIPPED). Nothing in the source review waits on it.

**Zero File-Search: PASS on the actual merged tree.** `origin/main` (`e115d1b4`) is an ancestor of this head, so running the guard here *is* the r2 P0 scenario rather than an approximation. Exit 0, **144 files scanned**, matching spec. The file count is the load-bearing part: an allowlist error in this guard exits *before* the scan, so a malformed pin can produce success-shaped output having scanned nothing. The three `/proc` pins are bare strings, and I confirmed from the guard's own source (not the entry's prose) that bare means exactly-once and declared counts are enforced exactly.

**P1 genuinely fixed, traced through all three links.** The diagnosis is right: `proc_listpids` does not use the -1/errno convention, libproc turns a failed `__proc_info` into a return of 0, so a `< 0` test never fires. The wrapper clears errno first and returns `Ok(0)` **only** when `written == 0 && errno == 0`; every other non-positive return becomes `Err`. `process_group_members` propagates both calls with `?`. Classification maps `Err(_) => Indeterminate`, so a failed enumeration can no longer reach `Empty`. Two supporting checks: `Indeterminate` is not soft, finalization maps it to `Err`, so an unreadable group fails loud; and the surviving ESRCH short-circuit is sound because ESRCH on a group probe is a definitive "no such group", while EPERM is not ESRCH and so falls through and fails closed.

**P2s: two absorbed, one explicitly deferred.** The durable reaper now calls `process_group_containment` with `Empty | OnlyExited` silent, `LiveMember` an error, `Indeterminate` a warn. The `try_reap` doc no longer promises the retired ESRCH contract. The over-broad `OnlyExited` is deferred in the lane report §6 as a captain follow-up, matching the r2's own "not a merge blocker" framing; ticketed as FIR-1731 so it cannot be lost.

**Workspace-suite skip: ACCEPTABLE.** I verified the load-bearing claim independently rather than accepting it: all four changed symbols have zero references outside the crate file, with a control confirming the search works (18 total hits, 0 external), and neither the enum nor the classification fn carries `pub`. One honest qualification the lane understates: the P1 does change *behaviour* reachable through the public `try_reap` path, so "private" is true of symbol visibility rather than behavioural reach. The ruling holds anyway, because the change is confined to the enumeration-failure path, the crate suite covers it, and hosted CI already ran the full workspace suite green on both macOS and ubuntu at this head. The local skip is backstopped by a hosted run that actually happened.

**Independent execution:** guard exit 0 / 144 files; `cargo test -p kin-daemon-spawn` **39 passed, 0 failed** in an isolated target dir under the `test-box` lock, matching the lane's reported 39/39 exactly.

The lane's stability evidence is credible and self-incriminating in the right direction: 12 extra repeats gave 11 pass / 1 fail, attributed to cause B by line-identity rather than by assertion text, showing the failing line byte-identical to the r2's cause-B assertion with an offset exactly equal to the lines its own commits added. Cause B is FIR-1730 and is neither caused nor cured here.

**Not claiming:** I verified the errno-clearing logic by reading it and did not construct a failing `proc_listpids` to observe `Indeterminate` end to end. I relied on the r2's workspace-suite run rather than repeating it. No claim about released bytes, performance, or FIR-1719 being closed in production.

PR #520, opened 2026-07-30
Bind the release tag mint job to the release-tag environment

troyjr4103:
The job resolves the kin-release-bot App credentials from whatever scope
GitHub exposes to it, and with no environment declared that is repository
scope. A key rotation applied to the release-tag environment therefore
never reaches the mint, which then fails while requesting an installation
token because the stale repository copy no longer decodes as a JSON web
token.

Declaring the environment makes the environment-scoped credentials
authoritative for this job. The environment's deployment branch policy
admits main only, which matches the ref guard the workflow already
enforces, so the binding adds no new approval or wait state.

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -47,6 +47,13 @@ jobs:
     name: Mint release tag
     runs-on: ubuntu-latest
     timeout-minutes: 15
+    # The kin-release-bot App credentials are scoped to this environment, which
+    # a job only resolves when it declares the environment. Without the binding
+    # the job silently falls back to repository-scoped copies of the same secret
+    # names, so a key rotation applied to the environment never reaches the mint.
+    # The environment's deployment branch policy admits main only, which matches
+    # the ref guard below.
+    environment: release-tag
     steps:
       - name: Guard — authorized actor and ref
         # Only the release captain or the bot may dispatch, and only against the


troyjr4103:
Evidence and blast radius, since this touches the release tag rail.

State today: this repository holds KIN_RELEASE_BOT_APP_ID and KIN_RELEASE_BOT_PRIVATE_KEY at repository scope with an updated_at of 2026-07-22T23:53:36Z, and the same two names in the `release-tag` environment with the private key updated 2026-07-30T10:46:10Z. `mint-release-tag` declared no environment, so `secrets.KIN_RELEASE_BOT_PRIVATE_KEY` resolved to the July 22 repository copy. The last successful mint here was 2026-07-24, before the rotation.

The same defect took down the registry release rail today: kin-model and kin-db reached `actions/create-github-app-token@v2` with repository-scoped copies of the same App key and failed with "A JSON web token could not be decoded" (kin-model runs 30552506775 and 30580301904; kin-db runs 30552502205, 30578122919 and 30579680400). kin-model published 0.7.2 and kin-db published 0.7.5 and 0.7.6 with no release tag minted. There the fix is a pin roll onto kin-actions v0.1.22, which binds the reusable workflow's mint job to the publish environment; here the workflow is repo-local, so the binding is declared directly.

Checks run before proposing this: `python3 scripts/test-release-workflow-authority.py` passes on this branch. The `release-tag` environment carries a deployment branch policy for main and no required reviewers and no wait timer, so a main dispatch gains no approval gate; a branch dispatch was already refused by the workflow's own actor and ref guard.

Not claimed: no tag has been minted to prove this end to end. The next real `vX.Y.Z` dispatch is the proof. The stale repository-scoped copies are left in place deliberately, so nothing regresses if this binding needs to be reverted; removing them is a separate cleanup.

PR #545, opened 2026-07-31
Give the release container two registries and compile both musl release targets before landing

troyjr4103:
Two hardenings of the paths a release commit has to survive, batched because
both are release-path build changes with independent evidence.

Part-of FIR-1762.
Closes FIR-1779.

## 1. The release container no longer depends on one registry

`Docker Image Build (no push)` publishes a check-run against every commit that
reaches main, and release-tag.yml's second sweep refuses a release commit
carrying any non-green check-run whether or not a ruleset requires it. The root
Dockerfile pulled `rust:slim` and `debian:trixie-slim` by floating tag from
docker.io, so minutes of registry unavailability refuse that release
permanently rather than delaying it. That is the cargo-deny failure class of
the v0.4.1 burn, in the one job PR 529 did not reach.

What changed:

- both `FROM`s are digest-pinned and name their registry explicitly
- both jobs that build the Dockerfile (`docker.yml:build-image` and
  `release.yml:build_daemon_image`) configure BuildKit to resolve `docker.io`
  through `mirror.gcr.io` first, falling back to `docker.io`
- `.github/workflows/base-image-pins.yml` (schedule + dispatch only) proves
  both pinned digests still resolve from **both** registries
- the release-authority suite refuses an unpinned base image, a base image that
  does not name its registry, and either build job losing its mirror

### Why not the GHCR mirror the ticket preferred

Verified and rejected on a specific fact, not on effort. GHCR container
packages are private by default, **do not inherit repository visibility**, and
visibility is changeable only through the org UI ([GitHub
docs](https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility):
"a package automatically inherits the access permissions (but not the
visibility) of the linked repository"). `firelock-ai/kin` is public with 4
forks and ships a `docker-compose.yml` with a `build:` section, so a private
mirror in the `FROM` would break fork pull requests and every external `docker
build`, and would make the release-critical path depend on a one-time manual org
UI action nothing in the repo can assert. Pushing itself is not the blocker:
`build_daemon_image` already pushes `ghcr.io/firelock-ai/kin` with the standard
`GITHUB_TOKEN`.

The sanctioned fallback is what landed, implemented at the registry-resolution
layer rather than as a shell retry, so it applies to every consumer of the
Dockerfile including forks and external builders.

### Evidence

Run locally against the exact BuildKit configuration in the diff:

- `mirror.gcr.io` returns **byte-identical digests** to `docker.io` for both
  images (`sha256:5c6f46a6...` rust:slim, `sha256:020c0d20...`
  debian:trixie-slim), so the pin is satisfiable from either door
- the pinned bases build through a mirror-first BuildKit builder: builder stage
  (pinned rust base, apt layer, tracked `[registries.kin]` assertion) rc=0, and
  runtime stage (pinned debian base, apt layer, non-root user) rc=0
- **fallback falsified**: the same build with the mirror host set to an
  unreachable name still completes against `docker.io`, rc=0, so the mirror is
  an additional door and not a redirect
- `scripts/verify-base-image-pins.sh` run against the committed Dockerfile:
  both pins `docker.io=ok mirror.gcr.io=ok tag=current`

`mirror.gcr.io` is hosted on Artifact Registry and is unaffected by the
Container Registry shutdown. It is a pull-through cache, so the honest claim is
two independent front doors, not our own copy of the bytes.

## 2. aarch64-unknown-linux-musl compiles before the tag exists

PR 541 closed the x86_64 half. The remaining gap: the second shipped Linux
release target first compiled inside `release.yml`, after the tag, which is the
shape that burned v0.4.3.

`Check & Test` now also runs `cargo check --locked --target
aarch64-unknown-linux-musl -p kin-cli -p kin-daemon`. Option (a) from the
ticket, not (b): a native `ubuntu-24.04-arm` leg would add a new context name,
which needs a ruleset change and perturbs the census release-tag.yml pins.

**The bound, stated plainly.** Ubuntu carries no aarch64 musl C toolchain, so
dependency build scripts compile their C sources with `gcc-aarch64-linux-gnu`:
right architecture, glibc headers. The Rust half is exact, and that is where
the recorded defect class lives. The C half against musl headers stays proven
by the native `ubuntu-24.04-arm` release leg, where `musl-tools` does supply an
aarch64 `musl-gcc`, which is why the release build is not a cross build either.

### Falsification against the recorded defect

`libc` declares the `renameat2` wrapper only for `target_env = "gnu"`, android
and redox, and kin-db 0.7.6 called it unconditionally on Linux. Run in a
container at the workspace `libc` pin (`=0.2.186`):

| scratch resolve | target | result |
| --- | --- | --- |
| `kin-db =0.7.6` | `aarch64-unknown-linux-musl` | `error[E0425]: cannot find function renameat2 in crate libc`, rc=101 |
| minimal `libc::renameat2` repro | `aarch64-unknown-linux-musl` | E0425, rc=101 |
| minimal `libc::renameat2` repro | `x86_64-unknown-linux-musl` | E0425, rc=101 |
| minimal `libc::renameat2` repro | `x86_64-unknown-linux-gnu` | **rc=0** |

The gnu control compiles, so the target selection is what catches it. Two facts
worth recording: `SYS_renameat2` (the constant the 0.7.8 fix uses) **is**
declared for aarch64 musl at this pin, so the fixed code is correct on the new
target; and `libc` 0.2.188 added the musl `renameat2` wrapper, so an
unconstrained scratch resolve silently stops reproducing the defect. The
workspace pin is 0.2.186.

### Cost

On the run this branched from (30641500854, main): the existing x86_64 musl
step cost **90s**, the ubuntu leg ran **23.2 min**, and the longest required leg
(`Windows authority tests`) ran **29.1 min**, leaving 5.9 min of slack behind
the critical path. The measured cost of the new step on this PR's own run is
reported in a comment below.

## Gates

- release-authority suite: green, with 4 new in-suite falsifications
  (unpinned base, registry-less base, each build job losing its mirror)
- 5 external falsifications, each confirmed to fail for the right reason and
  the control confirmed green after each: aarch64 `cargo check` removed,
  aarch64 `rustup target` removed, aarch64 C toolchain removed, new workflow
  missing from the census, new workflow job renamed to a required context name
- `bash -n scripts/verify-base-image-pins.sh` green, and wired into the
  existing ci.yml syntax step
- both zero-file-search guards green (146 files scanned)
- `actionlint`: 1 finding, byte-identical to the finding on `main` before this
  branch (pre-existing SC2034 in `registry-index-migrate.yml`, untouched here)
- no Rust changed, so fmt/clippy/build carry no delta from this diff

## What this does not fix

- **FIR-1762 fix 2, the policy half.** release-tag.yml's second sweep still
  terminally refuses any non-green check-run on the release sha, required or
  not. This change removes one producer's dependence on a single registry; it
  does not change what the sweep does with a failure. That is a release-
  authority change and needs its own adversarial review, which is why this is
  Part-of and not Closes.
- **We do not own the base image bytes.** `mirror.gcr.io` is a pull-through
  cache. If Docker Hub is unreachable *and* the mirror has never cached a
  pinned digest, both doors fail. For these two images (official, extremely
  popular, and now warm at the pinned digests) that is remote, and the
  scheduled job makes it observable rather than a surprise. Owning the bytes
  needs the GHCR mirror, which needs the visibility decision above.
- **The aarch64 C half.** Bounded on purpose and documented in the step, not
  hidden. Closing it would need either an aarch64 musl sysroot built in CI or
  the native arm leg that option (b) describes.

### DIFF
--- a/.github/workflows/base-image-pins.yml
+++ b/.github/workflows/base-image-pins.yml
@@ -3,16 +3,31 @@
 
 name: Base Image Pins
 
-# Scheduled and manual only, deliberately. A pull_request or push trigger would
-# publish a check-run against the commits that reach main, and release-tag.yml's
-# second sweep refuses a release commit carrying any non-green check-run whether
-# or not a ruleset requires it. A registry-reachability probe is exactly the
-# kind of job whose transient failure must never be able to refuse a release:
-# adding one to the release sha would recreate the failure class this workflow
-# exists to watch. Pre-landing coverage is already there without a new context.
-# `Docker Image Build (no push)` pulls both pinned digests on every pull request
-# and every push to main, and the release-authority suite refuses an unpinned
-# FROM inside the required Check & Test job.
+# A scheduled run's check-runs DO land on a commit. GitHub binds a check suite
+# to every workflow run, and for a `schedule` event `github.sha` is the default
+# branch HEAD at fire time, so a red scheduled run publishes a failure check-run
+# on a main commit. release-tag.yml's second sweep refuses a release commit
+# carrying any non-green check-run whether or not a ruleset requires it, and a
+# quiesced main during a release train holds HEAD still across exactly the
+# window that matters. This repo already carries a live instance of that shape:
+# fuzz.yml's scheduled run left two `failure` check-runs on a main sha.
+#
+# Dropping the push and pull_request triggers therefore does NOT make this job
+# safe on its own. What makes it safe is that the scheduled path is
+# structurally incapable of concluding failure: the prover's exit status is
+# captured rather than propagated, and the step exits 0 whatever it found. A
+# registry-reachability probe must never be able to refuse a release, because
+# that is the failure class this workflow exists to watch.
+#
+# Loudness comes from a surface that cannot poison a mint. When a pin is
+# definitively unreachable the run opens or updates a tracking issue, which
+# persists until someone acts on it, unlike a red check nobody is paged for.
+#
+# A `workflow_dispatch` run may fail hard, because an operator asked a direct
+# question and deserves a direct answer. The residual is real and is theirs to
+# own: dispatching against main during a release train can publish a failure
+# check-run on the candidate sha. Prefer the schedule unless you need the
+# answer now.
 on:
   schedule:
     - cron: "23 5 * * 1"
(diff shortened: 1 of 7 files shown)


troyjr4103:
## Measured cost of the aarch64 step, from this PR's own run

CI run [30650515943](https://github.com/firelock-ai/kin/actions/runs/30650515943), `Check & Test (ubuntu-latest)`:

| step | start | end | duration |
| --- | --- | --- | --- |
| `Check the Linux release target (musl)` | 17:29:32Z | 17:30:02Z | **30 s** |
| `Check the aarch64 Linux release target (musl)` | 17:30:02Z | 17:31:16Z | **74 s** |

The new step costs **74 s**, and it passed, so the aarch64 glibc cross toolchain
does drive every dependency build script in the release package set for that
target.

Against the job budget on this run: `Check & Test (ubuntu-latest)` 22m28s vs
`Windows authority tests` 24m45s. The ubuntu leg still finishes 2m17s inside the
longest required leg, so the step spends billed minutes and no merge-queue wall
clock. (The x86_64 step read 30 s here against a 90 s baseline on main because
this run restored a warm cargo cache; the two are not a like-for-like pair, which
is exactly why the aarch64 leg is a separate step with its own attributable
number.)

`Docker Image Build (no push)` passed in 12m3s against the digest-pinned bases
resolved through the mirror-first BuildKit config, which is the change proven on
the real substrate rather than locally.

## One unrelated red check, worth naming

`Analyze (actions)` (CodeQL, GitHub **default setup**, so it has no workflow file
in this repo and is outside the release-authority census) failed after 10 s:

```
Setup CodeQL tools
  Error: read ECONNRESET
```

A transient network failure fetching CodeQL's own tooling. It is not one of the
six required contexts on `main`, and the run cannot be re-run (`403 Jobs in this
workflow run cannot be re-run`), so clearing it needs a new head commit. I have
not pushed one, because ~25 min of CI to clear a non-required transient is worse
than saying what it is.

It is also, precisely, this PR's subject matter: a third-party fetch failing for
reasons unrelated to the diff, producing a red check-run on a commit. Had this
landed on a release sha, release-tag.yml's second sweep would have refused that
release, and no ruleset required the check that refused it. That is FIR-1762's
fix 2, the half this PR deliberately does not touch.

PR #551, opened 2026-07-31
feat(status): plumb embedding coverage into status.v2 semantic enrichment

troyjr4103:
Plumbs embedding coverage (`embeddings_indexed`, `embeddings_pending`, `embeddings_total`) from graph-owned authority truth into `SemanticEnrichmentStatus` on `kin status --json` (`kin.status.v2`). Restores progress visibility for MCP, kin-editor, and benchmark agents.

Closes FIR-1785

### DIFF
--- a/crates/kin-cli/src/commands/status.rs
+++ b/crates/kin-cli/src/commands/status.rs
@@ -91,6 +91,9 @@ pub struct SemanticEnrichmentStatus {
     pub entity_count: usize,
     pub relation_count: usize,
     pub semantic_change_count: usize,
+    pub embeddings_indexed: usize,
+    pub embeddings_pending: usize,
+    pub embeddings_total: usize,
     /// There is no repository-v6 completion attestation yet. Counts are exact;
     /// completeness is deliberately not inferred from them.
     pub completion_attested: bool,
@@ -106,6 +109,9 @@ struct SemanticEnrichmentStatusWire {
     entity_count: usize,
     relation_count: usize,
     semantic_change_count: usize,
+    embeddings_indexed: usize,
+    embeddings_pending: usize,
+    embeddings_total: usize,
     #[serde(deserialize_with = "deserialize_status_unattested")]
     completion_attested: bool,
 }
@@ -124,6 +130,9 @@ impl<'de> Deserialize<'de> for SemanticEnrichmentStatus {
             entity_count: wire.entity_count,
             relation_count: wire.relation_count,
             semantic_change_count: wire.semantic_change_count,
+            embeddings_indexed: wire.embeddings_indexed,
+            embeddings_pending: wire.embeddings_pending,
+            embeddings_total: wire.embeddings_total,
             completion_attested: wire.completion_attested,
         };
         enrichment.validate().map_err(serde::de::Error::custom)?;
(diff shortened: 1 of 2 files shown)


troyjr4103:
Adversarial review at head 1cd20da7: **NO-GO, 2 blocking findings**.

The direction is compliant (RepositoryAuthorityState is graph-owned truth; no file surface, no side counter). The blocking problem is that the plumbed numbers are structurally guaranteed wrong: `InMemoryGraph::from_snapshot` never attaches a vector index, so `embedding_status().indexed` is unconditionally 0 and `pending` is unconditionally `total`. A fully embedded repository would report zero coverage forever in a well-formed payload, replacing today's loud probe failure with a quiet falsehood. A check that cannot fail is not evidence, and a number that cannot be right is worse.

Required before re-review:
1. Source the counts from a surface that actually carries index state (the authority payload stats path per the ticket), or attach the index before reading.
2. Do not ship any path where indexed=0/pending=total is the unconditional shape for an embedded store.
3. Emit the correct absent/unknown shape when the index is genuinely not loaded (absence must stay loud, not read as zeros).
4. Align field names with the served schema consumers expect.
5. Add the `pending < uncovered` consistency check matching graph-status.v1.
6. Store-backed regression test asserting nonzero `embeddings_indexed` on an embedded repository (red on this head).
7. Fix or delete the two doc comments this change falsifies, and account for the full-graph materialization cost this adds to `kin status`/`kin init` (that cost is the FIR-1761/1793 hot path).
8. Retitle to plain prose (the squash is minted from the title), reword the commit subject, and drop the closing keyword unless the ticket is fully resolved end to end.

Full verdict with file:line for each finding is in the fleet's review records (review-kin-pr551-20260801).

troyjr4103:
Adversarial review verdict at head `1cd20da7`: **NO-GO**.

Blocking:

- **The plumbed numbers are structurally wrong.** `InMemoryGraph::from_snapshot` never attaches a vector index, so `embedding_status().indexed` is unconditionally `0` and `pending` is unconditionally `total`. `kin status --json` would report zero embedding coverage on a fully embedded repository, forever, with a well-formed and plausible payload. FIR-1785's current probe fails loud on a missing enrichment object; this change replaces that loud absence with a quiet falsehood, which is a strict regression against the ticket's own standard (a check that cannot fail is not evidence).
- **Served-schema break with no version bump.** `SemanticEnrichmentStatusWire` carries `deny_unknown_fields`, and the three new fields are required (no `#[serde(default)]`) while the schema string stays `kin.status.v2`. Both parse directions break between shipped binaries.

Also required in the fix round: the added test passes identically against a hardcoded `(0, 0, 0)` and needs a real admitted-store assertion (a repository with embeddings must report nonzero indexed); the PR title and commit subject drop the conventional-commit prefix (house style is plain prose, see the last three squashes on main); and `Closes FIR-1785` becomes a plain reference, since partial work takes no closing keyword.

Worth keeping in the redo: sourcing from `RepositoryAuthorityState` rather than a file surface or a daemon side counter is the right direction and clears the zero-file-search rule; the diff discipline is tight; field naming matches the module.

This PR stays queued but deferred from admission until the fix round lands and re-reviews.

troyjr4103:
Closing as superseded by merged PR #577. Source verification confirms this branch rebuilds `InMemoryGraph` from an authority snapshot whose vector index is structurally absent, so it can emit the plausible-but-false triple `indexed=0, pending=N, total=N` for a fully indexed repository. PR #577 reads coverage from the live graph that actually carries the validated vector sidecar and includes adversarial coverage for this exact failure mode.

PR #555, opened 2026-07-31
Make the pin prover green on what it finds, and prove it on the substrate rather than in a comment

troyjr4103:
The P1-1 remedy that landed in #545 was inert. This makes it real and proves it
on the substrate instead of asserting it in a comment.

Closes FIR-1794.

Spec: `.kin-coord/reports/audit-kin-pr545-postmerge-20260731.md` (VERDICT UNSOUND).

**Nothing has broken.** `base-image-pins.yml` is `schedule` + `workflow_dispatch`
only, `cron: "23 5 * * 1"`, and has **never run** (`total_count=0`). First fire
is Monday 05:23 UTC.

## What was wrong

The workflow claimed the scheduled path could not conclude non-green, and gave
as its reason that `set -e` was absent. It was never absent: a `run:` step's
script is executed by **`bash -e {0}`**, and the step's own `set -o pipefail` is
what completed the trap by making the pipeline carry the prover's non-zero
status for errexit to act on. The capture, the `$GITHUB_OUTPUT` write, the
tracking issue and the dispatch reporter were all unreachable. Net effect: worse
than before the fix, because a broken pin would produce a red check-run on a
main sha **and** no alarm.

Reproduced verbatim (landed body, documented shell, stub prover exiting 2):

```
STEP RC=2                       <- red
GITHUB_OUTPUT contents: []      <- never written
```

## The obvious remedy does not work, and is measured not to

Both the original review and the audit recommended job-level
`continue-on-error: true`. I ran the shapes on a scratch branch in this repo and
read the **check-run conclusions off the commit**, which is what
`release-tag.yml` sweeps (`non_failing_optional = {success, skipped, neutral}`):

| shape | check-run conclusion |
| --- | --- |
| job `continue-on-error: true` + failing step | **failure** |
| no `continue-on-error` + failing step | failure |
| job `continue-on-error: true` + exceeded `timeout-minutes` | **cancelled** |
| the step body in this PR, prover exit 1 | **success** |

Job-level `continue-on-error` only stops the *workflow run* being marked failed.
It would have looked like protection while changing nothing, so it is
deliberately **not** used, and the header records why.

## What this PR does

**P1-1 + N1 (load-bearing).** The verify step declares its shell explicitly with
no `-e`, calls `set +e` in its own body, captures the status, and ends in
`exit 0`. Either layer alone suffices; both exist because the second is what
survives someone deleting the first. `$GITHUB_OUTPUT` is written on every path,
and the alarm steps run under `always()` so they do not depend on earlier steps
having succeeded.

**N2 (job-level paths), honestly bounded rather than claimed away.**

- failed checkout: neutralised with step-level `continue-on-error`
- timeout: made *unreachable* by bounding work, not by tolerating it. Every
  registry call is `timeout 20`, retries drop 5 → 3, worst case ≈ 7.5 min
  against `timeout-minutes: 20`
- runner/infrastructure failure: **irreducible**, and the header says so. The
  complete fix is the sweep ignoring non-required producers, which is
  release-authority work and out of scope here

**N3.** The header now states the measured posture, separated into what is
guaranteed and what is not, and names the errexit mistake so the next reader
does not repeat it.

**N4.** `status == 2` emits a `::warning::` and a job-summary line, so a
permanently throttled probe cannot masquerade as a healthy one.

**N5.** The `*404*` match is anchored to a status token or whole phrase. The
error text embeds the reference, so a pinned digest containing `404` in its hex
(~1 in 70) turned *any* error, a throttle included, into a definitive absence.

**N7**: an authority-suite line for the aarch64 `sudo apt-get update` was added,
but it matches at job scope rather than pinning the intended step, so it does not
yet prove that guard. Proper scoping is recorded as follow-up.

## Acceptance, end to end on the real substrate

A probe branch ran the workflow with the step body, its shell, every
`continue-on-error` and every `if:` **verbatim**, changing only the trigger and
stubbing the prover to exit 1:

| criterion | result |
| --- | --- |
| check-run conclusion on the sha | **`success`** |
| tracking issue filed | **yes** (#554, since closed) |
| `Close the tracking issue` step | correctly skipped |
| dispatch reporter, event clause satisfied | **`failure`**, still fails hard |

Also verified locally that the fixed body survives `bash -e` (the shell that
killed the previous version) with `STEP_RC=0`, `status=1` written and the
summary line emitted, so the `set +e` layer alone is sufficient.

## Regression guard

The properties this depends on are now pinned by the release-authority suite,
because this mechanism has now been wrong twice and a guard whose green is
untested is exactly the shape being fixed. Five in-suite falsifications: the
shell regaining `-e`, the shell declaration vanishing, the body losing `set +e`,
the body losing `exit 0`, and the checkout losing its tolerance.

## Gates

`test-release-workflow-authority.py` rc=0 · `py_compile` rc=0 ·
`bash -n` rc=0 · `shellcheck -x` rc=0 · both zero-file-search guards pass ·
`actionlint` findings byte-identical to `main` · live prover run against the
committed Dockerfile: both pins `docker.io=ok mirror.gcr.io=ok tag=current`,
rc=0 · no `.rs`, no required-context names touched, no census change.

## Claims not being made

- Not claiming the job cannot conclude non-green. Runner/infrastructure failure
  still can, and `workflow_dispatch` still fails hard by design.
- Not claiming the timeout path was executed. It is made unreachable by budget,
  and the `cancelled` conclusion it would produce was measured separately.
- Not claiming anything about the Dockerfile pins, the BuildKit mirror wiring,
  or the aarch64 step, which are unchanged and were cleared by both prior passes.

### DIFF
--- a/.github/workflows/base-image-pins.yml
+++ b/.github/workflows/base-image-pins.yml
@@ -6,28 +6,58 @@ name: Base Image Pins
 # A scheduled run's check-runs DO land on a commit. GitHub binds a check suite
 # to every workflow run, and for a `schedule` event `github.sha` is the default
 # branch HEAD at fire time, so a red scheduled run publishes a failure check-run
-# on a main commit. release-tag.yml's second sweep refuses a release commit
-# carrying any non-green check-run whether or not a ruleset requires it, and a
-# quiesced main during a release train holds HEAD still across exactly the
-# window that matters. This repo already carries a live instance of that shape:
-# fuzz.yml's scheduled run left two `failure` check-runs on a main sha.
+# on a main commit. release-tag.yml's second sweep treats only `success`,
+# `skipped` and `neutral` as non-failing, so anything else on a release
+# candidate refuses that release terminally. A quiesced main during a release
+# train holds HEAD still across exactly that window. fuzz.yml is the standing
+# example: its scheduled run left two `failure` check-runs on a main sha.
 #
-# Dropping the push and pull_request triggers therefore does NOT make this job
-# safe on its own. What makes it safe is that the scheduled path is
-# structurally incapable of concluding failure: the prover's exit status is
-# captured rather than propagated, and the step exits 0 whatever it found. A
-# registry-reachability probe must never be able to refuse a release, because
-# that is the failure class this workflow exists to watch.
+# So this job must not conclude non-green on what it FINDS. What follows is the
+# measured posture, not the intended one. An earlier version of this file
+# claimed the same protection and did not have it, because a `run:` step's
+# script is executed by `bash -e {0}`: `set -e` was never absent, and adding
+# `set -o pipefail` is what armed it. Every claim below was checked by running
+# the shapes on a scratch branch and reading the check-run conclusions off the
+# commit.
 #
-# Loudness comes from a surface that cannot poison a mint. When a pin is
-# definitively unreachable the run opens or updates a tracking issue, which
-# persists until someone acts on it, unlike a red check nobody is paged for.
+# What IS guaranteed:
 #
-# A `workflow_dispatch` run may fail hard, because an operator asked a direct
-# question and deserves a direct answer. The residual is real and is theirs to
-# own: dispatching against main during a release train can publish a failure
-# check-run on the candidate sha. Prefer the schedule unless you need the
-# answer now.
+#   - No result the prover can report concludes non-green. The step runs under
+#     a shell carrying no `-e`, disarms errexit again in its own body, captures
+#     the prover's status instead of propagating it, and ends in `exit 0`.
+#     Measured: that shape publishes `success` with a prover exit of 1, and the
+#     alarm step still runs.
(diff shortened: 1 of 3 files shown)


troyjr4103:
Adversarial review at head 608d4182: **GO.**

Verified against the FIR-1794 acceptance bar (the PR 545 post-merge audit): the fixed step body was executed verbatim under the documented Actions shell with a stubbed failing resolve and concludes SUCCESS while the alarm surface fires; persistent failure behaves the same on schedule; workflow_dispatch still fails hard by design; the header comment now states the true posture. The timeout path is bounded by a recomputed budget. Non-blocking findings are recorded in the fleet verdict (review-kin-pr555-20260801), including two latent non-green paths named for a future touch.

Sequencing: the v0.4.5 release commit (PR 557) is entering the queue under quiesce; NOTHING lands on kin main between it and the tag. This PR is queued by the captain immediately after the tag exists, which keeps it comfortably ahead of the Monday 05:23 UTC first cron.

PR #566, opened 2026-08-01
Bind the MCP commit precondition to the authority revision it plans against

troyjr4103:
## The precondition semantics, and the ruling behind them

The exact MCP commit path gated every transaction on the daemon's live query
graph being byte-equal to the clean workspace authority it had just loaded, over
`entities`, `relations`, and `resolved_tree`.

That invariant is unachievable by design, and this PR proceeds from that as a
settled ruling rather than reopening it. The daemon's query graph legitimately
leads durable authority: the reconcile loop publishes only the tree through the
repository compare-and-swap and leaves parser semantics in the live graph, and
the asynchronous LSP enrichment worker upserts relations into the live graph
under nothing but the graph-authority epoch (`kin-daemon/src/daemon.rs:680`,
holding neither `coordination_gate` nor the persistence lock). Both reach
authority only when a change is committed. So the precondition held for a moment
after each commit and then failed at the next enrichment tick, refusing ordinary
agent commits with a message about absorbing an unrelated dirty overlay.

**Equality was never what kept derived enrichment out of the published change.**
`plan_exact_transaction` builds its prospective graph from `base.graph`, which is
repository authority, and applies only the staged operations before
`plan_native_commit_from_base`. A derived lead in the live graph cannot reach
publication regardless of what the precondition says. The guard's stated purpose
was already served by the planner.

What the commit does depend on is the *binding*, which the old guard never
checked. The live graph matters after the commit, because `install_authority_graph`
computes `transaction_delta_between(live, authority)` and corrects the daemon
graph from the live side. For that correction to mean anything, three properties
must hold at plan time:

1. the daemon is at the same authority generation this plan is loading,
2. it holds no exact tree state authority has not admitted,
3. it has neither dropped nor rewritten anything that generation owns.

A derived *lead* is fine. A derived *loss* is not.

That is exactly `require_fresh_daemon_workspace`, the boundary checkout, branch,
merge, and rollback already answer to. The MCP path now shares it rather than
carrying a second, stricter definition of freshness that only it could fail.
`NativeCommitBase` already carried `roots: RootBundle`, so binding the revision
needed no new plumbing.

The post-commit sites are unchanged, as predicted: `semantic_workspace_matches`
survives for no-op detection (both sides authority-derived) and for the
post-correction assertion.

## Why reusing that boundary is sound here

`state.snapshot_generation` is the repository authority roots generation on this
path. It is initialized from `lease.roots().generation` (`state.rs:1833`) and
advanced only by `record_repository_authority_commit`. The two sites that store a
*backend* generation instead (`state.rs:4004`, `4045`) are both inside the
`storage_backend.is_some()` branch, and `commit_exact_transaction_inner` returns
early for storage-backend daemons. `state.rs:4084-4090` states the same
invariant in its own words.

## Refusals now name the binding and the exit

Each refusal is wrapped with the transaction id and `No repository authority
moved and the staged operations are untouched, so re-send this commit unchanged
once the daemon is reading current repository authority.` That is verified, not
asserted: the precondition sits strictly before `prepare_transaction_commit` and
does not route through `unstage_failed_attempt`, so the staged set survives, and
the tests check the transaction is still `active` after each refusal.

One shared message was reworded. `require_fresh_daemon_workspace` interpolated
its gerund operation phrase as `but {operation} authority is at generation`,
which reads as "but checking out a path authority is at generation N" for every
existing caller. It now reads `but the authority for {operation} is at
generation`. No test asserted the old wording.

## Evidence

Four tests added to `mcp_commit::tests`, run red against the old precondition and
green against the new one.

RED, old precondition, tests unchanged:

```
test result: FAILED. 33 passed; 3 failed; 0 ignored; 578 filtered out
  a_commit_after_an_enrichment_tick_publishes_only_its_staged_operations
  a_commit_from_a_daemon_missing_authority_owned_semantics_is_refused
  a_commit_planned_against_a_trailing_daemon_cursor_is_refused
```

Two things that run made visible:

- The old guard returned the **identical** message for the legitimate
  post-enrichment commit and for the genuine authority-loss case. It could not
  distinguish them, which is why it wedged sessions instead of protecting them.
- The trailing-cursor case *succeeded* under the old guard, because it never
  inspected the authority revision at all. Stated precisely: that is a
  constructed case, not a live hole, since real external authority movement would
  also diverge the graph and trip the equality check with a misleading message.
  The claim is that the old guard had no revision binding, not that a soundness
  hole was being exploited.

GREEN, reworked precondition: `test result: ok. 37 passed; 0 failed`.

The guards are non-vacuous by construction. Both acceptance tests assert the live
graph matches authority *before* the tick and diverges *after* it, so a test that
stopped reproducing the divergence fails rather than passing silently.

Gates at final head (`RUSTFLAGS=-D warnings` throughout):

| gate | rc |
| --- | --- |
| `cargo fmt -- --check` | 0 |
| clippy, ci.yml allowlist, `--all-targets --all-features` | 0 |
| `cargo build --all-targets --locked` | 0 |
| `scripts/check_runtime_boundaries.sh` | 0 |
| `cargo nextest run --workspace` | not run (daemon lock) |
| `cargo test --doc` | not run (daemon lock) |
| `cargo test --workspace` (the authority) | not run (daemon lock) |

The three workspace test gates are queued behind another lane's daemon lock and
will be reported here when they land. The focused `mcp_commit` module suite is
green at this head (37 passed, 0 failed).

## What this does not claim

No live-daemon acceptance was run. FIR-1633 asks for one on the grounds that
in-process coverage does not reproduce enrichment timing; these tests instead
place the tick deterministically at the point that matters, in the exact write
shape the worker uses. That is a deviation from the written acceptance, so this
PR references FIR-1633 without a closing keyword and leaves the call to review.

Worth recording for whoever builds that live reproducer, because it narrows the
work: the enrichment worker is the only reachable source of this divergence, so
the reproducer needs rust-analyzer and not merely a live daemon. The other
divergence source, parser semantics left behind by reconcile, cannot reach this
precondition, because divergent parser semantics imply changed bytes, changed
bytes move the exact tree, and a moved tree makes the workspace dirty, which
`load_native_commit_base` refuses one step earlier.

The LSP worker's missing `coordination_gate` is not fixed here. A relation upsert
can still interleave with commit finalization. The reworked precondition
tolerates the resulting lead by design, but it does not serialize the writer.

FIR-1611 was assessed as a possible dependency and is not one, so nothing here
closes it. It binds storage namespace identity; this binds authority revision.
Two findings from that assessment are worth recording, because the second
changes what FIR-1611 is waiting on:

- The mislabeling is still live on main. `revalidate_pinned_local_namespace`
  rides `load_snapshot_authority(...)?` wholesale and `open_bound` maps all of
  its errors to `Identity`, so a missing `.lock` or a truncated snapshot on an
  intact namespace is answered as a replaced repository.
- It is **not** blocked on a kin-db API. The pinned `kin-db = "=0.7.8"` already
  exposes `LocalFileBackend::probe_pinned_repository_namespace` returning a typed
  `LocalNamespaceProbe`, documented to take no lock and decode no snapshot, which
  covers the single-load-bind item too. The kin-side consumer already exists as
  **PR 471**, whose own body gates it on "kin-db 0.6.5 publishing". That gate
  lifted long ago: 471 bumps kin-db to 0.6.6 while main is on 0.7.8. FIR-1611 is
  waiting on a rebase of PR 471, not on a design or a kin-db mint.

PR 471 also edits `crates/kin-daemon/src/local_repository_authority.rs`, in a
different region than this PR does. PR 562 edits `crates/kin-daemon/src/mcp_commit.rs`,
in an unrelated `kin history` test. Neither looks like a textual conflict, but
they are the same files, so they should not be batched on an assumption of
independence.

### DIFF
--- a/crates/kin-daemon/src/local_repository_authority.rs
+++ b/crates/kin-daemon/src/local_repository_authority.rs
@@ -153,8 +153,10 @@ impl ActiveLocalRepositoryAuthority {
 /// same authority generation, it holds no exact tree state authority has not
 /// admitted, and it has neither dropped nor rewritten any entity or relation
 /// that authority owns. Because a derived lead is permitted, every caller must
-/// plan its daemon-side transition from the live graph via
-/// [`plan_daemon_semantic_delta`] rather than reusing the authority-side delta.
+/// plan its daemon-side transition from the live graph rather than reusing the
+/// authority-side delta: repository commands do that through
+/// [`plan_daemon_semantic_delta`], and the exact MCP commit path does it through
+/// its own post-commit live-to-authority correction.
 pub(crate) fn require_fresh_daemon_workspace(
     state: &DaemonState,
     roots: &RootBundle,
@@ -164,8 +166,8 @@ pub(crate) fn require_fresh_daemon_workspace(
     let daemon_generation = state.snapshot_generation.load(Ordering::SeqCst);
     if daemon_generation != roots.generation {
         bail!(
-            "daemon repository cursor is at generation {daemon_generation}, but {operation} \
-             authority is at generation {}; reopen from repository authority before mutating",
+            "daemon repository cursor is at generation {daemon_generation}, but the authority for \
+             {operation} is at generation {}; reopen from repository authority before mutating",
             roots.generation
         );
     }
--- a/crates/kin-daemon/src/mcp_commit.rs
+++ b/crates/kin-daemon/src/mcp_commit.rs
@@ -25,7 +25,9 @@ use kin_model::{
 };
 use sha2::{Digest, Sha256};
 
-use crate::local_repository_authority::LocalRepositoryAuthorityContext;
+use crate::local_repository_authority::{
+    require_fresh_daemon_workspace, LocalRepositoryAuthorityContext,
+};
 use crate::repository_commit::{
     commit_native_plan_with_projection, load_native_commit_base, load_native_source_blob,
     plan_native_commit_from_base, recover_native_commit, NativeCommitBase, NativeCommitResult,
@@ -193,7 +195,7 @@ fn commit_exact_transaction_inner(
 
     let base = load_native_commit_base(&authority_context)
         .map_err(|error| format!("load exact MCP commit base: {error}"))?;
-    require_live_graph_matches_authority(state.graph.as_ref(), &base.graph)?;
+    require_bound_authority_revision(state, &base, &transaction_id)?;
     let plan = match plan_exact_transaction(
         state,
         &authority_context,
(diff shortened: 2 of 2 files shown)


troyjr4103:
CI is red on `Check & Test`, and it is not this change. Recording the reasoning rather than just re-running.

Two failures appeared, on two different tests in two different crates:

- macos-latest `Check & Test`, step `Test`: `kin-core registry::tests::restrictive_umask_and_relative_path_are_isolated`, panicking with `failed to anchor registry process lock: Permission denied (os error 13)` at `crates/kin-core/src/registry.rs:2170`. Summary `1798/4653 tests run: 1797 passed, 1 failed`.
- Local `cargo nextest run --workspace`: `kin-cli commands::update::tests::replacing_install_root_aborts_before_component_mutation`, panicking on `assertion failed: format!("{error:#}").contains("binding changed")` at `crates/kin-cli/src/commands/update.rs:12728`.

The decisive point is structural. Neither failing crate depends on `kin-daemon`:

```
crates/kin-core/Cargo.toml:63   kin-daemon-spawn = { workspace = true }
crates/kin-cli/Cargo.toml:32    kin-daemon-spawn = { workspace = true }
```

Both depend on `kin-daemon-spawn`, a separate crate this PR does not touch, and neither manifest carries a `kin-daemon` dependency. This diff is confined to `crates/kin-daemon/src/mcp_commit.rs` and `crates/kin-daemon/src/local_repository_authority.rs`, so the changed code is not linked into either failing test binary at all.

Supporting: the macos-failing test passes locally on this exact branch in isolation (`registry::tests::restrictive_umask_and_relative_path_are_isolated ... ok`), and PRs 559, 560, 561, and 562 are all green on `Check & Test (macos-latest)`. Both failures are filesystem, lock, and inode-atomicity security tests, which is the class that reacts to runner and scheduling conditions.

The evidence pointing the other way, stated rather than left out: main is green on its last twelve runs including at this branch's base `93a74a91`, so this is not a standing breakage. The reading consistent with everything above is a flake, and a re-run is what confirms it. Failed jobs are queued to re-run once the workflow finishes (`Windows authority tests` was still in flight, and `gh run rerun --failed` refuses while a run is in progress).

The ubuntu-latest job reads as failed but is cancelled collateral from matrix fail-fast: its steps are green through formatting, the Zero File-Search invariant and its falsification, hydration replay, clippy, runtime boundaries, private repo coupling, `Build`, and the Linux musl target, with the aarch64 musl step cancelled and everything after it skipped.

PR #614, opened 2026-08-04
Release Kin v0.5.0

troyjr4103:
## Pin moves folded in (captain amendment, head 8d5e6304)

This is the combined release change: alongside the version bump it moves kin-model `=0.7.4` to `=0.7.5` and kin-db `=0.7.11` to `=0.7.12`, carrying the FIR-1925 admission fix (legacy zero-padded tree modes; flask-class repositories) into the release. Root and fuzz locks re-locked patch-off; every moved row keeps its `sparse+` registry source and checksum. Landing this PR makes its squash the v0.5.0 tag anchor with the fix resolvable; the mint dispatcher independently refuses to tag a squash whose manifest or lock lacks these pins.

## What moved

One commit, the same nine-file shape as #603:

- **Five lockstep version surfaces** 0.4.9 → 0.5.0: `Cargo.toml`
  `[workspace.package]`, the `kin-spine` path-dependency pin in
  `crates/kin-cli/Cargo.toml`, and `packages/kin/package.json`,
  `packages/kin-mcp/package.json`, `packages/boundary-contracts/package.json`.
  `kin-db` is untouched — it tracks an external registry crate.
- **Both lockfiles refreshed** with `cargo update --workspace`. The movement is
  workspace members only: 23 rows in `Cargo.lock`, 1 in `fuzz/Cargo.lock`, no
  external dependency resolution changed. Every external `kin-*` row keeps its
  `sparse+https://kinlab.ai/registry/cargo/` source and checksum
  (`kin-db 0.7.11`, `kin-model 0.7.4`, `kin-lsp 0.6.5`, `kin-infer 0.2.43`,
  `kin-vfs-core 0.3.0`, `kin-blobs 0.1.4`, `kin-search 0.1.4`,
  `kin-vector 0.1.8`).
- **A 0.5.0 changelog section** carrying the five changes this release ships,
  plus the install proof's Windows leg becoming experimental and
  non-release-gating (that change lands separately and ships here).
- **The reviewed v0.4.9 abandonment record** in
  `scripts/abandoned-release-tags.json`, superseded by v0.5.0.

## Why

The v0.4.9 lane cannot finish at the tag. Its public install proof
(run 30947137583) failed all three attempts, and automatic release recovery
exhausted its bounded retries:

- `macos-15-intel` and `ubuntu-24.04-arm` fail the installed-capability proof
  with `Unix status did not prove complete observed embedding coverage:
  {"state":"unobserved","reason":"graph_mutation_in_flight"}` — a transient
  window in which the graph is still being mutated, read as a failure to prove
  coverage rather than waited out. It is a race, not a platform property, which
  is why a third Unix leg failed on the first attempt and passed later.
- `windows-latest` fails separately at the graph and MCP tool-call proof, where
  the daemon port file it reads never appears.

Both the proof workflow and the binary resolve from the tag, so no change to
main can reach those steps. The tag's only release object is the unpromoted
prerelease published before proof, so it never became Latest; Latest is still
v0.3.6. v0.5.0 ships the batch that fixes the race.

## Verification

- `node scripts/release-intent.mjs` → `v0.5.0 is a coherent release — ready to
  tag` (workspace, all three npm packages, the `kin-spine` pin, and the
  changelog section all agree at 0.5.0).
- `cargo metadata --locked` clean for both the root workspace and `fuzz/`.
- Selector self-tested against the real tag listing:
  reconcile and mint-v0.5.0 both anchor v0.3.6; minting v0.4.9 is refused as an
  abandoned tag; a corrupted record sha is refused loudly on the waiver/tag
  mismatch.
- `scripts/extract-release-notes.mjs --version 0.5.0` rolls the abandoned
  v0.4.9, v0.4.8, v0.4.6, v0.4.5, v0.4.4 sections forward into the v0.5.0 notes
  under the carried-forward notice, so no burned lane's notes are lost.
- `python3 scripts/test-release-workflow-authority.py` passes; the release
  policy node suite passes 67/67.

## Sequencing: this must land last

The tag mint normalizes a tag to the **first** reviewed first-parent main commit
carrying the workspace version (`.github/workflows/release-tag.yml`, the
`automatic` branch). If this bump lands before #604, #605, #606, #608, #609 and
#612, v0.5.0 tags this commit and excludes every change it claims to ship — the
v0.4.9 failure verbatim. Land those six first, then this.

The branch is currently based on `0a95bec8`, the tip of main at the time it was
cut. The batch does not touch any of these nine files, so the hosted queue's
rebase is clean.

## Notes for the captain

- The Windows-leg changelog entry cites #612 ("Declare the Windows install-proof
  leg experimental and pin that posture"), which is open. #612 must land before
  this PR for that line to be true.
- v0.4.7 is untagged and therefore not in the abandonment record, so its
  changelog section is not carried into these notes. Pre-existing, and worth a
  decision separately from this release.

Ticket linkage handled at merge time by the captain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.5.0] - 2026-08-04
+
+### Changed
+
+- Bind the deferred-uninstall wait to the child and wait for its helper to complete (#604)
+- Free the release mint from advisory flakes and stop blind release retries (#605)
+- Read the host-binding fixtures as scripts rather than exec targets (#606)
+- Serve entity source projections from held repository authority (#608)
+- Settle status embedding coverage over a transient unobservable window (#609)
+- Declare the Windows install-proof leg experimental and pin that posture (#612)
+
 ## [0.4.9] - 2026-08-04
 
 ### Changed
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3055,14 +3055,14 @@ dependencies = [
 
 [[package]]
 name = "kin-buildinfo"
-version = "0.4.9"
+version = "0.5.0"
 dependencies = [
  "sha2 0.11.0",
 ]
 
 [[package]]
 name = "kin-cli"
-version = "0.4.9"
+version = "0.5.0"
 dependencies = [
  "age",
  "anyhow",
@@ -3141,7 +3141,7 @@ dependencies = [
 
 [[package]]
 name = "kin-context"
-version = "0.4.9"
+version = "0.5.0"
 dependencies = [
  "kin-blobs",
  "kin-db",
@@ -3157,7 +3157,7 @@ dependencies = [
 
 [[package]]
 name = "kin-core"
-version = "0.4.9"
+version = "0.5.0"
 dependencies = [
  "cap-fs-ext",
  "cap-std",
@@ -3190,7 +3190,7 @@ dependencies = [
 
 [[package]]
 name = "kin-daemon"
-version = "0.4.9"
+version = "0.5.0"
 dependencies = [
  "anyhow",
  "async-stream",
@@ -3246,7 +3246,7 @@ dependencies = [
 
 [[package]]
 name = "kin-daemon-spawn"
-version = "0.4.9"
+version = "0.5.0"
 dependencies = [
  "fs2",
  "libc",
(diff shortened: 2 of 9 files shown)


troyjr4103:
HELD, do not submit or drain: founder ruling 2026-08-04 holds the v0.5.0 mint until the FIR-1925 kin-model fix reaches kin's pin (announcement moved to Wednesday for it). This PR needs a rebase and lock refresh over the pin-bump PR before it lands. Captain session 2d087f0f holds the merge lease; the mint dispatcher's ladder guard also refuses to tag without the prerequisite ancestry.

troyjr4103:
PLAN UPDATE (compression, founder direction ~22:40Z): this PR becomes tonight's COMBINED release change. After kin-model 0.7.5 and kin-db 0.7.12 publish, the captain amends this branch to carry the two pin moves plus the existing version bump, locks regenerated once over the final main, and it lands in a single queue round as the tag-anchor commit. Still HELD until that amendment: do not submit or drain before the head moves and the captain announces readiness here.

PR #615, opened 2026-08-05
Build the Windows daemon the authority tests drive before running them

troyjr4103:
`Windows authority tests` has failed identically on `main` and on every pull
request for days: 8 passed, 1 failed, always `daemon_status_and_stop_lifecycle`.
It is not a required context, so PRs have been landing over it, it has already
produced two false "blocked" diagnoses, and while it stays red it masks any real
Windows regression the leg exists to catch.

The build-identity guard is right. The job is wrong.

## Correction to the standing diagnosis

The red was being read as a stale cross-commit `kin-daemon.exe` carried into the
test step by the restored cargo cache. The log says otherwise. Run
`30954050156`, job `92142801364`, at `c2c68358`:

```
thread 'daemon_status_and_stop_lifecycle' panicked at crates\kin-cli\tests\common\mod.rs:3196:9:
kin-daemon at D:\a\kin\kin\target\x86_64-pc-windows-msvc\debug\kin-daemon.exe is
unusable after rebuild: D:\a\kin\kin\target\x86_64-pc-windows-msvc\debug\kin-daemon.exe
does not exist.
```

The variable part of that panic is `{reason}`, and the reason is
`does not exist`. "A daemon built from a different commit or working tree" is
static template text in the panic string, not the reason. A stale binary would
have produced `daemon build identity X does not match this test binary's Y` from
`daemon_compat_mismatch`. It did not, so the cache is not implicated: there is
no cross-commit `kin-daemon.exe` reaching the test step, because there is no
`kin-daemon.exe` there at all.

## Failing chain, as it stands on main

1. `daemon_status_and_stop_lifecycle` calls `runtime.daemon_bin()`
   (`crates/kin-cli/tests/common/mod.rs:989`), which resolves the daemon as a
   sibling of `CARGO_BIN_EXE_kin`, so
   `target/x86_64-pc-windows-msvc/debug/kin-daemon.exe`.
2. Nothing in the job writes that path first. Every `kin-daemon` invocation in
   the authority step is `--lib`, and the step that builds `--bin kin-daemon`
   runs two steps AFTER the tests.
3. `daemon_compat` returns `Err("... does not exist")`, so `fresh_daemon_bin`
   self-heals by shelling out to `cargo build -p kin-daemon --bin kin-daemon`,
   with no `--target`.
4. Cargo writes a host build to `<target-dir>/<profile>/` and a `--target` build
   to `<target-dir>/<triple>/<profile>/`. That rebuild therefore landed in
   `target/debug/kin-daemon.exe`.
5. The re-check reads the msvc path again. Still nothing. Panic.

The timing corroborates it. The other 8 tests finish inside one second at
`22:25:33`; the panic lands at `22:31:26`. That ~5m53s is the wasted rebuild, a
full default-features `kin-daemon` compiled on the runner and written where
nothing reads it.

## Fix chain, and why the failure cannot recur

1. A new step builds
   `--target x86_64-pc-windows-msvc -p kin-daemon --no-default-features
   --bin kin-daemon` after the cache restore and before the authority tests,
   writing exactly the path the harness reads.
2. That binary comes from the same checkout as the test binary, so
   `kin_buildinfo::sha_with_dirty` agrees and `daemon_compat` passes on the
   first check. The self-heal never runs and the job gets those ~6 minutes back.
3. The feature set matches the existing binary build step. `kin-cli` is declared
   `default-features = false` in `[workspace.dependencies]`, so
   `-p kin-daemon --no-default-features` and
   `-p kin-cli -p kin-daemon --no-default-features` resolve `kin-cli`
   identically and both steps produce one artifact, leaving the admission and
   npm proof steps downstream untouched.
4. `scripts/test-release-workflow-authority.py` now pins the build's presence
   AND its position ahead of `"daemon_status_and_stop_lifecycle"`, because the
   build sitting behind its consumer is precisely what this bug was.
5. Independently, `fresh_daemon_bin` now repeats the `--target` its own binary
   was written under, so a self-heal that does fire lands where the check looks.
   The triple comes from `KIN_CLI_TARGET_TRIPLE`, emitted by
   `crates/kin-cli/build.rs` from cargo's `TARGET`, compared against the path
   layout. This makes the guard's stated remedy true rather than decorative and
   covers cross-target runs outside CI.

The guard is not weakened. No test is skipped, filtered, or ignored, and the job
still runs on every event with no job-level `if:`.

## Falsification

| gate | mutation | result |
|---|---|---|
| authority suite | `ci.yml` reverted to `origin/main` bytes | exit 1, `native Windows daemon lifecycle prerequisite is missing required policy` |
| authority suite | `ci.yml` as fixed | exit 0 |
| authority suite | build block deleted, in-suite `expect_assertion` | raises, matched |
| authority suite | build block moved after the tests, in-suite `expect_assertion` | raises `before the lifecycle test reads it`, matched |
| harness unit test | `daemon_rebuild_target` stubbed to always return `None`, ie. pre-fix behavior | FAILED, `left: None, right: Some("x86_64-pc-windows-msvc")` |
| harness unit test | as fixed | ok |

The cargo layout rule in step 4 of the failing chain was measured before the
derivation was written: a probe crate built both ways puts `CARGO_BIN_EXE_probe`
at `target/debug/probe` and `target/aarch64-apple-darwin/debug/probe`.

## Local gates

`python3 ./scripts/test-release-workflow-authority.py` exit 0 ·
`actionlint .github/workflows/ci.yml` exit 0 · `cargo fmt -- --check` exit 0 ·
`python3 scripts/verify-zero-file-search.py` exit 0 ·
`bash scripts/zero_file_search_guard.sh` exit 0 ·
`cargo build --all-targets --locked` exit 0 ·
`cargo clippy --all-targets --all-features -- <ci.yml allowlist>` under
`RUSTFLAGS=-D warnings` exit 0 · the new harness unit test passes.

Not run locally, stated rather than glossed: the full
`cargo test -p kin-cli --no-default-features --lib --tests` step. Another lane
held the `daemon` and `gpu` locks for the whole window and that step spawns real
daemons. CI runs it here.

## What is proven where

The mechanism is verified locally and from the CI log. The FIX is verified by
CI, because the failure only manifests on a Windows runner and this lane has
none. `Windows authority tests` runs on `pull_request`, so THIS PR's own run is
the live test, not a proxy for it. The merge-group run is a second confirmation,
not the first.

Refs FIR-1929

---

**Post-review hardening (folded into the post-616 rebase).** The reviewer falsified two gaps in the new gate by execution: deleting `--target` from the sibling build left the suite green (the gate could not catch reintroduction of the very defect this PR fixes), and the ordering check indexed a bare substring over the whole job, so a comment quoting the test name inverted it into a false red. The gate now judges active lines only, scopes its policies to the sibling build step's own block (pinning `--target x86_64-pc-windows-msvc` inside it), and compares step-name positions for ordering. Falsified in all three directions on this head: suite passes on the real tree; deleting `--target` fails with "missing required policy: --target x86_64-pc-windows-msvc"; rewriting the explanatory comment to quote the test name stays green. The suite's existing removed-block and block-after-tests negative controls pass unchanged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,6 +167,22 @@ jobs:
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
+      # `daemon_status_and_stop_lifecycle` drives a real daemon, and the harness
+      # takes it from the `kin-daemon` beside the `kin` binary the test was
+      # built with. Every other `kin-daemon` invocation below compiles `--lib`,
+      # and the step that builds the binary runs after the tests, so nothing put
+      # a `kin-daemon.exe` in the msvc directory before the test read it. The
+      # harness covers a missing daemon by rebuilding one, which lands beside
+      # the test only when it carries the same `--target`, so a leg that omits
+      # this build depends on that rebuild rather than on its own inputs. The
+      # feature set matches the binary build below, so both produce one artifact.
+      - name: Build the sibling Windows daemon the authority tests drive
+        shell: bash
+        run: |
+          set -euo pipefail
+          cargo build --locked --target x86_64-pc-windows-msvc \
+            -p kin-daemon --no-default-features --bin kin-daemon
+
       - name: Compile and run native Windows authority tests
         shell: bash
         run: |
--- a/crates/kin-cli/build.rs
+++ b/crates/kin-cli/build.rs
@@ -26,4 +26,12 @@ fn main() {
     if std::env::var("CARGO_CFG_TARGET_ENV").as_deref() == Ok("msvc") {
         println!("cargo::rustc-link-arg-bins=/STACK:16777216");
     }
+    // Cargo writes a `--target` build to `<target-dir>/<triple>/<profile>/` and
+    // a host build to `<target-dir>/<profile>/`, and tells the compiled code
+    // neither which of the two it is nor what triple it was built for. The test
+    // harness reuses the `kin-daemon` sitting beside its own `kin` binary and
+    // has to be able to rebuild one into that same directory, so it needs the
+    // triple to reproduce the layout it is reading from.
+    let target = std::env::var("TARGET").expect("cargo sets TARGET for build scripts");
+    println!("cargo::rustc-env=KIN_CLI_TARGET_TRIPLE={target}");
 }
--- a/crates/kin-cli/tests/common/mod.rs
+++ b/crates/kin-cli/tests/common/mod.rs
@@ -3163,6 +3163,59 @@ fn daemon_compat(runtime: &IsolatedDaemonRuntime, path: &Path) -> Result<(), Str
     }
 }
 
+/// The `--target` a `kin-daemon` rebuild has to carry to land beside `kin`.
+///
+/// Cargo writes a build to `<target-dir>/<profile>/`, and moves it to
+/// `<target-dir>/<triple>/<profile>/` when it is given `--target`. The daemon
+/// this harness demands sits beside `CARGO_BIN_EXE_kin`, so a rebuild that
+/// drops the triple the test binary was built for writes a directory the
+/// compatibility check never reads: the rebuild succeeds and the check still
+/// reports the daemon missing, naming a path nothing ever wrote.
+///
+/// The triple is only in the path when cargo put it there, which is exactly
+/// when the rebuild needs it, so the layout answers the question directly.
+fn daemon_rebuild_target<'a>(kin_bin: &Path, triple: &'a str) -> Option<&'a str> {
+    let profile_parent = kin_bin.parent()?.parent()?;
+    (profile_parent.file_name()?.to_str() == Some(triple)).then_some(triple)
+}
+
+#[test]
+fn daemon_rebuild_follows_the_layout_its_own_binary_was_written_into() {
+    let triple = "x86_64-pc-windows-msvc";
+
+    assert_eq!(
+        daemon_rebuild_target(Path::new("/w/target/debug/kin"), triple),
+        None,
+        "a host build already writes the directory the check reads"
+    );
+    assert_eq!(
+        daemon_rebuild_target(
+            Path::new("/w/target/x86_64-pc-windows-msvc/debug/kin.exe"),
+            triple
+        ),
+        Some(triple),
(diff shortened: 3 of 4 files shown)


troyjr4103:
Feature parity between the new build step and the existing binary build, measured rather than reasoned, with `cargo tree -f '{lib} FEATURES=[{f}]'`:

| selection | `kin_cli` | `kin_db` |
|---|---|---|
| `-p kin-daemon` (control, defaults on) | `[embeddings,metal,vector]` | `[accelerate,embeddings,metal]` |
| `-p kin-daemon --no-default-features` | `[]` | `[accelerate,embeddings,metal]` |
| `-p kin-cli -p kin-daemon --no-default-features` | `[]` | `[accelerate,embeddings,metal]` |

The first row is a control, and it is here because the first version of this probe grepped for a string `cargo tree` never emits, so it answered "identical" for every input including inputs that are not. A probe that cannot separate the two cases is not evidence.

Rows 2 and 3 agree, so the new step and the existing one at the end of the leg produce one artifact rather than two competing ones, and the admission and npm proof steps keep driving the same build they drive today.

troyjr4103:
Bisection on `main` pushes, which attributes the red exactly and bounds what this PR can claim.

`daemon_status_and_stop_lifecycle` was added to the Windows leg by c512114e (#582), landed 2026-08-02 18:52Z.

| main push | `Windows authority tests` |
|---|---|
| 6 consecutive runs before c512114e (through 2026-08-02 00:29Z) | success, all 6 |
| 12 consecutive runs since, 35ca3f7e through 5d64c822 | failure, all 12 |

So the leg was healthy until the lifecycle test arrived without the daemon binary build it needs, and it has been red on every main push and every PR since. Two consequences:

1. The rest of the leg is sound and was proven green as recently as 2026-08-02, including the binary build and the admission and npm proofs downstream. This PR does not need to defend those.
2. `daemon_status_and_stop_lifecycle` has never once executed on Windows. Every one of those 12 failures panicked before the test body ran. This PR is the first run where it actually executes, so its Windows behavior is genuinely unknown until this run reports.

That second point is the honest limit of the claim here. If the test now runs and fails for a real Windows reason, that is a first-time finding rather than a regression from this change, and it is mine to follow.

PR #616, opened 2026-08-05
Ship native Windows semantic vector search

troyjr4103:
Windows was the only release leg built `--no-default-features`, so the published
`kin-windows-x86_64.zip` carried no vector or embedding support at all. Every platform now
builds the same default feature set, the assertions that pinned the vector-free shape are
inverted in both directions, and native Windows semantic search is proven by execution rather
than by compilation.

## The matrix key was doing two unrelated jobs

`skip_vector: true` on the Windows row selected `--no-default-features` **and** gated
`Build kin-vfs (native)` at `release.yml:593`. Deleting that one key would also have started
building an LD_PRELOAD/DYLD libc interceptor for `x86_64-pc-windows-msvc`.

It is split rather than deleted: the Windows row now carries `skip_vfs: true` and `:593` reads
that. kin-vfs behavior on Windows is byte-for-byte unchanged, and Windows packaging already
tolerates absent VFS binaries (`-ErrorAction SilentlyContinue`, `release.yml:912-913`).

## The required context is renamed, and the ruleset moves with it

`Windows installer + vector-free release build` becomes
`Windows installer + vector release build`.

The name is not a label. `release-tag.yml` resolves that exact string out of `REQUIRED_CHECKS`
to decide whether Windows evidence exists and is green on a release sha. Keeping it would leave
the release mint's own authority naming a "vector-free release build" as the proof for a release
that ships vectors, in the one place an operator would look to answer what proves Windows.

Eleven surfaces carry the name and all move together:

| Surface | Location |
|---|---|
| Job display name | `.github/workflows/ci.yml:423-429` |
| Mint `REQUIRED_CHECKS` | `.github/workflows/release-tag.yml:779` |
| Mint provenance map key | `.github/workflows/release-tag.yml:893` |
| `REQUIRED_RELEASE_CHECKS` | `scripts/test-release-workflow-authority.py:257` |
| `CI_JOB_DISPLAY_NAMES` | `scripts/test-release-workflow-authority.py:453` |
| `REQUIRED_CHECK_JOB_PRODUCERS` | `scripts/test-release-workflow-authority.py:575` |
| Workflow-id provenance table | `scripts/test-release-workflow-authority.py:600` |
| Main-only admission needle | `scripts/test-release-workflow-authority.py:8674` |
| Release-gate fixture required set | `scripts/test-release-workflow-authority.py:9246` |
| Wrong-head-sha falsification | `scripts/test-release-workflow-authority.py:9847,9856` |
| Operator documentation | `docs/release-bot.md:133` |

A twelfth surface, the `main` branch ruleset, cannot be changed from a pull request and is
updated as part of landing this.

Each rename was falsified rather than assumed. Reverting the job name alone turns the suite red
on the workflow-job census; reverting `REQUIRED_CHECKS` alone turns it red on the reviewed
required set; both restore green.

## Assertions are two-directional, not merely inverted

`bench_meta.rs:308` defines `metal_active()` as `cfg!(target_os = "macos")`, and records that the
cargo feature is vestigial for "is Metal built". A vectored Windows build therefore reports
`vector,embeddings` and no `metal`. So:

- **Feature graph** (`ci.yml`): `kin-db`'s `vector` and `embeddings` must be reachable, and
  `metal` and `accelerate` must not be. Counted with `grep -Ec` rather than `grep -q`, because
  under `pipefail` a `-q` match closes the pipe and the SIGPIPE reads exactly like the feature
  being absent.
- **Shipped binary** (`ci.yml`, `release.yml`): must report `vector` and `embeddings`, with
  `metal` expected if and only if the target is `apple-darwin`.

Resolving the same tree both ways confirms the guard distinguishes: for
`x86_64-pc-windows-msvc`, `vector` and `embeddings` are present with `metal` and `accelerate` at
zero; on a macOS host `metal` and `accelerate` appear three times each. Both halves can fail.

`release.yml`'s feature check previously ran only on the leg that stripped the features, so no
leg ever asserted they were present and a silent feature regression on Linux or macOS would have
shipped green. It now runs on every native leg whose artifact the runner can execute; the
`RUNNER_ARCH:$TARGET` case at `release.yml:626-631` still exits 0 for a cross-architecture
artifact.

## Surfaces that asserted Windows is vector-free

`install-proof.yml:566` asserted the installed Windows CLI proves `vector_enabled === false`.
That leg is `continue-on-error` and fails earlier today, so the assertion gates no release, but
it is the mint's own record of what Windows ships. It is inverted, along with its
falsification needle at `test-release-workflow-authority.py:7193`, the known-good validator
fixture at `:1021`, and that fixture's negative controls at `:1319-1331`. The suite executes the
validator against the fixture, so fixture and assertion had to move together. `metal` remains a
negative control in the same direction.

The same file's installed-capability proof asserted the vector-free answer twice more, for the
CLI a user actually installed. `semantic_query_readiness` was pinned to `unsupported` on
Windows, and Windows alone was denied the pre-embed `stale` tolerance the Unix legs are granted.
Neither holds once vectors ship: the vector arm of `check_semantic_query_readiness` returns
`unsupported` only outside a Kin repository, and inside one with a reachable daemon it reports
the runtime, which is `stale` while embeddings are pending. Both now allow `healthy` or `stale`
on every OS, which is strictly more permissive than the assertions they replace. The binary
metadata assertion that the installed CLI carries vector and embedding support moves out of the
Unix-only block and runs everywhere, while `embedding_coverage` validation stays Unix-only,
because on Windows asserting it would be asserting a timing race rather than a capability, for
the FIR-1921 reason below.

Public copy in `docs/windows-wsl2.md`, `docs/security/signing-and-update-trust.md`,
`packages/kin-mcp/README.md`, `packages/kin-mcp/src/index.js` and `scripts/install.ps1` no
longer describes the archive as vector-free. The `index.js` line is the only one of these a user
is shown by a running product rather than by reading. Two `ci.yml` comments and the npm
first-run proof header now name the job that still covers the featureless msvc configuration,
which is the authority job rather than the installer, and the Windows provenance step is renamed
because it asserts enabled features now instead of disabled ones. The native-Windows admission
posture is deliberately untouched.

## The Hugging Face cache probe never matched on Windows

`retrieval_profile.rs` read `HOME` to locate the hub cache. `HOME` is not a Windows name, so
`cross_encoder_model_cached()` returned false permanently there even with the model fully
downloaded, and the cross-encoder reranker silently never turned on.

It now resolves through `resolve_home_dir()` (`commands/setup.rs:1042`, promoted to
`pub(crate)`), so the crate has one home resolver rather than two. The probe splits into
`hf_cache_base()` for the home policy and `cached_under_base()` for the cache layout so the two
fail independently, and `hf_cache_base` takes the platform and environment as arguments rather
than sitting behind `cfg(windows)`, for the reason `setup.rs:1030-1036` already records: a
cfg-gated Windows arm is compiled and tested only on the one platform this fleet has no host for.

The fix is falsified: implementing the seam with the previous `HOME`-only policy fails 6 of 12
tests on the policy rather than the plumbing; the shared resolver passes 12 of 12. Coverage spans
a Windows-shaped home, an empty cache, `USERPROFILE` outranking both OS lookups, the known-folder
fallback including an empty `USERPROFILE`, the Unix arm ignoring `USERPROFILE`, `HF_HOME`
precedence on both platforms with its empty-value case, and no resolvable home.

One divergence is deliberate and documented at the call site: `hf-hub` resolves through
`dirs::home_dir()`, which on Windows is `FOLDERID_Profile` (`dirs-6.0.0/src/win.rs:5`), while
`resolve_home_dir` prefers a non-empty `USERPROFILE`. They agree unless `USERPROFILE` has been
redirected, and preferring it is the safe side to err on, because the opposite error is a false
positive that costs a download in the middle of a query. Measured: on a native Windows host with
both names redirected to an isolated directory, `hf-hub` still wrote its cache under the
known-folder profile.

## Executed runtime proof

`.github/workflows/windows-vector-proof.yml` is new and `workflow_dispatch`-only. It builds
release with vectors on `windows-latest`, seeds a three-file Git fixture, then runs `kin init`,
`kin embed`, `kin search --semantic` and `kin locate --explain`, asserting the binary's feature
flags, full embedding coverage from the embed command's own report, the absence of a
`vector_index`/`feature_disabled` degradation, and ranked results.

It is dispatch-only because it fetches a ~522 MB model, which is a tax on every pull request and
would make the release mint hostage to huggingface.co. That property is a review responsibility
rather than an enforced invariant: the authority suite's job census pins this workflow's path,
job id and display name, but never reads `on:` triggers.

The proof gates on `kin embed`'s own report rather than on `kin status` coverage. `kin status`
deliberately never starts a daemon, so it reports whatever happens to be running when it runs:
once the daemon from the preceding command has exited it returns
`unobserved`/`no_running_daemon` no matter how much was embedded, which three runs here hit.
The field itself is not broken. On a native Windows host with a live daemon,
`kin status --json` returns `{"state":"observed","source":"live_query_graph","indexed":6954,...}`,
so this is an availability race rather than a missing capability, and it is tracked as FIR-1921.
The status read is retained as a printed diagnostic so the race stays visible instead of being
masked, and can become an assertion again once FIR-1921 lands.

Results, on the shipping architecture: the C surface including `onig_sys` compiles on
`windows-latest`; the binary reports `embeddings,vector` and no `metal`; embedding produced 12
entity vectors with zero pending and wrote `graph.kvec` in 136 seconds including the model
download; semantic search returned 5 ranked results and locate returned 3 files, with no
`feature_disabled` degradation. A native `aarch64-pc-windows-msvc` host independently embedded a
137-file ripgrep fixture (3477 entities, 6954 vectors, zero pending) in 1943 seconds at a 1748 MB
peak daemon working set, returning 10 ranked hits over real ripgrep code.

The Arm A figures come from run `30969196206`, executed at `c3191d32` rather than at this head.
Every commit since changes only workflow assertions, public copy and the authority suite, with
no Rust source, manifest or lock touched and no change to the proof job's body, so the binaries
that run produced measured the same product this head builds.

## Verification

`cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features` under CI's lint
allowlist with `RUSTFLAGS=-D warnings`, `cargo build --all-targets`, `cargo test --workspace
--no-fail-fast`, and both featureless package tests all pass. The eight CI guard scripts pass.
`scripts/test-release-workflow-authority.py` passes, with red-then-green confirmed on the rename
surfaces. `actionlint` is clean on all five changed or added workflows.

Refs FIR-1285
Refs FIR-1949
Refs FIR-1921

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -392,8 +392,9 @@ jobs:
       # contract. The installer leg asserts the same behavior against the
       # release binary, but it runs only on the landing push, so on its own it
       # can report an admission change no earlier than a required check
-      # failing on a release commit. The feature set matches the installer
-      # leg's so both legs drive the same build of the command.
+      # failing on a release commit. This leg builds without default features
+      # and the installer leg builds with them, so the shared script asserts
+      # the same refusals against both feature sets of the command.
       - name: Build the Windows binaries the admission and npm assertions drive
         shell: bash
         run: |
@@ -532,7 +533,7 @@ jobs:
             -p kin-cli -p kin-daemon
           sccache --show-stats
 
-      - name: Verify Windows build provenance and disabled features
+      - name: Verify Windows build provenance and feature set
         shell: bash
         run: |
           set -euo pipefail
@@ -1086,9 +1087,9 @@ jobs:
       # Nothing else on these legs resolves kin-cli without its default
       # features, so a break that appears only with `vector` and `embeddings`
       # off compiles clean here and reaches main behind a green required
-      # context. The Windows job covers that configuration for the msvc target
-      # and runs three filtered subsets of the library, so the cases asserting
-      # the vector-free path execute in this step and nowhere else.
+      # context. The Windows authority job covers that configuration for the
+      # msvc target through filtered subsets of the library, so the cases
+      # asserting the featureless path execute in this step and nowhere else.
       # `RUSTFLAGS: -D warnings` is set at workflow level, so the unused-import
       # shape this class first took fails the step rather than warning.
       #
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -420,7 +420,12 @@ jobs:
           node ./scripts/prove-windows-npm-first-run.mjs
 
   windows-installer:
-    name: Windows installer + vector-free release build
+    # Renamed from "Windows installer + vector-free release build" when native
+    # vector search shipped on Windows. The name is not decoration: release-tag.yml
+    # resolves this exact string out of REQUIRED_CHECKS to decide whether Windows
+    # evidence exists for a release sha, so a name still claiming "vector-free"
+    # would be the mint's own authority asserting the opposite of what it built.
+    name: Windows installer + vector release build
     needs: changes
     # Deliberately off the pull-request path. This leg cold-builds the whole
     # dependency graph in the release profile for x86_64-pc-windows-msvc and
@@ -471,24 +476,47 @@ jobs:
             ~/.cargo/git
             target
             ${{ runner.temp }}/sccache
-          key: windows-vectorless-cargo-${{ hashFiles('Cargo.lock') }}
+          # Cache key renamed with the feature set. The vectorless key indexed a
+          # dependency graph with no ndarray, tokenizers, hnsw or onig in it, so
+          # restoring from it would warm almost nothing and would report a hit
+          # while doing so.
+          key: windows-vector-cargo-${{ hashFiles('Cargo.lock') }}
           restore-keys: |
-            windows-vectorless-cargo-
+            windows-vector-cargo-
 
-      - name: Assert vector-free feature graph
+      # Windows ships native vector search, so this asserts the feature graph in
+      # both directions rather than only forbidding features. `vector` and
+      # `embeddings` must be reachable, or the leg would silently certify a
+      # degraded build; `metal` and `accelerate` must not be, because those are
+      # macOS-only backends whose activation lives in a
+      # `cfg(target_os = "macos")` dependency block and must never unify in here.
+      #
+      # `grep -c` rather than `grep -q`: under `pipefail` a `-q` match closes the
+      # pipe early and the SIGPIPE fails the step, which reads exactly like the
+      # feature being absent. Reading a file cannot invert the guard that way.
+      - name: Assert the Windows feature graph
         shell: bash
         run: |
           set -euo pipefail
           cargo tree --locked --target x86_64-pc-windows-msvc \
-            -p kin-daemon --no-default-features -e features -i kin-db \
+            -p kin-daemon -e features -i kin-db \
             | tee "$RUNNER_TEMP/kin-db-feature-tree.txt"
-          if grep -Eq 'kin-db feature "(vector|embeddings|metal|accelerate)"' \
-            "$RUNNER_TEMP/kin-db-feature-tree.txt"; then
-            echo "::error::Windows vector-free build still enables a heavy kin-db feature"
-            exit 1
-          fi
+          for required in vector embeddings; do
+            if [ "$(grep -Ec "kin-db feature \"${required}\"" \
+              "$RUNNER_TEMP/kin-db-feature-tree.txt")" -eq 0 ]; then
+              echo "::error::Windows release build is missing kin-db feature ${required}"
+              exit 1
+            fi
+          done
+          for forbidden in metal accelerate; do
+            if [ "$(grep -Ec "kin-db feature \"${forbidden}\"" \
+              "$RUNNER_TEMP/kin-db-feature-tree.txt")" -ne 0 ]; then
+              echo "::error::Windows build pulled the macOS-only kin-db feature ${forbidden}"
+              exit 1
+            fi
+          done
 
-      - name: Build Windows release binaries without vector features
+      - name: Build Windows release binaries with vector features
         shell: bash
         env:
           # Compile cache keyed per crate, so a Cargo.lock change warm-starts
@@ -501,7 +529,7 @@ jobs:
           CARGO_INCREMENTAL: "0"
         run: |
           cargo build --locked --release --target x86_64-pc-windows-msvc \
-            -p kin-cli -p kin-daemon --no-default-features
+            -p kin-cli -p kin-daemon
           sccache --show-stats
 
       - name: Verify Windows build provenance and disabled features
@@ -544,10 +572,18 @@ jobs:
               `does not match tracked Git bytes ${expectedLock}`
             );
           }
-          for (const feature of ["vector", "embeddings", "metal"]) {
-            if (meta.feature_flags.includes(feature)) {
-              throw new Error(`Windows vector-free build unexpectedly enabled ${feature}`);
+          // The shipped binary must report the features it was built with, not
+          // merely resolve them in the dependency graph. `metal` tracks the
+          // compiled backend rather than the marker cargo feature
+          // (bench_meta.rs `metal_active()` is `cfg!(target_os = "macos")`), so
+          // a correct Windows build reports vector and embeddings and no metal.
+          for (const feature of ["vector", "embeddings"]) {
+            if (!meta.feature_flags.includes(feature)) {
+              throw new Error(`Windows release build did not enable ${feature}`);
             }
+          }
+          if (meta.feature_flags.includes("metal")) {
+            throw new Error("Windows release build reported the macOS-only metal backend");
           }
           NODE
 
--- a/.github/workflows/install-proof.yml
+++ b/.github/workflows/install-proof.yml
@@ -1426,11 +1426,11 @@ jobs:
               throw new Error(`${label}: unknown embedding coverage state ${coverage.state ?? "missing"}`);
             }
           };
+          if (cliMeta.embeddings?.vector_enabled !== true || cliMeta.embeddings?.embeddings_enabled !== true) {
+            throw new Error(`${runnerOs} bench metadata does not prove vector and embedding support`);
+          }
           if (!isWindows) {
             validateEmbeddingCoverage(status.embedding_coverage, "kin-status.json");
-            if (cliMeta.embeddings?.vector_enabled !== true || cliMeta.embeddings?.embeddings_enabled !== true) {
-              throw new Error(`${runnerOs} bench metadata does not prove vector and embedding support`);
-            }
           }
           const reports = ["kin-health.json", "kin-doctor.json"].map((path) => [
             path,
@@ -1519,7 +1519,7 @@ jobs:
                   check.status === "healthy" ||
                   check.status === "unsupported"
               );
-            if (report.healthy !== true && (isWindows || !onlyPendingEmbeddingsBlocks)) {
+            if (report.healthy !== true && !onlyPendingEmbeddingsBlocks) {
               throw new Error(
                 `${path}: overall healthy=${report.healthy}; non-healthy checks: ${nonHealthyChecks(report)}`
               );
@@ -1531,9 +1531,7 @@ jobs:
               }
             }
             const semanticReadiness = checks.get("semantic_query_readiness")?.status;
-            const allowedSemanticReadiness = isWindows
-              ? ["unsupported"]
-              : ["healthy", "stale"];
+            const allowedSemanticReadiness = ["healthy", "stale"];
             if (!allowedSemanticReadiness.includes(semanticReadiness)) {
               throw new Error(
                 `${path}: semantic_query_readiness is ${semanticReadiness ?? "missing"}, expected ${allowedSemanticReadiness.join(" or ")}`
--- a/.github/workflows/install-proof.yml
+++ b/.github/workflows/install-proof.yml
@@ -558,12 +558,15 @@ jobs:
           if (meta.dependency_provenance !== expectedLock) {
             throw new Error(`installed Windows Cargo.lock provenance ${meta.dependency_provenance ?? "missing"} does not match release source ${expectedLock}`);
           }
+          // What a user actually installed, not what CI built. `metal_enabled`
+          // tracks the compiled backend rather than the marker cargo feature,
+          // so it stays false on Windows even with default features on.
           if (
-            meta.embeddings?.vector_enabled !== false ||
-            meta.embeddings?.embeddings_enabled !== false ||
+            meta.embeddings?.vector_enabled !== true ||
+            meta.embeddings?.embeddings_enabled !== true ||
             meta.embeddings?.metal_enabled !== false
           ) {
-            throw new Error("installed Windows CLI does not prove the vector-free feature contract");
+            throw new Error("installed Windows CLI does not prove the vector feature contract");
           }
 
           const authority = JSON.parse(fs.readFileSync("kin-windows-registry-authority.json", "utf8"));
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -793,7 +793,7 @@ jobs:
             DCO Sign-off
             cargo-deny
             gitleaks (full history)
-            Windows installer + vector-free release build
+            Windows installer + vector release build
         run: |
           set -euo pipefail
           gh api "repos/$REPO/actions/runs/$CURRENT_RUN_ID" \
@@ -907,7 +907,7 @@ jobs:
                   "push",
                   release_branch,
               ),
-              "Windows installer + vector-free release build": (
+              "Windows installer + vector release build": (
                   245803170,
                   ".github/workflows/ci.yml",
                   "push",
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -425,7 +425,12 @@ jobs:
             target: x86_64-pc-windows-msvc
             artifact: kin-windows-x86_64
             shim_name: kin_vfs_shim.dll
-            skip_vector: true
+            # kin-vfs only. This row used to carry `skip_vector: true`, which
+            # gated BOTH the vector feature set and the kin-vfs build, so
+            # shipping vectors on Windows by deleting that one key would also
+            # have started building an LD_PRELOAD libc interceptor for msvc.
+            # The two are unrelated and are now named separately.
+            skip_vfs: true
             cross: false
 
     steps:
@@ -554,17 +559,16 @@ jobs:
         if: ${{ !matrix.cross }}
         shell: bash
         run: |
-          FLAGS=""
-          if [ "$SKIP_VECTOR" = "true" ]; then
-            FLAGS="--no-default-features"
-          fi
           # The daemon is the runtime: `kin init` works without it but
           # `kin status`/`kin search`/the MCP server all require kin-daemon on
           # PATH, so a clean public install must ship it alongside `kin`.
-          cargo build --locked --release --target "$TARGET" -p kin-cli -p kin-daemon $FLAGS
+          #
+          # Every platform now builds the same default feature set. Windows was
+          # the only leg that stripped it, and dropping that split is what puts
+          # native vector search in the shipped Windows archive.
+          cargo build --locked --release --target "$TARGET" -p kin-cli -p kin-daemon
         env:
           TARGET: ${{ matrix.target }}
-          SKIP_VECTOR: ${{ matrix.skip_vector }}
 
       - name: Build kin-cli + kin-daemon (cross)
         if: ${{ matrix.cross }}
@@ -590,7 +594,7 @@ jobs:
           TARGET: ${{ matrix.target }}
 
       - name: Build kin-vfs (native)
-        if: ${{ !matrix.cross && !matrix.skip_vector }}
+        if: ${{ !matrix.cross && !matrix.skip_vfs }}
         working-directory: kin-vfs
         shell: bash
         run: |
@@ -617,7 +621,6 @@ jobs:
         shell: bash
         env:
           TARGET: ${{ matrix.target }}
-          SKIP_VECTOR: ${{ matrix.skip_vector }}
         run: |
           set -euo pipefail
           case "$RUNNER_ARCH:$TARGET" in
@@ -639,7 +642,7 @@ jobs:
           "$kin_bin" bench-meta --json > "$provenance_dir/kin-build-meta.json"
           expected_sha="$(git rev-parse HEAD)"
 
-          EXPECTED_SHA="$expected_sha" SKIP_VECTOR="$SKIP_VECTOR" PROVENANCE_DIR="$provenance_dir" node <<'NODE'
+          EXPECTED_SHA="$expected_sha" PROVENANCE_DIR="$provenance_dir" node <<'NODE'
           const crypto = require("crypto");
           const fs = require("fs");
           const expected = process.env.EXPECTED_SHA;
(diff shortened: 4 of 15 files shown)


troyjr4103:
## Arm A run 1: the runtime claim is proven; my harness had an ordering defect

Run [30965647931](https://github.com/firelock-ai/kin/actions/runs/30965647931) on `windows-latest` x86_64. Every step through the queries succeeded. The final assertion step failed, and it was my workflow, not the product:

```
Error: status could not observe embedding coverage: {"state":"unobserved","reason":"no_running_daemon"}
```

`kin status` deliberately does not auto-start the daemon (install-proof records this at `install-proof.yml:865-868`), and the daemon `kin embed` used had already exited. I had ordered the status read **before** the query step that starts one.

### What the artifacts actually show

| Evidence | Value |
|---|---|
| `kin embed` | `embedded_entities: 12`, `pending_entities: 0`, `time_limited: false` |
| Vector index written | `D:\a\_temp\vector-fixture\.kin\kindb\graph.kvec` |
| Wall clock **including the 522 MB model download** | **136 seconds** |
| `kin search --semantic --json` | ranked array with cosine scores `0.2460`, `0.2439`, `0.2413`, … |
| `kin locate` degradations | **no `vector_index` / `feature_disabled`**; only `capability_tier: minimal` |

The query was `credentials`, which appears in **none** of the three fixture files, so those ranked hits are the embedding arm answering rather than a substring match. The single degradation present is a hardware-tier note (runner has <8 cores) that disables the reranker; it is unrelated to the vector feature.

**Two of the scout's largest unknowns are now closed:** `onig_sys` and the rest of the C surface compile on `windows-latest`, and embedding inference has executed on native Windows x86_64 for the first time in this fleet.

### Fixes in `881b5b91`

- `kin status` moved after the queries, with the reason recorded in place so the next reader does not re-derive it.
- The assertion now also checks `embed.embedded_entities > 0` and `embed.vector_index_path` **at the source**. "Zero pending" is satisfied vacuously by a run that embedded nothing, so the count that proves inference ran needed checking directly rather than only through a status read.
- The `hf-hub` cache divergence in `retrieval_profile.rs` is now recorded as **measured**: on the native Windows guest, with `HOME` and `USERPROFILE` both redirected to an isolated directory, `hf-hub` still wrote its 522 MB cache under the known-folder profile and left the isolated home empty. That confirms from behavior what the comment previously argued from source, and it confirms `USERPROFILE`-first is the safe side to err on, since the opposite error is a false positive that costs a mid-query download.

Run 2 is [30967429595](https://github.com/firelock-ai/kin/actions/runs/30967429595).

troyjr4103:
## Both arms: native Windows semantic vector search is proven, on two architectures

### Arm B — native `aarch64-pc-windows-msvc` guest, COMPLETE

Single-commit import of the pinned ripgrep source (`e89fff89`), 137 files, isolated `HOME`.

| Evidence | Value |
|---|---|
| Both binaries | `kin` and `kin-daemon` at `bc093a3d` |
| `bench-meta` `feature_flags` | `embeddings,vector` — and **no** `metal` |
| `kin init` | exit 0 in **15 s**; 3477 entities, 12464 relations, repository-v6 graph-owned |
| `kin embed` | `total_entities: 3477`, **`embedded_entities: 6954`**, `pending_entities: 0`, `time_limited: false` |
| Vector index | `...\rg-src\.kin\kindb\graph.kvec` |
| Embed wall clock | **1943 s (32.4 min)** |
| Peak daemon working set | **1748.7 MB**, under the documented 2 GiB floor |
| `kin search --semantic` | **10 ranked hits on real ripgrep code**: `RegexMatcherBuilder::multi_line` @ 0.429, `CommandReader::new` @ 0.362, `CommandReaderBuilder::build` @ 0.352 |
| `kin locate` | 5 files, **no `vector_index/feature_disabled`** |

**Arch caveat: this is ARM64 and the release target is x86_64.** The 32-minute embed is a QEMU-guest number and describes neither the shipping arch nor real hardware. Arm A is the shipping-arch proof.

### `kin status` is the wrong proof surface on Windows, and that is a product finding

Runs 1 and 2 both failed the same assertion with `{"state":"unobserved","reason":"no_running_daemon"}`. Run 2 had the status read moved **after** a query that starts a daemon, and failed identically — so ordering was not the cause.

On Windows the daemon does not stay resident between CLI invocations, and `kin status` deliberately never starts one, so **any** read of `embedding_coverage` there races the daemon's lifetime. Two independent hosts reproduce it: the `windows-latest` runner and the native aarch64 guest.

`c3191d32` keeps that read as a **diagnostic** and gates on the embed command's own report instead — the direct output of the process that did the work, rather than a second process's view of a daemon that may have exited.

**Worth a ticket beyond this PR:** the Unix install-proof legs assert on exactly this field and pass, because there the daemon is still up. Any future Windows proof reaching for `kin status` coverage hits the same wall.

### An aside worth recording

The full-history ripgrep import was tried first and abandoned: past 12 minutes at ~4.9 GB RSS it had still not produced `.kin`. The single-commit import of the same pinned tree did the whole job in 15 seconds.

Artifacts: `.kin-coord/reports/night-20260805-nwvec-impl-logs/{armA-artifacts,armB-artifacts}/`. Full write-up: `.kin-coord/reports/night-20260805-nwvec-impl.md`.

troyjr4103:
## Arm A run 3: GREEN, all 18 steps

Run [30969196206](https://github.com/firelock-ai/kin/actions/runs/30969196206) on `windows-latest` x86_64:

```
feature_flags: embeddings,vector
embedded_entities=12 index=D:\a\_temp\vector-fixture\.kin\kindb\graph.kvec
semantic search returned 5 ranked results; locate returned 3 files
status embedding_coverage (diagnostic): {"state":"unobserved","reason":"no_running_daemon"}
```

That last line is the whole argument for the change in `c3191d32`: the diagnostic still reports `no_running_daemon` on a **fully green** run, so gating on it would have failed a run in which everything worked. Every substantive assertion passes.

`e5a93334` removes the temporary lane-branch `push:` trigger now that the proof is green, so the workflow ships `workflow_dispatch`-only as designed, with a note recording why that trigger was needed (a `workflow_dispatch` workflow is not dispatchable until it exists on the default branch, so it cannot be proven on the branch that introduces it).

Incidentally, the caches saved for the first time on this run — the earlier `Post Cache` steps show `skipped`, because `actions/cache` does not save on a failed job. A future dispatch will not pay the ~45 minute cold build.

### Both arms, final

| | Arm A (`x86_64-pc-windows-msvc`, windows-latest) | Arm B (`aarch64-pc-windows-msvc`, guest) |
|---|---|---|
| Build with vectors | ✅ (`onig_sys` + C surface compile) | ✅ |
| `feature_flags` | `embeddings,vector`, no `metal` | `embeddings,vector`, no `metal` |
| `kin init` | ✅ | ✅ 15 s, 3477 entities / 12464 relations |
| `kin embed` | 12 entities, 0 pending, **136 s incl. 522 MB download** | 6954 embedded, 0 pending, **1943 s**, peak 1748.7 MB |
| Semantic search | 5 ranked results | 10 ranked hits on real ripgrep code |
| `feature_disabled` | absent | absent |

**Arm A is the shipping-arch proof. Arm B's timings are QEMU-guest numbers on ARM64 and describe neither the shipping arch nor real hardware.**

troyjr4103:
## Correction: the daemon finding is not a new ticket — it is FIR-1921

I twice wrote that the `kin status` / `no_running_daemon` behaviour "needs a ticket". It already has one, and I should have searched before saying that.

**[FIR-1921](https://linear.app/firelock-ai/issue/FIR-1921/windows-install-proof-leg-daemon-endpoint-file-never-published-leg) — "Windows install-proof leg: daemon endpoint file never published; leg demoted to experimental until fixed" — In Review.**

That root cause explains what I saw exactly. Both FIR-1921's daemon-provenance step and my coverage read resolve the daemon through the endpoint file, so both report "no daemon" while a daemon is up and answering queries. Nothing new to file.

I added my reproduction as a comment on FIR-1921 rather than opening a duplicate. Two things it contributes:

1. **Independent repro from a different direction** — a vector proof rather than the install proof, on two hosts: `windows-latest` x86_64 and a native `aarch64-pc-windows-msvc` guest entirely outside CI.
2. **Wider blast radius than the ticket currently records.** The consequence is not only a failed provenance assertion. On Windows, `kin status --json` reports `embedding_coverage` as `unobserved/no_running_daemon` **regardless of what was actually embedded** — a silent wrong answer on a healthy system, wearing the costume of a legitimate "not ready" state. Run [30969196206](https://github.com/firelock-ai/kin/actions/runs/30969196206) passes all 18 steps and still prints that line.

Nothing in this PR changes as a result. The diagnostic is printed on every run including green ones precisely so the defect stays visible instead of being masked by my gating change, and it can be turned back into an assertion the moment FIR-1921 lands.

troyjr4103:
## Correction: I overclaimed what the job census protects

`1bf78e83`. I wrote — in the workflow comment, the census entry, and the PR body — that registering `windows-vector-proof.yml` in the authority suite's job census keeps it dispatch-only and stops it becoming a required-context producer.

**That is false, and I falsified it directly.** The census (`assert_workflow_job_census`) reads each workflow's path, job ids and display names. It never reads `on:`. I injected `pull_request:` and `merge_group:` into this workflow, re-ran the suite, and it stayed **green**, then restored the file and confirmed green again.

So the accurate statement is narrower:

- Registering the workflow pins its **name**. That is real value: it is what stops a second required-context producer appearing unreviewed.
- It does **not** pin the triggers. Keeping this dispatch-only is a **review responsibility, not a guarded invariant**.

Both the workflow comment and the census entry now say that explicitly and name what a trigger would cost: a check run on every pull request carrying a ~522 MB model fetch and a cold ~45 minute Windows release build. The PR body is corrected too.

This corroborates a blind-gate finding another lane made independently against the same suite. Nothing about the vector change itself moves — this is purely about not crediting a gate with a guarantee it does not provide.

Reviewers: the general form is worth applying to the rest of this PR. Where I say a suite "pins" something, the claim earns its place only if a violation of that exact property has been injected and seen to go red. I did that for the two rename surfaces and for the feature-graph guard (both directions); I had not done it here, which is how the overclaim survived.

troyjr4103:
## 🚨 CAPTAIN: ruleset context rename required before this merges (moved here from the PR body)

Moved out of the PR body deliberately. On a hosted-queue repo the squash message is minted from the PR title and body with nobody at the merge button — verified against `c38ce1a0`, whose commit message is its PR body verbatim. Operational instructions and session narrative would become permanent public history, so the body is now the durable technical commit message and everything below lives here instead.

### The blocker

- **Ruleset:** `19746451` — "Require status checks on main"
- **OLD context:** `Windows installer + vector-free release build`
- **NEW context:** `Windows installer + vector release build`
- The merge-queue ruleset `20050419` carries **no** required status checks, so `19746451` is the only one to edit. Verified by direct read of both.

Measured, not predicted: `gh pr checks 616 | grep -Fc "vector-free"` returns **0**. The old context is absent from the rollup and no job in this PR produces it, so it will never report. `gh pr view` shows `MERGEABLE` / `BLOCKED`.

### Sequencing, and the part that bites

**There is no zero-downtime rename.** While the ruleset names the new context, every OTHER open PR still emits the old name from its own `ci.yml`, so the new one never reports and they block. Flip it first and everyone blocks; flip it last and this PR blocks. No ordering avoids both.

**Do it with the queue drained, and expect every open PR to need a rebase onto this squash.**

**Also gated:** do not land before the v0.5.0 tag exists. Tonight's release resolves workflows at its tag ref, and a pre-tag landing on `main` changes what the mint sees.

### Corrections I made during this lane (kept out of history, kept available here)

1. **kin-db needs no change.** I claimed it had no Windows CI and opened kin-db#162. False — its `Check & Test` matrix already includes `windows-latest` (`ci.yml:89`, from kin-db#158) and runs the full suite there under `vector` + `embeddings`. I had verified against a local checkout fifteen commits stale. PR closed, branch and worktree removed.
2. **The daemon finding is FIR-1921, not a new ticket.** "Windows install-proof leg: daemon endpoint file never published", already In Review. My repro is commented there, including the wider blast radius: `kin status --json` on Windows reports `embedding_coverage` as `unobserved/no_running_daemon` regardless of what was embedded.
3. **The job census does not protect dispatch-only.** I claimed it did; it reads job names and never `on:`. Injected `pull_request:` + `merge_group:`, suite stayed green, restored. Fixed in `1bf78e83`; both comments now say it is a review responsibility.

Run evidence: Arm A [30969196206](https://github.com/firelock-ai/kin/actions/runs/30969196206) (green, all 18 steps). Full write-up and both artifact sets: `.kin-coord/reports/night-20260805-nwvec-impl.md`.

troyjr4103:
## Falsified the bench-meta assertions too: 14/14 defect shapes rejected

I had falsified the rename surfaces and the feature-graph guard, but **not** the bench-meta assertions I added to `ci.yml` and `release.yml`. By the standard I stated earlier in this PR, those claims had not earned their place. Closing that gap.

Extracted both assertions verbatim and ran them against synthetic `bench-meta` payloads. **14/14 behaved as specified** — every correct configuration passes, every defect throws.

Two results are worth calling out:

- **The pre-change vector-free shape now throws.** `{"feature_flags": []}` on Windows is rejected, so reverting this flip is caught rather than silently restored. That makes this a regression guard, not just an inverted assertion.
- **"macOS lost its metal backend" throws.** Nothing covered that case before. The old check ran *only* on the leg that stripped features, so a macOS build quietly losing Metal would have shipped green. This is coverage the change **adds**, not coverage it merely flips.

Full matrix: vector stripped / embeddings stripped / fully vector-free / metal leaked, across `x86_64-pc-windows-msvc`, `x86_64-unknown-linux-musl` and `aarch64-apple-darwin`, plus the four correct configurations.

Harness and output are kept as evidence at `.kin-coord/reports/night-20260805-nwvec-impl-logs/falsify-feature-assertions.mjs` and `feature-assertions-falsification.log`. Deliberately **not** committed to the repo: the assertions are ten inline lines of YAML, and adding a permanent test harness for them is a maintenance surface this PR does not need. If a reviewer would rather have it committed, say so and I will add it.

troyjr4103:
## Review of `ce19fa0a`: three surfaces I missed, and one change that I believe fails at release time

Someone pushed `ce19fa0a` onto this branch while I was idle. I updated my worktree and reviewed it rather than assume. **Three of its four parts are catches I missed and am glad to have**, and one looks like it will redden the release.

### Good catches I missed

- `packages/kin-mcp/src/index.js:252` was telling **users**, in a guided-failure message, "Native Windows is vector-free; use WSL2 when you need vectors or projection." I swept docs and workflows and never swept runtime user-facing strings. That is the most user-visible vector-free claim in the tree and I walked past it.
- `scripts/install.ps1:393` and `scripts/prove-windows-npm-first-run.mjs:7` comments, likewise.
- The `~45 minute` build figure in my workflow comment corrected to `~30 minute job, of which the release build is ~23 minutes`. **Mine was wrong.** I estimated 45 from watching the failing cold runs; the green run 30969196206 created at 02:23:04 logged its feature check at 02:46:40, so the build was ~23 minutes and the job ~30. The corrected figure is the measured one.

### The concern: `install-proof.yml:1535`

The change makes `allowedSemanticReadiness` unconditional:

```diff
-            const allowedSemanticReadiness = isWindows
-              ? ["unsupported"]
-              : ["healthy", "stale"];
+            const allowedSemanticReadiness = ["healthy", "stale"];
```

That step, "Validate installed capability proof" (line 1333), runs on Windows. But on this exact commit, `check_semantic_query_readiness()` in the vector-enabled build returns:

- `Unsupported` when `KinLayout::discover` finds no repository (`health.rs:1474-1481`)
- **`Missing`** when a repository exists but `resolve_daemon_url_if_running_async` yields nothing (`health.rs:1484-1492`)

And FIR-1921 is precisely that the Windows daemon endpoint file is never published, so that resolution yields nothing. **Neither `unsupported` nor `missing` is in `["healthy","stale"]`, so I believe this throws on the Windows leg.**

It is the same shape as the `install-proof.yml:566` defect fixed earlier in this PR, with one nastier property: install-proof runs on `release:published` and on schedule, **not on pull requests**, so PR CI will stay green and this surfaces first on a release commit.

### What would make it correct

This change is right **after FIR-1921 lands** — once the endpoint file is published, Windows resolves a daemon and reports `healthy`/`stale` like the Unix legs. So I read it as a correct change with an unstated ordering dependency, not a wrong one. Options, in my order of preference:

1. Land it behind FIR-1921, and note the dependency here so the sequencing is deliberate.
2. Keep the Windows arm permissive for now, e.g. `isWindows ? ["unsupported", "missing", "healthy", "stale"] : ["healthy", "stale"]`, which tolerates the FIR-1921 state without asserting the vector-free posture I am removing.
3. If the author has evidence the Windows leg already resolves a daemon at that point, that beats my reading and I withdraw it.

**Stated plainly: I have not executed the Windows install proof.** This is read from `health.rs` on `ce19fa0a` plus FIR-1921's own description, not from a run. Option 3 is a live possibility and I would rather be corrected than have this land unexamined.

I have **not** edited the branch. It is someone else's in-flight work and this is a judgement call for its author.

troyjr4103:
## Withdrawing my concern about `install-proof.yml:1535`: I was wrong

I raised this an hour ago and then kept digging. The digging refuted me, so retracting promptly rather than leaving a wrong objection standing against someone else's work.

**What I got wrong.** I reasoned "FIR-1921 says the daemon endpoint file is never published on Windows, therefore the daemon is unresolvable, therefore `semantic_query_readiness` is `missing`". That chain breaks at the first link, because it conflates two different files:

- FIR-1921 is specifically about **`.kin/daemon.port`**, a per-repo file that a **bash step** in `install-proof.yml` (~line 856) reads directly. Its failure line is `.kin/daemon.port: No such file or directory`.
- `check_semantic_query_readiness()` does not read that file. It calls `resolve_daemon_url_if_running_async` → `supervisor_route_for_repo_if_running_async` → `live_supervisor_endpoint()`, which reads a **supervisor endpoint snapshot** out of `supervisor_dir()` (`daemon_client.rs:3398-3410`, `:6068-6089`). Different artifact, different publication path, different consumer.

**And the ticket contains direct evidence against me.** FIR-1921's own account of run 30947137583 records that "the store served `kin search hello --json` and `kin locate hello --json --explain` with full JSON output, meaning **the daemon auto-started and answered queries**." Those queries route through the same supervisor lookup the health check uses. If that resolution worked well enough to serve queries, it resolves for the health check too, and the status is `healthy` or `stale` — which is exactly what `ce19fa0a` now asserts.

**So `ce19fa0a` looks correct and my objection was noise.** There is no ordering dependency on FIR-1921 that I can support, and my suggested "permissive Windows arm" would have weakened a correct assertion for no reason. Ignore options 1 and 2 in my earlier comment; option 3, that the author had the better reading, is what happened.

The one caveat I can still honestly hold is much smaller and not an objection: the health check's route lookup carries a 2 second HTTP timeout (`daemon_client.rs:6072-6074`), so an unusually slow Windows runner could yield `missing` as a flake rather than as a posture. That is a robustness question about the health check, not a reason to hold this change.

What I should have done before posting: read FIR-1921's body rather than reasoning from its title. The title says "daemon endpoint file never published"; the body says which file, and which consumer, and that the daemon served. I had the ticket open and used the summary.

troyjr4103:
## Correction: my "Windows cannot observe embedding coverage" claim was false

`a51fa8bf`. Second retraction from me on this PR, and this one corrects a claim that had made it into a shipped code comment, the PR body, and a comment on FIR-1921.

**What I claimed:** on Windows the daemon does not stay resident between CLI invocations, so `kin status --json` reports `embedding_coverage` as `unobserved/no_running_daemon` regardless of what was embedded, and the surface itself is wrong.

**The disproof.** On the native aarch64 Windows guest, with `tasklist` confirming `kin-daemon.exe` alive on both sides of the call:

```json
{"state":"observed","source":"live_query_graph","indexed":6954,"pending":0,"total":6954}
```

`state: observed`, all 6954 vectors, on Windows. The field works.

**What was really happening:** `kin status` deliberately never starts a daemon, so it reports whatever is running at that instant, and the daemon from the preceding command had exited. It was reporting a real state accurately. A timing race in my harness, not a platform defect.

**Two compounding errors, both mine:**

1. I reasoned from FIR-1921's **title** rather than its body. The body names the file (`.kin/daemon.port`), names the consumer (a bash step at install-proof.yml ~856), and states the daemon "auto-started and answered queries". The health check never reads that file — it resolves through `live_supervisor_endpoint()`, a different snapshot with a different publisher.
2. **Reproducing the symptom on two hosts felt like corroboration. It was not.** Both reproductions ran through my harness with the same ordering, so I replicated my own mistake and counted it as independent evidence. Varying the host while holding the harness fixed proves nothing about the harness.

**What does not change:** gating on `kin embed`'s own report is still correct and stays. It depends on nothing being alive afterwards, which is the property worth having on every platform. `a51fa8bf` changes only the stated reason, which was wrong, and records the measured counter-example inline so the next reader is not re-misled.

Also retracted on FIR-1921 so that ticket does not carry a blast-radius claim it never had.

PR #626, opened 2026-08-05
Report which assertion failed in the release follow-up proofs

troyjr4103:
`Publish Release Follow-Ups` run 31004252835 failed with no diagnosable cause:
twenty-three seconds of no output, then

```
##[error]Process completed with exit code 1
```

No error line, no step summary, nothing naming what diverged.

## Cause

Every guard in this workflow was a bare test under `set -e`:

```bash
[ "$(jq -r .conclusion <<< "$state")" = success ]
```

A bare `[ ]` that fails exits 1 and prints nothing, so the run reports only an
exit code. In this case the assertion was **correct**: the downstream Homebrew
run `31004285786` attempt 1 had concluded `failure` at 12:07:57Z, because its
`KIN_RELEASE_APP_PRIVATE_KEY` was missing. The proof step did its job and caught
a real broken release channel, then reported it in the least useful form
available.

That is the whole defect. The signal was right; only the reporting was missing.

## Change

Assert through small helpers that name the field and the value actually
observed. The same failure now reads:

```
::error::Homebrew run 31004285786 conclusion: expected 'success', got 'failure'
```

**No condition is relaxed.** Every comparison, regex, and either-or is preserved
exactly; only the reporting changes. Verified case by case below.

Applied across all five steps that used bare guards, not only the two that
failed today — leaving the rest silent would relocate the trap rather than
remove it:

- `Require current callback authority…` — ref, head, ancestry
- `Revalidate completed public release provenance` — 3 regex + 9 field + 4 misc
- `Validate and hash exact installer sources` — checkout sha, both digests
- `Prove Homebrew workflow and public formula` — 5 downstream fields
- `Prove installer activation and public byte parity` — 5 downstream fields

## Verification

Behavioral test of the extracted helpers, both directions:

| case | exit | output |
|---|---|---|
| matching field | 0 | passes through |
| mismatched field | 1 | `::error::Homebrew run 123 conclusion: expected 'success', got 'failure'` |
| matching regex | 0 | passes through |
| non-matching regex | 1 | `::error::release tag: '0.5.2' does not match ^v[0-9]+\.[0-9]+\.[0-9]+$` |
| both digests valid | 0 | passes through |
| one digest invalid | 1 | `::error::install.ps1 digest 'NOTAHASH' is not 64 lowercase hex characters` |

Exit 1 on failure is preserved in every case, so nothing that previously failed
now passes.

`actionlint` passes. A scan for remaining bare `[ ]` / `[[ ]]` assertions in the
file returns none.

Note this does not change today's outcome: the underlying cause was the Homebrew
private key, which is already fixed. This ensures the next such failure names
itself.


troyjr4103:
Pushed a follow-up commit for the `npm launcher tests` failure, which was a real consequence of this change rather than an unrelated flake.

`scripts/test-release-workflow-authority.py` pins the literal text of the provenance assertions the completed-release callback must contain. Four of the 56 pinned policies were assertions this PR converted, so the policy test correctly reported them missing:

```
AssertionError: completed-release follow-up callback is missing required policy:
[ "$(jq -r .status <<< "$run")" = completed ]
```

The four pins are updated to the form that carries the message. The other 52 are untouched, including `[ "$ancestry" = ahead ] || [ "$ancestry" = identical ]`, which survives unchanged because the message was appended rather than replacing it.

The guarantee is unchanged, and I verified the test still has teeth rather than assuming it:

- with the workflow as proposed: policy test exits 0
- with `assert_field "source run conclusion" …` deleted from the workflow: exits 1 with `missing required policy: assert_field "source run conclusion" "$(jq -r .conclusion <<< "$run")" success`
- restored: exits 0 again

So the same four conditions remain required, and dropping any one of them still fails the gate by name.

PR #641, opened 2026-08-06
Make the daemon pid-file assertion wait, and prove the leak is not a race

troyjr4103:
Refs FIR-1991

## What this PR now establishes, which is not what it set out to do

`daemon_status_and_stop_lifecycle` asserted `daemon.pid` was gone immediately
after observing the worker process exit, with no wait between the two events.
The obvious reading was a race: the daemon unlinks its own pid file, and on
Windows an open handle can keep an unlinked name visible briefly, so a parent
that has only seen the process leave might still see the path.

**That reading is wrong, and this PR is the evidence.** The change adds
`wait_until_gone`, a bounded wait mirroring the `wait_until_dead` helper above
it, and gives the pid file five seconds to disappear before asserting. On this
PR the Windows authority tests still failed, at the same assertion:

```
thread panicked at crates\kin-cli\tests\daemon_stop.rs:162:5:
  daemon.pid must be cleared after a confirmed stop
```

The wait ran, exhausted its full deadline, and the file was still there. A
sub-second visibility window cannot explain a file that survives five seconds.
**`kin daemon stop` reports a confirmed stop and leaves `daemon.pid` behind.**

This is corroborated on another platform. On macOS, `kin daemon stop` was
observed returning zero while both the supervisor and the per-repo daemon
remained alive and had to be reaped by pid. The same shape appears in both
places: the stop path reports success before its cleanup is complete.

## What to do with this change

It is worth keeping on its own merits even though it does not make the test
green:

- It removes the timing hypothesis as a confound. Any future failure of this
  assertion is now known not to be a short visibility window, because the test
  waits.
- It fails loud rather than vacuously. Driving the wait against a path that
  never disappears exhausts the deadline and fires the assertion, rather than
  passing; the assertion keeps its full strength.
- The failure it reports is a real product defect, so the assertion is correct
  to fail. The test is not what needs fixing first.

The nearby `supervisor.pid` assertion is deliberately unchanged. It asserts a
file was never created at all, so waiting there would weaken it.

This does not close FIR-1991. The remaining work is in the daemon stop path,
not in the test.

### DIFF
--- a/crates/kin-cli/tests/daemon_stop.rs
+++ b/crates/kin-cli/tests/daemon_stop.rs
@@ -67,6 +67,24 @@ fn wait_until_dead(pid: u32, timeout: Duration) {
     }
 }
 
+/// Wait for a path to disappear, the file-side analogue of `wait_until_dead`.
+///
+/// Process death and pid-file removal are two different events. The exiting
+/// daemon unlinks its own pid file, so a parent that has only observed the
+/// process leave can still see the path for a short window, and on Windows an
+/// open handle keeps an unlinked name visible for longer than it does on Unix.
+/// Asserting the file is gone the instant the process is gone therefore fails
+/// intermittently while describing the failure as a product defect.
+///
+/// This still fails when the file genuinely never goes away: the caller asserts
+/// after the wait, so a real leak times out and reports exactly as before.
+fn wait_until_gone(path: &Path, timeout: Duration) {
+    let deadline = Instant::now() + timeout;
+    while path.exists() && Instant::now() < deadline {
+        std::thread::sleep(Duration::from_millis(50));
+    }
+}
+
 #[test]
 fn daemon_status_and_stop_lifecycle() {
     let root = tempfile::tempdir().expect("temp root");
@@ -140,6 +158,7 @@ fn daemon_status_and_stop_lifecycle() {
         !is_process_alive(worker_pid),
         "worker daemon pid {worker_pid} still alive after `kin daemon stop`"
     );
+    wait_until_gone(&pid_path, Duration::from_secs(5));
     assert!(
         !pid_path.exists(),
         "daemon.pid must be cleared after a confirmed stop"


troyjr4103:
Not landing this, and the reason is in its own CI run. Full root cause on FIR-1991.

`Windows authority tests` on `de4b48d27` **failed** — the job this PR exists to de-flake. Same test, same message, at the shifted line: `daemon_stop.rs:162` is `143 + 19` inserted lines, i.e. the same `!pid_path.exists()` assertion. The new bounded wait ran its full 5 seconds and the file was still there.

The premise does not match the code. `kin daemon stop` does not leave the unlink to the exiting daemon; it removes the pid file **itself, synchronously**, before returning:

```rust
let outcome = stop_worker_at(&kin_root, pid, stop_timeout(), None)?;
if outcome.is_success() {
    remove_stale_daemon_files(&kin_root);
}
```

and that call is guarded by a liveness probe where `authorizes_cleanup()` is `matches!(self, Self::Dead)`. `Alive` and `Unknown` both **skip** the removal, log a warning, and return `()`. Nothing retries. On Windows the probe reports `Alive` whenever `GetExitCodeProcess` still says `STILL_ACTIVE`.

So when the OS has not finished teardown at the instant the stop returns, cleanup is silently skipped and `daemon.pid` stays forever, while the command reports success. That is the intermittency, and it is why a wait in the test cannot close it: the skip decision has already been made and nothing re-attempts it.

To be clear about what this PR got right: `wait_until_gone` only waits, and the original assertion is untouched, so it **failed loud instead of passing vacuously**. A change that widened or swallowed the assertion would have gone green here and buried the finding. That property is what made the root cause findable.

The assertion is correct and should not be softened. `daemon.pid` is what this codebase calls the thing that makes an endpoint published, so a survivor after a successful stop is a stale published endpoint.

Three fix directions are on FIR-1991; the shortest is to stop swallowing the skip, since `remove_stale_daemon_files` returns `()` and the caller cannot tell cleanup did not happen.

troyjr4103:
Superseded by #642, merged as `b00fdb76727dfad2d46398a697f57e15c76d1c33`.

This PR did its job: its own red CI is what disproved the async-unlink premise it was written on, and that is what led to the actual root cause. The bounded `wait_until_gone` it added is **absorbed unchanged** into #642 as the end-to-end regression, sitting beside the assertion it never weakened.

The real defect was not a race the test could wait out. `remove_stale_daemon_files` probed liveness ONCE and skipped retirement on `Alive` or the indeterminate `Unknown`, warned, and returned `()`; nothing retried the skip, so a probe landing inside the OS teardown window decided the endpoint's fate permanently while `kin daemon stop` exited 0. By the time this PR's five-second wait started, the decision not to remove had already been made.

#642 waits boundedly for the recorded owner to be affirmatively gone before deciding, judges the incarnation in `daemon.owner` rather than the bare PID, and returns a `#[must_use]` outcome so the stop path exits nonzero naming the preserved pid file.

Proof it works on the platform that failed: `Windows authority tests` **passed** on #642 while the same job was red on `main` and red on an unrelated PR of mine that did not carry the fix.

Closing unmerged. Nothing here is lost.

PR #761, opened 2026-08-11
Report a completed fill as ready when new work arrives after it

troyjr4103:
The v0.5.18 Public Install Proof ran `kin embed` to completion, read complete observed coverage out of `kin status`, and then had the health report call the same store `pending`. On ubuntu-latest the embed reported no pending entities and no pending artifacts, status reported 14 indexed of 14 total with 0 pending, and about seventy milliseconds later readiness read that store as 14 of 17 with 3 pending and blocked the capability gate. macos-15-intel failed the same way at 13 of 16. The three files in between were ones the proof itself had just written, admitted into the working copy as they arrived.

Two defects compound there, and fixing either alone leaves the gate red.

An explicit `kin embed` that drained its queue never recorded that the fill finished. Only the background worker published the has-ever-completed marker, at its own drain point, and that worker stands down for the duration of an explicit pass and breaks out before it ever reaches the publish. A store filled by `kin embed` therefore went on reporting a first embedding pass still filling for the rest of its life, which is false on its face.

The classifier then read the same partial counters as lost ground and returned stale, and stale on this check flips the whole report unhealthy by design. That reaches well past CI. Entity-granular embedding invalidation means an edit invalidates the embeddings it touched, and a watched working copy admits new files as they are written, so coverage goes partial again constantly on any repository somebody is actually working in, and comes back on its own every time.

So record the completion from the pass's own report rather than from a counter re-read afterwards, which cannot survive work arriving in the moment between, and treat a backlog on a store whose fill completed as a top-up that is filling rather than as coverage lost. The counts stay in the detail either way. The states that do mean lost ground keep their own arms, keyed on the cause instead of inferred from counters that grew: a discarded index stays stale, a failed worker stays missing. A backlog nothing will consume, on a store whose graph authority is a remote storage backend with no durable local vector-sidecar contract, now reports unsupported with that cause named rather than promising progress that host cannot make.

The tests carry their own falsification. The daemon-side one asserts first that recording from a counter re-read publishes nothing on its fixture, which is the old behavior failing, and then that recording from the drained pass succeeds. The classifier tests assert healthy on the exact counters from the failing run and then flip the store back to one that never finished a fill and require pending again, and the unsupported arm has to return to pending the moment local persistence is available. End to end, on a fixture mirroring the proof, the same script reads stale and blocks the aggregate before this change and healthy with the backlog named after it.

One note for whoever plans the next cut. This closes two of the three gating legs. ubuntu-24.04-arm failed somewhere else, at the semantic locate coverage assertion, where the payload read 13 indexed of 18 with 5 pending one second after readiness had read 13 of 13 and healthy. That is the same corpus drift seen from another surface, it is untouched here, and the other legs become exposed to it once they stop dying early.

Closes FIR-2203


troyjr4103:
Merge-queue ejection record, 2026-08-11 22:16Z: the speculative batch containing this PR failed Check & Test (ubuntu) in run 31540443053 on a single test, branch_switch_preserves_graph_only_gitlinks_without_traversing_nested_checkout, with a 409 refusal over vendor gitlink paths (owned.txt, untracked.bin). PR-level CI on this head is green and this change does not touch branch switching, so first disposition is flake-or-batch-interaction; the queue captain will requeue once the current batch lands. If it ejects a second time on the same test, treat the interaction between this fill-readiness change and the kin#748 pending-workspace carry as real and investigate before resubmitting.

troyjr4103:
Follow-up to the ejection record: this PR is exonerated. Three batches failed on three DIFFERENT tests (git_init killed-mid-capture on macOS, branch_switch gitlinks on ubuntu, captured-init phase summaries on ubuntu with a git multi-pack-index maintenance lock visible as the root cause), including a batch containing only main plus the workflow-only kin#762. The queue is being thrashed by known init-wave timing races, not by this change. Tracked with evidence on the flake-family ticket; the captain will resubmit this PR once the current batches settle.

PR #777, opened 2026-08-12
Make the retrieval negative consume the signals printed beside it

troyjr4103:
One `semantic_locate` reply carried a coverage figure, a populated `degradations[]`, and a `negative` object reporting coverage unknown and no degraded signals. A reader keying on the top level concluded full coverage, one keying on the envelope concluded unknown, and both stood next to an active degradation the verdict ignored. Nothing here over-claimed absence, so this is honesty coherence rather than a fabrication risk, but an agent reading two contradictory fields learns which field to trust rather than which answer.

Four seams, all in envelope assembly.

**The cosine arm's coverage could not be read.** It publishes coverage as a bare `indexed / total` float, which carries no counts to build an envelope field from, so the lift skipped it and the negative beside it said coverage unknown. The counters now ride alongside under `semantic_coverage_detail`, the key the fused arm already uses, taken from the same single observation as the float so the two cannot disagree. The legacy float is unchanged on the wire.

**The payload's own degradations are now consumed.** They appear in `degraded_signals` and in the advice sentence, and they downgrade `safe_to_conclude_absent`. An entry missing `component` or `reason` is dropped rather than published as a half-named label. `trace_data_flow` keeps stating this in its own walk vocabulary instead of saying it twice.

**An empty graph no longer certifies absence.** A resolved-but-empty result on a graph holding no entities is a fact about the graph. The resolution-miss path already refused to certify that; the two ways a tool reports nothing now agree.

**`semantic_search` moves to the structural class.** It filters the entity index by name, kind, language, and role and never reads a vector, which is why kin-daemon deliberately withholds embedding coverage from its payload. Gating it on embedding coverage made every empty search report `coverage_unknown` and advise re-checking after embedding completes, on stores whose embeddings were already complete, about a lookup that read none. Its absence now answers to the graph gate, which is the substrate it actually reads. The alternative fix, feeding it coverage, was rejected: the only cross-tool source is HEAD-scoped `/health`, and borrowing that would mix two graph authorities in one response, which is exactly what the graph-status path goes out of its way to avoid.

Tests fail before and pass after, each with a control that can distinguish the states: a degraded page against a clean one, an empty graph against a populated one, a malformed degradation entry against a well-formed one, and `semantic_search` on a loaded graph against an unloaded one. Three existing class-boundary tests moved from `semantic_search` to `semantic_locate`, which is the tool that is actually embedding-backed, so the Semantic gate keeps its coverage.

Full lane gate green from the lane worktree: 5518 tests, doctests, no tracked lock contamination.

Closes FIR-2216

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -27368,12 +27368,18 @@ mod tests {
             .expect("cosine payload carries the structured coverage counters");
         for counter in ["indexed", "total", "pending"] {
             assert!(
-                detail.get(counter).and_then(|value| value.as_u64()).is_some(),
+                detail
+                    .get(counter)
+                    .and_then(|value| value.as_u64())
+                    .is_some(),
                 "semantic_coverage_detail must carry {counter}: {detail}"
             );
         }
         assert!(
-            detail.get("complete").and_then(|value| value.as_bool()).is_some(),
+            detail
+                .get("complete")
+                .and_then(|value| value.as_bool())
+                .is_some(),
             "semantic_coverage_detail must state completeness: {detail}"
         );
         let lifted = kin_mcp::Envelope::daemon().with_payload_metadata(&parsed);
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -6720,6 +6720,7 @@ fn semantic_locate_payload(
     query: &str,
     file_granularity: bool,
     semantic_coverage: f32,
+    coverage_detail: &kin_cli::commands::locate::SemanticCoverage,
     key: &str,
     page: usize,
     page_size: usize,
@@ -6739,6 +6740,7 @@ fn semantic_locate_payload(
         "granularity": if file_granularity { "file" } else { "entity" },
         "routing": "cosine-v0",
         "semantic_coverage": semantic_coverage,
+        "semantic_coverage_detail": coverage_detail,
         "page": page,
         "total_ranked": total,
         "next_cursor": next_cursor,
@@ -6960,11 +6962,19 @@ fn build_semantic_locate_result(
 
     // Coverage is reported, never gated: a partially-embedded graph still
     // returns whatever the index can answer (graceful degradation per R5).
-    let status = graph.embedding_status();
-    let semantic_coverage = if status.total == 0 {
+    //
+    // Both shapes come from ONE observation. The legacy `indexed / total` float
+    // is what this arm has always published, and it is the only shape here that
+    // no response envelope can be built from: a bare number carries no counts,
+    // so the negative beside it reported coverage unknown next to a coverage
+    // figure this same payload had just printed. The counters ride alongside
+    // under the key the fused arm already uses, and taking both from one
+    // measurement is what keeps them from disagreeing.
+    let coverage_detail = kin_cli::commands::locate::local_semantic_coverage(graph, None);
+    let semantic_coverage = if coverage_detail.total == 0 {
         0.0_f32
     } else {
-        status.indexed as f32 / status.total as f32
+        coverage_detail.indexed as f32 / coverage_detail.total as f32
     };
 
     // Paging fast path: a valid cursor whose ranking is still cached at the
@@ -6990,6 +7000,7 @@ fn build_semantic_locate_result(
                     &query,
                     file_granularity,
                     semantic_coverage,
+                    &coverage_detail,
                     &parsed.key,
                     parsed.page,
                     page_size,
@@ -7264,6 +7275,7 @@ fn build_semantic_locate_result(
         &query,
         file_granularity,
         semantic_coverage,
+        &coverage_detail,
         &key,
         0,
         page_size,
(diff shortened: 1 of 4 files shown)


troyjr4103:
## Adversarial review: BLOCKING. Do not submit this to the queue as-is.

This PR's head (`58a0b558`) certifies a false absence. Caught pre-submit, so it is cheap now: auto-merge is not armed and the branch is BEHIND.

### The defect

The coverage lift is right, and it is also what makes the emptiest possible index authoritative for the first time.

Before this change, `with_payload_metadata` read only `semantic_coverage`. On the cosine arm that key is a float, `as_object()` fails, coverage stays `None`, and the Semantic gate answered `coverage_unknown`, so `safe_to_conclude_absent` was `false` unconditionally. After the lift it returns `true` whenever `complete` is set, and **the default profile serves cosine**.

An index holding no vectors reports `total == 0`, and `embedding_status_complete` calls that complete, correctly: nothing is missing when nothing is eligible. That answers a coverage question. For an absence question the same number means a ranked search examined nothing. So one reply now carries:

- `semantic_coverage: 0.0` (the float, forced to zero at `api.rs:6974-6976`)
- `semantic_coverage_detail.complete: true`
- `negative.semantic_coverage.percent: 100.0`
- `negative.safe_to_conclude_absent: true`, advice "Absence is authoritative: safe to treat the target as genuinely absent/unused."

A reader keying on the top level sees 0% coverage. A reader keying on the envelope sees 100% and an authoritative absence. That is the contradiction this PR set out to remove, moved one field over and made confident instead of unknown.

`graph.semantic_search` returns `Ok(vec![])` rather than erroring on a missing or empty vector index, so this is an ordinary empty payload, not an error path.

### Reproduced, not argued

With the fix's gate removed, the added test fails `left: Bool(true), right: Bool(false)` on `safe_to_conclude_absent` and cargo exits 101.

### The fix is ready to take

`b38b9952` on `origin/chore/lane-neg2216-kin`, stacked directly on `58a0b558`:

```
git cherry-pick b38b9952
```

It adds a `coverage_empty` arm so completeness has to hold over a non-empty index, with the exact cosine payload as the test plus a populated positive control so the gate reads the count rather than always firing. Verified: fmt clean; `clippy -p kin-mcp --all-targets --all-features` clean under `-D warnings` with CI's allowlist; `cargo test -p kin-mcp --lib` 376 passed; `cargo check -p kin-daemon --all-targets` clean; the 4 kin-daemon negative-envelope tests pass. Full `cargo test --workspace` was not run.

### Second finding, filed separately as FIR-2230

Not fixed here and it must not be closed by this PR. The `graph_empty` gate added in this PR reads `envelope.graph_state.entity_count`, which comes only from `/health`, and `/health` reports the HEAD graph. But non-mutating MCP tools answer from `state.graph_for_request_with_authority(session_id)`, which returns a reconstructed historical graph under a live temporal session scope. So the gate written to stop absence being certified on an empty graph reads the count of a different graph and is blind in exactly that case. `semantic_search` newly joins the structural family here, and it previously could never return `true` at all.

This repo already documents the hazard twice about this endpoint, at `server.rs:1016-1020` and `envelope.rs:256-263`.

### Three non-blocking notes

1. `api.rs:28468-28472` asserts `semantic_coverage == semantic_coverage_detail`, commented "the two coverage surfaces agree rather than drifting". On the cosine arm they can never be equal, and at `total == 0` they mean opposite things.
2. `envelope.rs:354-355` says the key "carries the counters on both arms". True in shape, false in provenance: the fused arm passes the HEAD graph as vector source, the cosine arm passes `None`.
3. `coverage_from_status` is cfg'd on kin-cli's `vector` feature while `embedding_status` is cfg'd on kin-db's, so a build with one and not the other reports `complete: false` beside a float of `1.0`. Conservative direction, but new coupling.

Full review: `.kin-coord/reports/honesty-fir2216-review-20260812.md`

troyjr4103:
Evidence from the v0.5.19 re-dogfood (2026-08-12) that lands in this PR's blast radius, specifically `crates/kin-mcp/src/negative.rs`. Flagging it here rather than opening a competing branch, since this PR is already editing that file today. Full detail in `.kin-coord/reports/kin-dogfood-0519-20260812.md`, findings G3 and G5.

**This PR's premise is confirmed on the shipped artifact (G5).** One `semantic_locate` reply on v0.5.19 carries top-level `semantic_coverage: 1.0`, a populated `degradations[]` naming `vector_sidecar / retired_entity_keys`, and `_kin.semantic_coverage: null`, while `semantic_search` and `find_references` on the same daemon seconds later report `negative.semantic_coverage: null` and downgrade to `trust: inconclusive`, `trust_reason: "coverage_unknown: embedding coverage was not reported"` with advice ending "Re-check after embedding is complete and the daemon is healthy". Embedding was complete (`indexed: 10432, pending: 0`) and the daemon was healthy. The advice text also asserts "no degraded signals" beside a populated degradation block. Exactly the contradiction this PR describes.

**The part worth adding to this pass, because it is the same predicate and it fails in the dangerous direction (G3).** FIR-2216 notes the current direction is safe because trust stays inconclusive and nothing over-claims absence. On the fused arm at full coverage that is no longer true. A bare exact symbol name produces:

```json
"kind": "no_named_match",
"interpretation": "unnamed_ranking",
"safe_to_conclude_absent": true,
"trust": "authoritative",
"trust_reason": "semantic_authoritative: daemon-owned truth with complete embedding coverage and no degraded signals",
"advice": "…no ranked entity carries the name the query asked for, and on a complete graph
           that means no entity carries it at all."
```

The symbol exists. `find_references` for the same bare name, 1.9 s later in the same driver run, resolved `InMemoryGraph::prune_orphaned_vectors` (`6534bdb7-e447-4e65-bf94-69436cc1952e`, `crates/kin-db/src/engine/graph.rs`). Control: a fabricated symbol on the same store in the same run returned the byte-identical envelope, so the envelope cannot separate a symbol locate missed from one that is absent.

The unsound step is the inference, not the plumbing this PR is fixing. "No ranked entity carries this name" does not imply "no entity carries this name", because ranking is a bounded candidate set, and coverage completeness says nothing about a name the ranker never considered. Once this PR makes the negative consume the signals beside it, the same assembly site is the natural place to gate the conclusion:

> `safe_to_conclude_absent` must not be true on `no_named_match` unless an exact-name index lookup actually ran and missed.

Composes with FIR-2191's exact-name entity admission, which is the real fix for why the symbol was missing from the page at all. If that is out of scope here, ignore it and I will carry it on FIR-2178, which I have recommended reopening with this evidence. Raising it now only because the file is open in front of you.

Two notes so nobody has to re-derive them:

- kin#741 is confirmed present in the v0.5.19 binary by `git merge-base --is-ancestor` against `adf8b395`, with a negative control, so this is a live regression rather than an unshipped fix.
- On the `compat-v0` default arm there is still no `negative` key at all on a non-empty page, plus no `all_fallback` and no `match_kind`. That half is kin#786 and is not this PR's problem.

PR #791, opened 2026-08-12
Stop an unnamed ranking from certifying that the symbol is absent

troyjr4103:
The v0.5.19 dogfood asked a fully covered kin-db store for `prune_orphaned_vectors`, a bare exact symbol name. `semantic_locate` returned ten entities, none of them the target, and attached this:

```json
"kind": "no_named_match",
"safe_to_conclude_absent": true,
"trust": "authoritative",
"trust_reason": "semantic_authoritative: daemon-owned truth with complete embedding coverage and no degraded signals",
"advice": "...no ranked entity carries the name the query asked for, and on a complete graph
           that means no entity carries it at all."
```

`find_references` resolved that exact name to a real method 1.9 seconds later in the same driver run: `InMemoryGraph::prune_orphaned_vectors`, `crates/kin-db/src/engine/graph.rs`. **The entity exists and the envelope said authoritatively that it does not.** The fabricated control `zzqqxx_nonexistent_symbol_9f3a` returned the *identical* envelope on the same store in the same run, so the verdict could not separate a symbol retrieval missed from one that does not exist, and stamped both authoritative.

Against 0.5.17 this is a regression in kind, not degree. There, locate returned confident wrong rows and said nothing. Here it returns confident wrong rows and adds a truth claim.

**The inference was wrong, not the plumbing.** A ranking is a bounded candidate set. Complete coverage says every entity has an embedding; it does not say the ranker considered every entity, and a name absent from a window says nothing about the rows outside it. So `no_named_match` is now never authoritative, regardless of coverage.

The reason names the bound and keeps the substrate observation after it rather than replacing it, because both facts are true and the pair is the point: on a complete index it reads "the substrate is fine, the ranking is the limit", which is exactly the distinction that was being collapsed. The advice stops asserting graph-wide absence and sends the caller to `find_references` or `semantic_search`, which resolve a name directly and which the same dogfood found answer this correctly.

**Nothing the envelope was entitled to claim is lost.** An empty page on the same envelope stays `safe_to_conclude_absent: true`. The bound applies to a populated ranking that missed a name, not to a query that ranked nothing, and the test asserts that third case explicitly so this cannot quietly become a blanket downgrade.

The test carries the dogfood's own controls: the existing symbol, the fabricated symbol (both inconclusive now, which is the honest answer for both, because this surface cannot tell them apart and must not pretend to), and the empty page that stays authoritative.

**Why this is a separate PR.** It was written on #777, whose coverage lift is what makes this verdict reachable on the default arm at all: before it, the cosine arm reported coverage unknown and could never certify. I intended to ship them together, #777 entered the merge queue first, and a queued branch cannot be updated. So this lands right behind it and closes the window. Reviewers of #777 should know that landing it alone widens this claim from the fused arm to the default one, which is the reason this PR should not wait.

Verified before opening: `cargo fmt --all -- --check` clean, `cargo clippy --all-targets --all-features` clean under ci.yml's allowlist, `negative::` suite 55 passed standalone on `main` without #777, and the full lane gate green at 5544 tests plus doctests with no tracked lock contamination.

Not claimed: this does not add a relevance floor and does not change which rows are returned or their order. It changes only what the envelope is willing to certify. FIR-2178's ranking-quality half is untouched.

Advances FIR-2178

### DIFF
--- a/crates/kin-mcp/src/negative.rs
+++ b/crates/kin-mcp/src/negative.rs
@@ -449,12 +449,12 @@ fn absence_consequence(always: bool, trustworthy: bool) -> &'static str {
 /// empty result: here the rows are real and stay ranked, and what is absent is
 /// the NAME, so the advice has to separate "these are neighbors" from "the
 /// symbol does not exist" instead of collapsing them into one verdict.
-fn unnamed_ranking_consequence(trustworthy: bool) -> &'static str {
-    if trustworthy {
-        "These hits are neighbors, not the symbol: no ranked entity carries the name the query asked for, and on a complete graph that means no entity carries it at all. Do not treat the top hit as the requested symbol."
-    } else {
-        "These hits are neighbors, not the symbol: no ranked entity carries the name the query asked for. The name's absence is NOT authoritative — it may simply not be indexed yet — so do not conclude the symbol does not exist. Re-check after embedding is complete and the daemon is healthy."
-    }
+fn unnamed_ranking_consequence() -> &'static str {
+    "These hits are neighbors, not the symbol: no ranked entity carries the name the query \
+     asked for. That is a fact about this ranking, not about the graph, because a ranking is \
+     a bounded candidate set and the name may belong to an entity it never considered. Do not \
+     treat the top hit as the requested symbol, and do not conclude the symbol does not exist: \
+     settle that with find_references or semantic_search, which resolve a name directly."
 }
 
 /// The kind the payload reports for its focal entity, whichever shape carries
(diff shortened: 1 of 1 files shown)


troyjr4103:
Compatibility with #777 verified locally, since this PR is based on `main` without it and both touch `negative.rs` in the same function region.

I materialized the merge of the two heads and gated the result rather than reasoning about it:

- `git merge-tree --write-tree --messages` on the two heads: clean, three files auto-merged (`api.rs`, `entities.rs`, `negative.rs`), zero CONFLICT lines.
- The check was falsified first, because a merge check that cannot report a conflict is not evidence: a probe branch editing the same function a second way produced `rc=1` and `CONFLICT (content): Merge conflict in crates/kin-mcp/src/negative.rs`. So the clean result above is a real negative.
- The merged tree was then built and tested, not just merged. The signature and the call site land consistently at the zero-argument form (`unnamed_ranking_consequence()` at both definition and call), `negative::` passes 59, and the full lane gate on the merged commit is green at **5550 tests** plus doctests with no tracked lock contamination.

So whichever order the queue lands these in, the pair compiles and passes. Nothing is needed from a reviewer here beyond the ordering note already in the PR body: #777 alone widens the false-authoritative verdict to the default arm, and this closes it.

PR #802, opened 2026-08-13
Retain tracked artifact text at ingest so retrieval has something to index

troyjr4103:
A docs store still could not answer for its own content on v0.5.21, with the store fully embedded, after the candidacy door for exactly this shipped in kin#760. The door was never the obstacle: it reads its candidates from a text search, and the text search had nothing to find. Real ingest stored every opaque artifact with `text_preview: None` in both branches of `IndexPipeline::index_any_content`, ref materialization capped the preview at 64 words, and kin-db indexes an opaque artifact's preview or nothing (`opaque_artifact_fields`), then embeds that same preview or nothing (`format_opaque_text`). On a real store both retrieval arms therefore held one path, one MIME type, and the literal word opaque_artifact per document, and every phrase past the head of a file was unfindable by construction. kin#760's twelve tests all passed because their fixtures handed `put_opaque` the full body as the preview, a state no ingest path ever produced.

Ingest now retains bounded full text for textual artifact bytes, in one helper both live enrichment and ref materialization call. A textual MIME hint qualifies the bytes outright; without one, printable-majority UTF-8 qualifies, so extensionless text is kept and binaries stay out. The bound is 256,000 characters, matching the retention ref materialization already grants historical source text. Because `upsert_opaque_artifact` re-queues the artifact for embedding, the retained text reaches the vector arm through the existing queue with no new plumbing.

That bound is a retention cap rather than a display size, so each surface that renders the text sets its own. Search records still render the 320 characters they carried before. Context packs needed a real change, because a pack entry took the retained text verbatim and admission stopped at the first entry it could not afford instead of skipping it. Measured on a 53 KB doc against the default 8,000 token budget, that one entry estimated 8,957 tokens, and since entries sort by file path it emptied the pack for every other file too, taking supporting artifacts from two entries to none. Each entry is now excerpted to 2,000 characters and admission continues past what it cannot afford, so both entries hold. The whole text stays reachable through the artifact read path.

Six new tests pin it. Markdown through the real enrichment keeps a phrase placed past any head-sized preview. An extensionless text file qualifies by content and the bound holds exactly. The existing binary case gains an assertion that binary bytes still retain nothing. The ticket's own shape runs the real pipeline output through a store write, the text index, and locate, and requires the artifact-labelled row for a phrase 2,000 characters deep. Another holds the same deep retention on the ref materialization path. The last two cover the pack, one per half of that change. Reverting the two pipeline lines fails the three deep-text tests while the binary control and kin#760's own fixture tests stay green, which is the original defect reproduced in miniature: fixture green, real ingest red. Reverting either half of the pack change fails that half's test and only that one.

Not claimed: stores admitted before this change keep their empty previews until their files re-enrich, so felt acceptance on an already-built store still needs a fresh admit; the vector arm stays dark until FIR-2202 lands, because artifact vectors do not reach real stores yet, so the re-queue above is the plumbing being right rather than the vector arm working; structured artifacts keep their 64-word normalized preview, a shallower version of the same gap left for its own decision because deepening it re-prices manifests; and this changes recall on fresh stores in a way that has not been measured against the benchmark suite.

FIR-2183

### DIFF
--- a/crates/kin-cli/src/commands/locate.rs
+++ b/crates/kin-cli/src/commands/locate.rs
@@ -17341,6 +17341,59 @@ mod tests {
             .all(|entity| entity.id_space == LocateIdSpace::Artifact));
     }
 
+    /// The same promise through the REAL ingest enrichment instead of a
+    /// hand-rolled preview. Every fixture above hands `put_opaque` its full
+    /// body as `text_preview`, which is a state ingest never produced: the
+    /// enrichment pipeline stored every opaque artifact with no text at all,
+    /// so on a real store the text index held only a path and a MIME type and
+    /// every one of these green tests described a store that could not exist.
+    /// This test routes the bytes through `IndexPipeline::index_any_content`,
+    /// exactly what the daemon runs on admit, and puts the phrase past any
+    /// head-sized preview so head-only retention fails it too.
+    #[test]
+    #[serial_test::serial]
+    fn a_pipeline_enriched_docs_store_answers_for_deep_content() {
+        let mut body = String::from("# Kin Ecosystem AGENTS.md\n\n");
+        for index in 0..40 {
+            body.push_str(&format!(
+                "Ordinary paragraph {index} about workspace upkeep and lane hygiene, nothing \
+                 doctrinal here.\n\n"
+            ));
+        }
+        body.push_str("## Verification Rule\n\nChecks That Cannot Fail is doctrine here.\n");
+        assert!(
+            body.find("Checks That Cannot Fail").unwrap() > 2_000,
+            "the phrase must sit beyond any head-sized preview"
+        );
+
+        let bytes = body.as_bytes();
+        let indexed = kin_index::pipeline::IndexPipeline::new()
+            .index_any_content(
+                &FilePathId::new("AGENTS.md"),
+                bytes,
+                kin_blobs::digest(bytes),
+            )
+            .unwrap();
+        let kin_index::pipeline::IndexedAny::OpaqueArtifact(artifact) = indexed else {
+            panic!("markdown must take the opaque enrichment facet");
+        };
+
+        let graph = kin_db::InMemoryGraph::new();
+        graph.put_opaque(&artifact).unwrap();
+        graph.flush_text_index().unwrap();
+
+        let result = agent_locate(&graph, "Checks That Cannot Fail");
+        assert!(
+            artifact_row(&result, "AGENTS.md").is_some(),
+            "a pipeline-enriched docs store must answer for its own deep content; entities: {:?}",
+            result
+                .entities
+                .iter()
+                .map(|entity| entity.name.as_str())
+                .collect::<Vec<_>>()
+        );
+    }
+
     /// The bound holds, and a short page says it was short. A page cut to its
     /// limit with nothing said reads as the store having only that much.
     #[test]
--- a/crates/kin-cli/src/commands/search.rs
+++ b/crates/kin-cli/src/commands/search.rs
@@ -915,6 +915,22 @@ fn entity_to_daemon_record(
     }
 }
 
+/// Characters of artifact text a search record renders.
+///
+/// The stored preview is a retrieval retention (up to
+/// `kin_index::artifacts::ARTIFACT_TEXT_RETENTION_CHARS`), not a display
+/// string; a record that carried it whole would put an entire document on the
+/// wire per row.
+const ARTIFACT_PREVIEW_DISPLAY_CHARS: usize = 320;
+
+fn artifact_display_preview(preview: Option<&str>) -> Option<String> {
+    let text = preview?.trim();
+    if text.is_empty() {
+        return None;
+    }
+    Some(text.chars().take(ARTIFACT_PREVIEW_DISPLAY_CHARS).collect())
+}
+
 fn resolved_item_to_daemon_record(
     item: &ResolvedRetrievalItem,
     match_kind: SearchMatchKind,
@@ -964,7 +980,7 @@ fn resolved_item_to_daemon_record(
             file,
             artifact_kind: format!("{:?}", artifact.kind),
             line: 1,
-            preview: artifact.text_preview.clone(),
+            preview: artifact_display_preview(artifact.text_preview.as_deref()),
             score,
             match_kind: Some(match_kind),
         },
@@ -980,7 +996,7 @@ fn resolved_item_to_daemon_record(
                 .clone()
                 .unwrap_or_else(|| "opaque".to_string()),
             line: 1,
-            preview: artifact.text_preview.clone(),
+            preview: artifact_display_preview(artifact.text_preview.as_deref()),
             score,
             match_kind: Some(match_kind),
         },
@@ -1102,8 +1118,12 @@ fn record_preview(item: &ResolvedRetrievalItem) -> Option<String> {
                 ))
             }
         }
-        ResolvedRetrievalItem::StructuredArtifact(artifact) => artifact.text_preview.clone(),
-        ResolvedRetrievalItem::OpaqueArtifact(artifact) => artifact.text_preview.clone(),
+        ResolvedRetrievalItem::StructuredArtifact(artifact) => {
+            artifact_display_preview(artifact.text_preview.as_deref())
+        }
+        ResolvedRetrievalItem::OpaqueArtifact(artifact) => {
+            artifact_display_preview(artifact.text_preview.as_deref())
+        }
     }
 }
 
--- a/crates/kin-context/src/builder.rs
+++ b/crates/kin-context/src/builder.rs
@@ -847,6 +847,30 @@ where
     Ok(file_ids)
 }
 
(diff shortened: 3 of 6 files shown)


troyjr4103:
Independent adversarial review at head c5861a2b. **NO-GO**, with one blocking regression and two body corrections. The fix itself is real and I proved it.

### The fix works on the arm it targets

I reverted only the two `text_preview` lines in `crates/kin-index/src/pipeline.rs`, left every test untouched, and re-ran. Three new tests go red, including the ticket-shaped one, which returns `entities: []` for "Checks That Cannot Fail" against a pipeline-enriched store. That is FIR-2183's defect reproduced in miniature. `build_graph_at_ref_keeps_deep_opaque_text` correctly stays green, since the revert does not touch `ref_view`. These tests can fail, and they fail only because of the production change.

The mechanism in the body checks out against kin-db 0.7.21: `opaque_artifact_fields` (search/text.rs:481) and `format_opaque_text` (embed/mod.rs:3291) both take `text_preview` verbatim or nothing, and `upsert_opaque_artifact` (engine/graph.rs:7204) does re-queue for embedding.

### Blocking: this empties `get_context_pack` for the same docs stores

`append_supporting_artifacts` (`crates/kin-context/src/builder.rs:912`) admits entries under the token budget with `break`, not `continue`. Entry content for an opaque artifact is `format_opaque_text`, which now carries the full retained text, and entries sort by file path first. So one large doc that sorts early ends the loop and drops every later entry, including small useful structured-artifact and shallow-file entries for other files.

Measured with a 53 KB `AGENTS.md` plus a `Makefile` structured artifact, against a `text_preview: None` control on the same harness:

| | docs entry tokens | budget | `supporting_artifacts` |
|---|---|---|---|
| control (main today) | 12 | 8000 | **2** |
| this PR | **8957** | 8000 | **0** |

Default budget is `TokenBudget::Small8k` = 8000. An ordinary 53 KB doc clears it alone. The ticket is "docs stores cannot answer questions about their own content", and this makes `locate` answer while making the surface an agent reads the content from return nothing for the same store. No existing test covers it, so CI is green on it.

Fix shape, in keeping with what the PR already does for search records: bound the artifact entry content at pack-build time the way `ARTIFACT_PREVIEW_DISPLAY_CHARS` bounds the display preview, and/or change `break` to `continue`. Add a regression test asserting a large opaque artifact does not empty `supporting_artifacts`.

### Body corrections

1. "context packs already consume artifact text under a token budget and are unchanged" is refuted by the table above. The code is unchanged; the behavior inverts. The queue mints this body as the squash message, so it would become the permanent record of the claim.
2. "the retained text reaches the vector arm through the existing queue with no new plumbing" is true at the kin-db API level, but FIR-2202 keeps artifact vectors out of real stores until its own PR lands. Worth naming in the otherwise honest "Not claimed" paragraph.

### Checked and clean

No em dashes in the diff or body. No `cfg` gates introduced, so nothing to pre-flight clippy for another target. No thesis drift: retention happens at the ingest and materialization boundary, and the answer path stays graph-backed. No dependency cycle, and `preview_text` in `ref_view` stays live at line 426 so the delegation leaves no dead code. Bound arithmetic in the exact-equality assertion is correct. No version bump required.

One bounded observation for a follow-up rather than this PR: `extract_source_text_signals` (locate.rs:9377) now populates `full_source_texts` on ordinary stores, which activates a previously-dead symbolic suppression filter and holds up to three copies of the qualifying corpus per call. `graph_source_path_set` excludes docs paths and markdown matches no source-like MIME token, so it is confined to source-like artifacts.

troyjr4103:
Remediation of the review NO-GO, at `fdfbbe0e`. All four items resolved, none refuted.

**The context-pack regression was real, and it was two defects, not one.** Entry content came from
`format_opaque_text`, which appends the retained text whole, so a 53 KB doc estimated 8,957 tokens
against the default 8,000 budget. Separately, admission used `break`, and entries sort by file path,
so one large early file also cost every later file its entry. Fixed both in
`crates/kin-context/src/builder.rs`: a new `ARTIFACT_ENTRY_CONTENT_CHARS` of 2,000 applied to all
three entry kinds, mirroring what `search.rs` does for display, and `break` becomes `continue`.
Applied to shallow and structured entries too, so the failure class is closed rather than the one
instance of it. Neither half is sufficient alone. `continue` by itself leaves the docs entry
unadmitted, which still fails what the ticket is actually about, and the bound by itself leaves
starvation live for any budget tight enough to reject an entry. The whole text stays reachable
through `kin_artifact_read`, which returns blob bytes losslessly from the content-addressed store
and never reads the working directory.

**Two tests, one per defect, each falsified on its own.** Reverting `continue` to `break` fails
`an_unaffordable_artifact_does_not_cost_the_next_file_its_seat` and nothing else. Neutering the
bound fails `a_large_docs_artifact_neither_empties_nor_dominates_supporting_artifacts` and nothing
else. `build_context_pack_from_plan_gates_artifact_metadata_on_budget`, the adjacent test that pins
starved-budget emptiness, stays green under both reverts, so `continue` did not loosen it.

**Body corrected.** The sentence claiming context packs were unchanged is gone, replaced by the
measurement. The "Not claimed" paragraph now names FIR-2202, so the re-queue reads as the plumbing
being right rather than the vector arm working.

**Checked for the same shape elsewhere and found none.** `format_opaque_text` and
`format_artifact_text` have exactly one consumer outside `kin-index`, and it is the pack builder,
so the regression surface was enclosed. Of the other non-test `text_preview` consumers,
`graph_source_text` wants the full graph-owned body by design, and `tracked_non_entity_files` widens
locate matching, which is this ticket's intent and belongs to the follow-up already filed against
`extract_source_text_signals`.

Full lane gate re-run raw to completion on this head: 5,611 tests run, 5,611 passed, 11 skipped,
zero failures, doctests green, tracked `Cargo.lock` restored with no contamination. Both new tests
are in that run.

troyjr4103:
The red on this PR was not this PR, and no code changed to clear it. Recording the diagnosis so a reader who sees it flip green knows what happened.

`Check & Test (ubuntu-latest)` failed on `concurrent_resolutions_from_one_view_leave_one_winner`, a merge-resolution test this branch does not touch. `Check & Test (macos-latest)` read CANCELLED only because the matrix fails fast, so there was one real failure, not two.

The panic came from the shared harness at `crates/kin-cli/tests/common/mod.rs:1342`, where `IsolatedDaemonRuntime::drop` proves the runtime's process group went quiescent. The group still held a live process after the sentinel reap. That is the kill-to-exit window rather than a leak: the cleanup barrier delivers SIGKILL before finalization, and a loaded runner retires the condemned member whole scheduler quanta later.

The same assertion site failed today on five different tests across five different branches:

- this PR, `concurrent_resolutions_from_one_view_leave_one_winner`
- kin#812, `branch_delete_rejects_default_and_checked_out_ref_without_mutation`
- main at `a7cac53d`, run 31731789188, `a_parked_merge_survives_a_daemon_restart`
- kin#410, run 31732299329, `a_daemon_that_served_a_branch_command_keeps_answering_its_recorded_endpoint`
- `gh-readonly-queue/main/pr-771`, run 31656801146, `locate_json_keeps_tracing_warnings_off_stdout`

Main is the decisive one. `a7cac53d` is an ancestor of `origin/main` and is not in this branch at all, so main failed this assertion on a commit this PR does not contain.

Two other explanations were checked and ruled out rather than assumed. It is not a platform-gated path, because ubuntu compiled clean and ran 1768 tests before the failure with zero compile or clippy errors, and `Test` was the only failed step. It is not the excerpt-bounding commit, because this branch changes six files across `kin-cli/src/commands`, `kin-context`, `kin-core`, and `kin-index`, none of them a test file and none of them in the merge-resolution path.

kin#803 already fixes this. Its `settle_pinned_group_member_exits` waits out the kill-to-exit window before finalization, and it settles without judging, so the single post-reap probe stays the only thing that can fail. Worth landing before this PR is submitted: run 31656801146 shows the flake already ejecting from the merge queue, where the failing run names no pull request, so the ejected PR is never marked and keeps reading clean and mergeable while it never merges.

The failed jobs were re-run, and the full local gate was re-run to completion against this branch:

```
kin-dev: local-test: kin  .kin-coord/worktrees/fir2183-kin @ fdfbbe0e (chore/lane-fir2183-kin)
Summary [ 354.076s] 5611 tests run: 5611 passed (1 slow), 11 skipped
```

All six tests this PR adds passed, and so did `concurrent_resolutions_from_one_view_leave_one_winner`. That gate was worth running rather than trusting the rerun alone, because ubuntu's fail-fast left 3801 of 5569 tests unrun, so that leg never actually exercised this PR's own tests. It runs on macOS, so it confirms the code rather than reproducing the ubuntu leg. The exoneration rests on the five-branch CI evidence above.

troyjr4103:
Green, with no code change: 24 SUCCESS, 3 SKIPPED, nothing failed and nothing left pending. Both matrix legs passed on the rerun, including `concurrent_resolutions_from_one_view_leave_one_winner` on ubuntu.

That confirms the diagnosis above rather than settling it. The flake is a race, so passing once is what a race does; kin#803 is still what actually removes it, and it is worth landing before this PR is submitted so the queue does not eject it for someone else's reason.

PR #803, opened 2026-08-13
Settle dying group members and reserve the tests' closed ports

troyjr4103:
Two live CI flakes were ejecting green PRs from the merge queue. Both had the
same shape, a check that judged a transient as a terminal state, and both are
now held still.

The containment guardian judged a SIGKILLed process mid-death. The cleanup
barrier delivers SIGKILL to the whole pinned group, then finalization reaps the
sentinel and probes containment exactly once, and a grandchild the kernel had
not yet retired classified as a live escape. That one probe consumed the
guardian, failed the teardown, and ejected whichever test was unlucky:
locate_json_keeps_tracing_warnings_off_stdout in run 31656801146 (ejected
kin#771), drift_reports_diverged_tracked_bytes_without_admitting_them in run
31640059080. The guardian now settles before it judges. While the unreaped
sentinel still pins the numeric group id, a bounded wait lets condemned members
finish dying, and the single post-reap probe stays the one authority that can
fail, every diagnosis unchanged. FIR-1719 closed the exited-but-unreaped
window; this closes the killed-but-not-yet-exited one.

The daemon-client tests trusted a port to stay closed after releasing it.
closed_loopback_port() bound a listener, took the port, and dropped it, while
1347 sibling tests in the same binary bind ephemeral loopback mocks the whole
time. When the kernel re-handed the port mid-probe, the probe received a real
health answer naming a different repository, which is positive evidence of a
stale record, so retirement proceeded and
a_successor_endpoint_survives_a_verdict_about_its_predecessor got None instead
of LiveNotReady (run 31622548773, ejected kin#781). The helper now returns a
guard that keeps the port bound and never listening: nothing can rebind it and
nothing can ever answer on it. Linux and Windows reset such connects; macOS
drops the SYN, which every prober here reads identically through its own
deadline, and the helper doc records the difference.

Two tests in that file still took a port and released it. In
live_daemon_endpoint_returns_alive_pid_even_before_port_binds that was the same
live defect: it asserts daemon_is_up returns None, daemon_is_up connects to the
recorded port, so a sibling that took the freed port flipped the assertion to
Some(port). Binding a listener on the released port reproduced that failure
deterministically. probing_a_recycled_pid_endpoint_proves_it_invalid used the
pattern without being able to fail, because the probe returns Invalid off
endpoint_owner_liveness before it issues any HTTP request. Both hold the
reservation now, so the file leaves no released-port endpoint for a later test
to copy.

Determinism was measured, not assumed. With the settle stubbed back to the old
single-probe behavior, both new kin-daemon-spawn tests fail; with it real, the
spawn suite is 48/48. Reverting the reservation to a released port fails the new
reservation test on its rebind assertion, and while the reservation is held a
listener can no longer take the port at all. The daemon_client module is 116/116
including that test, and three consecutive full kin-cli lib runs are 1375/1375,
which is the run shape that generates the sibling port pressure. Formatting, the
CI clippy allowlist, and the full workspace gate run clean.

Closes FIR-2259
Closes FIR-2288

### DIFF
--- a/crates/kin-cli/src/daemon_client.rs
+++ b/crates/kin-cli/src/daemon_client.rs
@@ -7908,14 +7908,54 @@ mod tests {
     // later command repeated the same failure until the first daemon exited.
     // A timeout is evidence that a daemon is slow, never that it is dead.
 
-    /// A loopback port with nothing listening: connections are refused
-    /// immediately, so the probe exercises the unreachable-endpoint path
-    /// without waiting on a real timeout.
-    fn closed_loopback_port() -> u16 {
-        let listener = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
-        let port = listener.local_addr().unwrap().port();
-        drop(listener);
-        port
+    /// A loopback port that refuses connections for as long as this guard
+    /// lives, so the probe exercises the unreachable-endpoint path without
+    /// waiting on a real timeout.
+    ///
+    /// The refusal has to hold for the width of the probe, not just at
+    /// acquisition. This test binary runs hundreds of sibling tests that bind
+    /// ephemeral loopback listeners the whole time, so a port that was merely
+    /// observed closed and then released can be handed by the kernel to a
+    /// sibling's mock daemon mid-probe. The probe then receives a real health
+    /// answer naming a different repository, which is positive evidence of a
+    /// stale record, and the verdict the test meant to be about "nothing
+    /// listens here" becomes a retirement. Holding the socket bound but never
+    /// listening keeps both properties at once: without a listen queue every
+    /// connect is refused, and while the reservation is held no other socket
+    /// can bind the port.
+    struct ReservedClosedPort {
+        port: u16,
+        _reservation: tokio::net::TcpSocket,
+    }
+
+    fn reserved_closed_loopback_port() -> ReservedClosedPort {
+        let socket = tokio::net::TcpSocket::new_v4().expect("create the port reservation socket");
+        socket
+            .bind("127.0.0.1:0".parse().expect("loopback bind address"))
+            .expect("bind the port reservation");
+        let port = socket.local_addr().expect("read the reserved port").port();
+        ReservedClosedPort {
+            port,
+            _reservation: socket,
+        }
+    }
+
+    #[test]
+    fn a_reserved_closed_port_refuses_connections_and_cannot_be_rebound() {
+        let reserved = reserved_closed_loopback_port();
+        let address = std::net::SocketAddr::from(([127, 0, 0, 1], reserved.port));
+
+        let error = std::net::TcpStream::connect(address)
+            .expect_err("a reserved closed port must refuse connections while held");
+        assert_eq!(
+            error.kind(),
+            std::io::ErrorKind::ConnectionRefused,
+            "refusal must be immediate, not a timeout: {error}"
+        );
+        assert!(
+            std::net::TcpListener::bind(address).is_err(),
+            "no listener may take the port while the reservation is held"
+        );
     }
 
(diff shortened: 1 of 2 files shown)


troyjr4103:
Independent review. Both fixes hold up under falsification and CI is fully green, but one
thing in the title's scope is still open, so this is a hold rather than an approval.

`live_daemon_endpoint_returns_alive_pid_even_before_port_binds` at daemon_client.rs:10300
still binds a listener, takes the port, drops it, writes it as `daemon.port`, and then
asserts `daemon_is_up(...) == None`. That assertion is a 200ms connect against a port this
binary just handed back to the kernel, with 1374 siblings binding ephemeral loopback mocks
around it. Same mechanism as FIR-2259, same binary, narrower window.

Reproduced deterministically. Adding one line after the drop,
`let _hijack = std::net::TcpListener::bind(("127.0.0.1", port)).unwrap();`, fails it:

```
  left: Some(59067)
 right: None
```

Fix is to use `reserved_closed_loopback_port()` the way the other nine sites now do.
`probing_a_recycled_pid_endpoint_proves_it_invalid` at :10164 uses the same pattern but is
benign, because `probe_daemon_endpoint_with_warming_signal` returns `Invalid` off owner
liveness before it issues any HTTP request. Worth converting for consistency while you are
in there.

What was checked and passed:

- Stubbing the settle back to a single probe fails both new kin-daemon-spawn tests, with
  `LiveMember { pid } != OnlyExited` and `gave up on a live member after 18.209µs`.
- Releasing the reservation fails the new reservation test on the rebind assertion. The
  connect assertion alone does not discriminate, since a released port also refuses.
- The settle is on the path the logs blame: `terminate_and_confirm` reaches
  `finalize_owned_process_group` through `confirm_containment_empty` and `try_reap`, and
  kin-daemon-spawn is a workspace path dep, so it lands without a publish.
- Running the settle before the reap does not starve it. `reap_child_until` calls
  `try_wait` before it checks the deadline, so an exited sentinel is still collected after
  a full-budget settle. Added worst case is 1s on `try_reap`.
- 48/48 kin-daemon-spawn, 116/116 `daemon_client::tests`, and `cargo test -p kin-cli --lib`
  three consecutive times at 1375/1375, which is the run shape that produces the port
  pressure in the first place.

One residual gap for later, not a blocker: deleting the settle call from
`finalize_owned_process_group` leaves all 48 spawn tests green. The compiler catches
deletion, since `-D warnings` turns the now-unused function into an error, but nothing
catches a mis-ordering such as settling after `sentinel.take()`.

troyjr4103:
Evidence this is load-bearing right now, gathered while diagnosing a red on kin#802.

The assertion this PR settles, `crates/kin-cli/tests/common/mod.rs:1342`, failed today on five different tests across five different branches, including main and the merge queue itself:

- kin#802, `concurrent_resolutions_from_one_view_leave_one_winner`
- kin#812, `branch_delete_rejects_default_and_checked_out_ref_without_mutation`
- main at `a7cac53d`, run 31731789188, `a_parked_merge_survives_a_daemon_restart`
- kin#410, run 31732299329, `a_daemon_that_served_a_branch_command_keeps_answering_its_recorded_endpoint`
- `gh-readonly-queue/main/pr-771`, run 31656801146, `locate_json_keeps_tracing_warnings_off_stdout`

Every one carries the shape this PR describes: the group still held a live member after the sentinel reap, which is the kill-to-exit window rather than a leak.

The merge-group failure is the one worth acting on. That run belongs to the merge group rather than to any pull request, so the PR it ejected was never marked on itself, and it kept reading clean and mergeable while it never merged. This flake is already costing the fleet a defer rotation per ejection, quietly.

Landing this before the rest of today's wave is submitted would stop clean PRs being ejected for a reason that has nothing to do with them.

PR #804, opened 2026-08-13
Name why an embedding count is zero instead of leaving it to read as loss

troyjr4103:
`kin graph status` after a dirty daemon restart announced `Embeddings: 0/10456 indexed (10456 pending)` with a pending warning, on a store that recovered seconds later with zero embed dispatches. Nothing was wrong, and the surface said everything was. That is the v0.5.21 battery's B7 finding, reproduced on shipped v0.5.23.

The zero is structural rather than measured. A dirty restart leaves a sidecar whose stamp trails graph truth, the daemon refuses to install it at open, and while no index is attached `embedding_status` answers zero indexed for every retrievable object. The counters are true, so they were never the problem. What a reader cannot get from them is why, and a zero with no cause reads exactly like a repository that lost its work.

So the counters always render now, and the daemon's reason is appended when it has one. A discard at open names itself beside the numbers and says the daemon restores coverage in the background, which is what tells an operator no manual GPU pass is owed. A store with no attached index and no discard on record is told the live graph carries no vector index, the same words `kin status` already uses for that absence, plus the coverage-completion marker when it is set.

No clause claims the vectors are sitting intact somewhere, because nothing on this path can know that. `load_validated_vector_index` returns `None` both when the index loaded and when there was none to load, and a loaded index stays attached, so that branch is only reached when there was no sidecar to load or a later reset detached one. Both of those rebuild rather than re-attach. An earlier revision of this branch read the same `None` as "nothing was discarded" and told the reader to await the index rather than re-embed, which was false in every state that could reach it, and it dropped the pending warning that was the only remaining signal.

A store whose graph authority is a remote backend never embeds at all. Its worker does not start and `/embed` refuses, so the discard clause would name a background recovery that will never run. That fact now outranks both the discard reason and the marker here, matching the order semantic query readiness already checks them in.

The three runtime facts travel in `commands::resources::EmbedRuntimeState`, the carrier `/commands/resources` and readiness already use, and the daemon builds it in one named function so a test can drive the real construction instead of asserting a hand-built literal. A session-scoped graph still gets the default and renders as before.

The coverage counter and the pending warning now come from one embedding-status sample, so they can no longer disagree by exactly one embed batch. That was the battery's B12 residue, one of twelve paired samples off by 512. The two were not even drawn from the same population: `graph_stats().pending_embedding_count` counts entity ids while `embedding_status()` counts retrievable keys, so they could differ at a single instant with no race at all.

Both surfaces that described a discard as a from-scratch rebuild stop doing so. The open-time daemon log and `kin health` now say the same thing, that the daemon restores coverage in the background and reuses prior vectors where they still apply. That is what the dirty-restart repro observed, full coverage restored with zero embed dispatches.

The new arms are pinned by value, and each assertion was checked against its own canary: restoring the rejected claim, dropping the remote-backend arm, and putting "from scratch" back each turned the matching test red. One of those tests opens a real store with no sidecar and a coverage marker, reads both facts off the daemon rather than asserting them into a struct, and renders through the production builder.

The root cause, a dirty restart guaranteeing a whole-sidecar discard, is filed as FIR-2290 and is not changed here.

Closes FIR-2289

### DIFF
--- a/crates/kin-cli/src/commands/graph.rs
+++ b/crates/kin-cli/src/commands/graph.rs
@@ -422,10 +422,17 @@ fn build_graph_status_response(
     lines.push(format!("Kinds: {}", kind_parts.join(", ")));
 
     lines.push(String::new());
-    if let Some(reason) = &embedding_runtime.vector_index_discarded {
+    if let Some(reason) = embedding_runtime
+        .vector_index_discarded
+        .as_ref()
+        .filter(|_| embed_status.pending > 0)
+    {
         // A real gap, reported with its cause. The bare counters here read as
         // discovered loss; naming the open-time discard and the automatic
-        // recovery is what tells the operator no manual GPU pass is owed.
+        // recovery is what tells the operator no manual GPU pass is owed. The
+        // discard reason is recorded once at open and never cleared, so it is
+        // named only while the gap it explains is still open; a recovered
+        // store goes back to the plain measured line.
         lines.push(format!(
             "Embeddings: {}/{} indexed ({} pending); the persisted vector index was not loaded \
              when this daemon opened ({reason}); the daemon restores coverage in the background",
@@ -1715,6 +1722,39 @@ mod tests {
                 .any(|line| line.contains("embeddings are still pending")),
             "a real gap keeps its warning: {:?}",
             discarded.lines
+        );
+
+        // The reason is recorded once at open and never cleared, so once the
+        // gap it explains is closed the line must not keep naming it. A graph
+        // with nothing pending renders the plain measured line.
+        let (_temp2, binding2, empty_graph) = graph_validation_fixture();
+        let recovered = build_graph_status_response(
+            &binding2,
+            &empty_graph,
+            &Default::default(),
+            &GraphStatusEmbeddingRuntime {
+                vector_index_discarded: Some(
+                    "fixture: metadata no longer matches graph truth".to_string(),
+                ),
+                coverage_ever_complete: true,
+            },
+        )
+        .unwrap();
+        assert!(
+            !recovered
+                .lines
+                .iter()
+                .any(|line| line.contains("was not loaded")),
+            "a closed gap must not keep naming its discard: {:?}",
+            recovered.lines
+        );
+        assert!(
+            recovered
+                .lines
+                .iter()
+                .any(|line| line == "Embeddings: 0/0 indexed (0 pending)"),
+            "a store with nothing pending renders the plain measured line: {:?}",
+            recovered.lines
         );
     }
 
--- a/crates/kin-cli/src/commands/graph.rs
+++ b/crates/kin-cli/src/commands/graph.rs
@@ -10,7 +10,7 @@ use kin_model::{
 };
 use serde::{Deserialize, Serialize};
 
-use super::graph_health::inspect_graph;
+use super::graph_health::{inspect_graph, inspect_graph_with_embedding_observation};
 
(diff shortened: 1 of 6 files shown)


troyjr4103:
Merge-queue removal recorded by lane wave6-captain, so this PR carries its own history.

This PR was removed from the hosted merge queue at 2026-08-13T21:04:36Z (`removed_from_merge_queue`). Nothing landed on the PR when it happened, and it still reads OPEN, MERGEABLE, CLEAN with 25 SUCCESS, 3 SKIPPED and zero failures, which is exactly what a healthy PR waiting its turn looks like.

**This PR did not fail anything.** Its own speculative merge-group run, 31740696072, completed SUCCESS. It was collateral from the entry ahead of it: kin#809's batch (run 31740694061) failed on the containment-guardian teardown flake, and when that entry was evicted the batch riding on top of it was torn down with it.

Disposition: needs re-admission. No code change is called for and no re-review is needed. It should be re-submitted once kin#803 lands, since 803 fixes the flake that started this cascade.

troyjr4103:
Ejected from the merge queue at 21:01:55Z, and nothing marked this PR at the time.

The `merge_group` run for batch `gh-readonly-queue/main/pr-804-d1673ca7...` failed on
job `Check & Test (macos-latest)`, at the `Clippy` step:

- run: https://github.com/firelock-ai/kin/actions/runs/31743756292
- job: https://github.com/firelock-ai/kin/actions/runs/31743756292/job/94593257507

That run belongs to the merge group rather than to this pull request, so the ejection left
no trace here. The PR's own checks stayed green and it simply stopped moving, which is what
a healthy PR waiting its turn looks like. Worth knowing before re-submitting: the rest of
that cohort (811, 810, 803, 815) went on to merge, so main has moved and this branch is now
well behind.

I could not read the clippy diagnostic itself. The job log endpoint returned only the
cleanup tail, so the specific lint is not stated here on purpose rather than guessed. Open
the job link for it.

Recorded by the readpath-captain lane while landing kin#813 and kin#814, so this PR carries
its own history.

troyjr4103:
Ejected from the merge queue twice this evening, for two different reasons.
Recording both here, because an ejection leaves no mark on the pull request by
design and nothing above this comment shows either one happened.

The first was collateral. Merge group `pr-804-d1673ca7` at 2026-08-13T20:24:23Z
ran CI 31740696072 and that run SUCCEEDED, along with SAST and PR Text Hygiene.
This PR was still dropped, because the batch it shared was failed by the
containment guardian teardown flake that hit #809 two seconds earlier. #803
(338e2342) fixed that flake on main at 21:44:33Z.

The second was its own. Merge group `pr-804-d1673ca7` at 21:01:55Z, CI run
31743756292, failed in macOS Clippy before compiling anything:

```
error: failed to get `chrono` as a dependency of package `kin-core v0.5.24`
```

That is a registry fetch failure on the runner, not a defect in this branch and
not the containment flake. It carries no fix and needs none.

Current readiness, parsed on check states rather than exit codes, since
`gh pr checks --json` always exits zero: head 772b9d5b reads SUCCESS on 24 and
SKIPPED on 3, with no failures and nothing pending. Mergeable, not conflicting.

No rebase is needed. The hosted queue builds its speculative merge on the
current main tip, and 338e2342 is an ancestor of that tip, so the next merge
group this PR enters already carries the flake fix.

Not enqueuing. This PR already holds exactly one row in the kin ledger, and
`merge enqueue` appends with no dedup, so a second row could never reach a
terminal state.

PR #809, opened 2026-08-13
Give the compat-v0 locate arm the shared file surface

troyjr4103:
The default retrieval profile is entity-centric. It demotes the file to `provenance.file` on each hit and publishes no `files[]` of its own, so a caller asking which files a query is about had to page the entity ranking and aggregate the paths itself. The fused arm answers that in one field. The A/B recorded on FIR-2137 put the correct file in accuracy-v1's top five on 18 of 18 graded queries while compat-v0 returned no file surface on 100 of 100 calls, and compat-v0 is what nearly every caller runs.

This rolls the retained cosine ranking up into the same `LocateFileEntry` shape `kin locate --json`, `POST /locate`, and the fused arm already emit, so one consumer parses every Kin locate surface. Files appear in the order their best hit ranked and carry that hit's score, because the ranking arrives rank-ordered. `signals` is `embeddings` alone, which is the literal provenance: every candidate came from the vector index. The explain-only fields stay unset, so both arms' plain file surfaces carry the same field set.

Two things this deliberately does not do. It adds no retrieval, so it costs nothing per call and cannot rank a file whose entities did not rank; the fused arm can do that only because it runs a lexical and path pool this profile does not. And it does not touch the default, whose flip stays refused on the measured evidence recorded on the ticket.

The surface is derived from the whole retained ranking rather than the returned page, the same rule `all_fallback` follows, since which files a query is about is a property of the ranking and must not change as the caller pages through it. It serializes unconditionally, so an empty ranking says `[]` instead of omitting the key and making absence read like emptiness.

Three tests, run and named in the gate log. The page-invariance one was falsified against the obvious wrong implementation: switching the rollup to the page window drops it to `["src/daemon.rs"]` against the expected three files, so it fails when the claim is false. The third test pins an interaction worth stating, because the two payload shapes now look alike enough that a later tidy-up could merge them: the negative envelope must keep counting `results` alone on this arm. Its `files[]` is a rollup of those same rows, so counting both the way the fused branch does would double-count, and the fused branch counts both only because its two surfaces are independently ranked.

Gate: `cargo nextest run --workspace` plus `cargo test --doc`, green at 5608 passed on this worktree.

Closes FIR-2137

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -6772,6 +6772,13 @@ fn semantic_locate_payload(
         "next_cursor": next_cursor,
         "results": window,
     });
+    // The file surface, in the same `files[]` shape `kin locate --json`, `POST
+    // /locate`, and the fused arm already emit, so one consumer parses every Kin
+    // locate surface. Serialized unconditionally, including as `[]` for an empty
+    // ranking, because the fused arm serializes it unconditionally too and a key
+    // that appears only when non-empty makes absence and emptiness the same
+    // reading.
+    payload["files"] = json!(cosine_file_surface(rows));
     if !degradations.is_empty() {
         payload["degradations"] = json!(degradations);
     }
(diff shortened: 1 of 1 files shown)


troyjr4103:
Merge-queue ejection recorded by lane wave6-captain, so this PR carries its own history rather than only the merge group's.

This PR was ejected from the hosted merge queue at 2026-08-13T20:41Z. Nothing landed on the PR when it happened, which is the documented shape: the failing run belongs to the merge group, not to any pull request, so 809 kept reading OPEN, MERGEABLE, CLEAN with 25 SUCCESS, 3 SKIPPED and zero failures while its queue entry read UNMERGEABLE.

Ejecting run: 31740694061, branch `gh-readonly-queue/main/pr-809-d1673ca787c9afa872ea39e009225c9e0033f9ee`. Failing jobs: Check \& Test (ubuntu-latest) at the Test step, and Check \& Test (macos-latest) cancelled behind it by fail-fast, so there is one real failure here and not two.

The failure is not a defect in this PR. One test died in harness teardown rather than on an assertion:

```
contextbench_locate_keeps_query_selection_and_normalization_inside_kin
panicked at crates/kin-cli/tests/common/mod.rs:1342:17
containment termination=ok; quiescence=inspect test process containment:
process group 28740 still holds live process 28869 after sentinel reap
```

`termination=ok` says the test's own work finished. What failed is the containment guardian proving quiescence when a process group still held a live child at reap. That is the same flake kin#803 fixes, and 803 was sitting behind this PR in the queue when this ran.

Disposition: holding 809 until kin#803 lands, then re-admitting it. Re-submitting first would only re-roll the same race.

troyjr4103:
Ejected from the merge queue at 20:24 UTC, and nothing marked this PR, so it still reads CLEAN and MERGEABLE with its own checks green.

The merge group run is 31740694061, and the failing job is Check & Test (ubuntu-latest): https://github.com/firelock-ai/kin/actions/runs/31740694061/job/94583121204

It failed on `kin-cli::contextbench_locate_json contextbench_locate_keeps_query_selection_and_normalization_inside_kin`. The panic is a containment quiescence assertion at `crates/kin-cli/tests/common/mod.rs:1342`, reporting that process group 28740 still held live process 28869 after the sentinel reap. It is not an assertion about locate output. The run stopped at 1474 of 5573 tests under fail-fast, so 4099 tests never ran.

The queue moved on without this entry; PR 810 entered a merge group at 20:41.

Recording it here because a `merge_group` failure belongs to the group rather than to any pull request, so with no note the only trace is an Actions run naming no PR, and a captain watching PR-level state sees a healthy PR that simply never merges.

Not claiming a cause. Whether this is the branch's own change or the known containment flake needs a re-run to tell.

troyjr4103:
Ejected from the merge queue earlier this evening, and recording it here so this
PR carries its own history. An ejection leaves no mark on the pull request by
design, so nothing above this comment shows it happened.

The merge group was `gh-readonly-queue/main/pr-809-d1673ca7`, CI run
31740694061, created 2026-08-13T20:24:21Z. It failed on the containment
guardian teardown check:

```
contextbench_locate_keeps_query_selection_and_normalization_inside_kin
panicked at crates/kin-cli/tests/common/mod.rs:1342:17
containment termination=ok; quiescence=inspect test process containment:
process group 28740 still holds live process 28869 after sentinel reap
```

That is the flake, not this branch. #803 ("Settle dying group members and
reserve the tests' closed ports", 338e2342) fixed it on main at 21:44:33Z, in
`crates/kin-cli/src/daemon_client.rs` and `crates/kin-daemon-spawn/src/lib.rs`.

Current readiness, parsed on check states rather than exit codes, since
`gh pr checks --json` always exits zero: head e4e0113c reads SUCCESS on 25 and
SKIPPED on 3, with no failures and nothing pending. Mergeable, not conflicting.

No rebase is needed to pick the fix up. The hosted queue builds its speculative
merge on the current main tip, and 338e2342 is an ancestor of that tip, so the
next merge group this PR enters already carries the fix. Force-pushing a rebase
would only discard 25 green checks to re-earn the same result.

Not enqueuing. This PR already holds exactly one row in the kin ledger, and
`merge enqueue` appends with no dedup, so a second row could never reach a
terminal state.

PR #812, opened 2026-08-13
Run the install proof's capability contract against main instead of only at a tag

troyjr4103:
The Public Install Proof is the best gate Kin has and it had the worst feedback
loop. It installs from the public endpoint with no checkout and no secrets,
which is what makes it worth trusting, and it therefore runs only once a tag
exists. A change that breaks it lands green, ships, and fences the release it
shipped in. That is the v0.5.18 story: a readiness classifier started reporting
pending on a host that could never make progress, the proof caught it correctly
at tag time, and the tag could not be repaired, because a tag run resolves its
workflows from the tag. Prevention has to fire before a tag exists.

A canary now builds Kin from the branch, installs it into the same layout the
public installer uses, drives a real first run, and holds the result to the
portable half of the proof's capability contract. It runs nightly against main
and path-filtered on the proof's known inputs, which are the proof itself, the
health and readiness classifiers, status, resources, embed, and init. It
publishes nothing and promotes nothing, and it makes no claim about a release,
because a branch binary is not a release binary. Release provenance stays in the
proof, where it belongs.

The contract it enforces is the part that broke. A report's aggregate must agree
with its own checks by the proof's own rule. Every check id the proof requires
must be present, so a rename fails here rather than at a tag. And readiness is
judged against the coverage counters rather than against its own word for
itself: a settled store must report healthy, a store with nothing to embed must
not report pending, and an unobserved coverage settles nothing either way. That
last tolerance is correct and it is also the one way this could pass while
proving nothing, so the run after the embedding pass asserts under
--require-observed and fails a run that judged readiness against nothing.

The proof cannot read a tracked script, so the contract exists twice and the
release authority suite compares the copies. Two copies drift in the direction
that matters most here, where the canary stays green while the proof it was
built to predict starts failing at a tag. Six falsifications drive that link: a
check the proof requires and the canary drops, a path filter that loses the
classifier which fenced v0.5.18, a canary that can pass without ever observing
coverage, one that grows the authority to publish, one that stops running the
contract at all, and one that keeps only its path filter and loses the nightly
run that covers the paths nobody anticipated.

Closes FIR-2221

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,12 +110,14 @@ jobs:
             ./scripts/release-archive-shape.test.cjs \
             ./scripts/release-order.test.mjs \
             ./scripts/release-hold-alarm.test.mjs \
+            ./scripts/verify-capability-proof.test.mjs \
             ./scripts/verify-npm-attestation.test.mjs
           node --check ./scripts/read-update-build-identity.cjs
           node --check ./scripts/check-release-version.mjs
           node --check ./scripts/prepare-release.mjs
           node --check ./scripts/release-order.mjs
           node --check ./scripts/release-hold-alarm.mjs
+          node --check ./scripts/verify-capability-proof.mjs
           node --check ./scripts/verify-npm-attestation.mjs
           bash -n ./scripts/publish-npm-release.sh
           bash -n ./scripts/verify-npm-release.sh
--- a/.github/workflows/install-proof-canary.yml
+++ b/.github/workflows/install-proof-canary.yml
@@ -1,0 +1,201 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 Firelock, LLC
+
+name: Install Proof Canary
+
+# The Public Install Proof runs only once a tag exists, because it installs from
+# get.kinlab.dev with no checkout and no secrets, which is exactly what makes it
+# worth trusting. The cost of that is a feedback loop one release long: a change
+# that breaks the proof lands green, ships, and fences the release it shipped in.
+# That is the v0.5.18 story. The classifier change landed with green CI, the
+# proof caught it correctly at tag time, and the tag could not be repaired,
+# because a tag run resolves its workflows from the tag. Nothing on main can
+# reach an existing tag's run, so prevention has to fire before a tag exists.
+#
+# This is that prevention. It builds Kin from the branch, installs it into the
+# same layout the public installer uses, drives a real first run, and holds the
+# result to the portable half of the proof's capability contract. It publishes
+# nothing, promotes nothing, and proves nothing about a release: a branch binary
+# is not a release binary, and the release provenance assertions stay where they
+# belong, in the proof itself. What moves here is the part that broke.
+#
+# It is deliberately not on every pull request. The paths below are the proof's
+# known inputs, so a change that cannot affect the proof does not pay for a Rust
+# build, and the nightly run covers anything the path list fails to anticipate.
+on:
+  schedule:
+    # Nightly, so a break that arrives through a path this filter does not name
+    # is still found within hours instead of at the next cut.
+    - cron: "0 7 * * *"
+  pull_request:
+    branches: [main]
+    paths:
+      - ".github/workflows/install-proof.yml"
+      - ".github/workflows/install-proof-canary.yml"
+      - "scripts/verify-capability-proof.mjs"
+      - "scripts/verify-capability-proof.test.mjs"
+      - "crates/kin-cli/src/commands/health.rs"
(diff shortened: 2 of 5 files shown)


troyjr4103:
The red on this PR is not this PR's defect. It is the containment-guardian flake that #803 fixes, and the evidence says so three ways.

The failure is a single Rust test, `kin-cli::repository_branches branch_delete_rejects_default_and_checked_out_ref_without_mutation`, panicking in the test harness rather than in an assertion (run 31731917605, job 94554258362):

```
containment termination=ok; quiescence=inspect test process containment: process group 36337
still holds live process 36431 after sentinel reap; callers must kill and wait owned direct
children before guardian finalization; reap=runtime containment guardian retained because
quiescence was not proven
```

That string is emitted from `crates/kin-daemon-spawn/src/lib.rs:2034`, the `LiveMember` arm of the single post-reap probe in `finalize_owned_process_group`. This PR touches no Rust at all. Its diff is one workflow, one new workflow, one Python test, and two Node scripts.

kin#802, an unrelated PR from a different lane, failed in the same window with the byte-identical panic at the same source line, on a different victim test (`concurrent_resolutions_from_one_view_leave_one_winner`, run 31729654610, job 94546737432). Two PRs, two unlucky victims, one guardian. That is the random-victim shape #803 describes, where the barrier's SIGKILL is delivered but the kernel has not yet retired the member, and the one post-reap probe judges a process mid-death as a live escape.

#803 adds `settle_pinned_group_member_exits()` and calls it inside `finalize_owned_process_group` before the sentinel reap, so the group settles before anything judges it. That is exactly the code path failing here.

This PR's own content is green on this same sha. The npm launcher tests job, which runs the `verify-capability-proof.test.mjs` and `node --check` lines this PR adds, is SUCCESS. The `test-release-workflow-authority.py` step this PR modifies ran inside the failing job and passed before the Rust suite started. Locally the new Node suite is 18/18.

So nothing here should change to make CI green, and weakening the guard would defeat the point of the PR. The ordering is the fix: #803 is CLEAN, MERGEABLE, and carries zero failing checks, but it currently sits behind both #802 and #812 in the kin queue. Landing #803 first makes this PR's run deterministic instead of a lottery, and does the same for #802.

I have re-run the failed job here. If it comes back green the PR is admissible on its own, but the queue order above is still the right call for the wave.

troyjr4103:
Re-run confirms the diagnosis. Run 31731917605 came back `success` with no change to the PR: 26 SUCCESS, 3 SKIPPED, zero failures, and the macOS leg that had read CANCELLED is green on its own. The ubuntu leg cleared the same nextest step it died in last time, so the guardian flake simply did not recur.

This PR is ready to submit as it stands.

The ordering note still holds for the wave, just not for this PR. kin#802 is red from the identical guardian panic and #803 is the fix for it, so landing #803 early in the next drain stops the flake rather than dodging it one re-run at a time.

troyjr4103:
Dropped from the merge queue, and nothing marked this PR at the time.

This one is not a CI failure. The entry sat `UNMERGEABLE` in the queue for roughly
45 minutes and was then dropped without ever being built: there are no `merge_group` runs
for a `pr-812` batch at all. The PR now reads `mergeable=CONFLICTING`, `mergeState=DIRTY`,
so it conflicts with main and needs the conflict resolved before it can enter the queue
again.

Main moved a long way while it sat there. 811, 810, 803 and 815 merged, then 813, so the
base this branch was written against is several squashes old.

Recorded by the readpath-captain lane while landing kin#813 and kin#814, so this PR carries
its own history rather than only reading green and never merging.

troyjr4103:
Conflict-dequeued and rebased.

This PR read CONFLICTING across two consecutive polls, so it was a genuine
conflict-dequeue rather than patience. It is now rebased from 94873525 onto
1fe13fe7, with base 1255772c ("Recognize a bare symbol name against the
qualified entity that declares it", #813).

The conflict was two independent additions landing at the same insertion
points, not a disagreement. #810 added the release hold alarm to the node test
list and the `node --check` list in `ci.yml`, and added its constants and its
contract function to `scripts/test-release-workflow-authority.py`. This branch
adds the capability-proof verifier at those same four places. Both belong, so
the resolution keeps both. Taking either side whole would have silently dropped
a gate from CI: main's side would have lost the capability contract this PR
exists to add, and this branch's side would have un-gated the hold alarm that
#810 just landed.

Verified after the rebase, against main's updated `install-proof.yml` rather
than the one this branch was written against:

- `scripts/test-release-workflow-authority.py` green, both contracts asserted
- `node --test` 36/36 across the capability-proof and hold-alarm suites
- full repo gate green, 5627/5627 tests, no tracked lock contamination

Not claiming anything about the merge queue run, which has not happened yet.

PR #814, opened 2026-08-13
Read the locate honesty verdict from the answer rather than the page size

troyjr4103:
Asking for more results removed the honesty envelope. On the default locate arm
a fabricated symbol at limit 8 came back with one ranked row and the
`no_named_match` disclosure; the same symbol at the same daemon seconds later at
limit 50 came back with fifty confident scored rows and no disclosure at all.

`locate_ranking_names_nothing` read the two arms differently. The fused arm used
the payload's own `all_fallback`, computed over the full ranking before paging.
The default cosine arm inferred the same fact from every row's
`match_evidence.name_match`, and only when `total_ranked <= rows.len()`, so it
refused to guess about rows it could not see. That refusal was right. What made
it a guard that could not fire is that `total_ranked` is not a fixed property of
a query. It grows with the requested limit, measured at 1 to 56 and 15 to 107
for two fabricated symbols on a fresh fully embedded store, so the page covers
the ranking only when the ranking collapses to a degenerate row or two. The
disclosure was present in the one case nobody queries and absent in every case
they do, which is a check that cannot fail pointed at the product's own honesty
layer.

kin#786 removed the reason for the asymmetry. The cosine arm now publishes
`all_fallback` itself, computed over the whole retained `rows` with exactly the
fused rule, before the daemon windows a page. Both arms omit the field when it
is false, so its presence is authoritative and routing does not enter into it.

The verdict now reads that published field whatever the arm, and keeps the
`match_evidence` inference only for a payload that carries no such field, which
is an older daemon. Disclosure becomes a property of the answer instead of a
property of how many rows the caller happened to ask for.

## Ordering

kin#786 and kin#791 both merged on 2026-08-12, so the two constraints the ticket
named are met. kin#791 landing first is what keeps the widened verdict
inconclusive by construction, saying the ranking named nothing and never that
the graph holds nothing, which is the safe direction the ticket asked for.

kin#813 is a third constraint the ticket did not name, and it is the one that
decides whether this widened verdict is honest or noise. On main without it,
`query_names_entity` takes only the last DOTTED segment and cannot see through
`::`. The parsers store every Rust and C++ method as `Type::method`, so no row
of a method query is labelled a name match, so `locate_rows_are_all_fallback` is
true, so the daemon publishes `all_fallback` even on a page whose top row IS the
symbol. This change makes the negative envelope act on that field at any limit,
and its admission gate fires on any token of three characters or more containing
an underscore, which is the snake_case shape a developer types. So main carrying
this change without kin#813 would tell an agent that no ranked entity carries
the name it asked for, on a page that does contain it, for essentially every
method query against a Rust or C++ graph. That converts a silent miss into a
confident false statement on the honesty layer itself, which is the failure the
0.5.19 dogfood called a regression in kind. kin#813 is what makes the published
field true only when it should be.

So this lands behind kin#813, either with kin#813 on main first or in the same
merge-group batch with kin#813 ahead. The two change different files,
`crates/kin-cli/src/commands/locate.rs` against `crates/kin-mcp/src/negative.rs`,
so batching them is mechanically fine. If kin#813 is ever reverted or ejected,
this has to go with it.

## Evidence

Three tests. The decisive one is the measured pair as one query at two limits
against one daemon, and it goes red on the unfixed code, verified by reverting
the branch locally and rerunning:

```
a_cosine_page_wider_than_its_window_still_carries_the_unnamed_verdict
  panicked: a wider page is still a ranking that named nothing
```

The other two are the controls that keep the widening honest rather than
proving it: a ranking holding a name hit is never qualified, on a windowed page
as well as a whole one, and a payload without `all_fallback` keeps today's
inferred verdict exactly, so an older daemon does not change answer. Both pass
before and after, which is what a regression guard is supposed to do.

Gate: `kin-dev local-test kin` on this worktree at 676386b1, 5608 tests run,
5608 passed, tracked lock restored with no contamination. CI-equivalent
`cargo clippy --all-targets --all-features` with the ci.yml allowlist exits 0
with no new warnings. `cargo fmt --all --check` clean.

## Not claiming

No real-store rerun of the limit-8 against limit-50 pair. The payload shapes
under test are transcribed from the recorded wire measurements rather than
captured fresh.

Closes FIR-2262

### DIFF
--- a/crates/kin-mcp/src/negative.rs
+++ b/crates/kin-mcp/src/negative.rs
@@ -230,17 +230,31 @@ fn query_names_a_symbol(query: &str) -> bool {
 /// scores cannot separate them: they are ranks within a result set, not evidence
 /// that anything matched.
 ///
-/// The verdict is read from what each arm already publishes about its FULL
-/// ranking, never from the page alone, because a page is a window: an exact name
-/// match sitting on page two would otherwise be reported as absent from the
-/// whole ranking.
+/// The verdict is read from what the answer publishes about its FULL ranking,
+/// never from the page alone, because a page is a window: an exact name match
+/// sitting on page two would otherwise be reported as absent from the whole
+/// ranking.
 ///
-/// - Fused arm: `all_fallback` is computed over the full ranking before paging
-///   and means precisely this, so it is used verbatim.
-/// - Cosine arm: every row's `match_evidence.name_match` is inspected, and only
-///   when the page holds the entire ranking. A row that reports no
-///   `match_evidence` at all leaves the question unanswered, so nothing is
-///   qualified — the same refusal to guess the rest of the module makes.
+/// `all_fallback` IS that published verdict. Both arms compute it over the whole
+/// retained ranking before the daemon windows a page, with the same rule, and
+/// both omit the field when it is false, so its presence is authoritative and
+/// routing does not enter into it.
+///
+/// A payload carrying no `all_fallback` at all is an older daemon that never
+/// published one, and only there is the fact inferred from the page: every row's
+/// `match_evidence.name_match` is inspected, and only when the page holds the
+/// entire ranking. A row that reports no `match_evidence` leaves the question
+/// unanswered, so nothing is qualified, which is the same refusal to guess the
+/// rest of the module makes.
+///
+/// The inference cannot be the primary reading, because the condition it needs
+/// is one a real store almost never meets. `total_ranked` grows with the
+/// requested limit, so the page covers the ranking only when the ranking
+/// collapses to a degenerate row or two. A fabricated symbol measured at limit 8
+/// ranked 1 and carried the disclosure; the same symbol at limit 50 ranked 56
+/// and carried none. Asking for more results removed the honesty envelope, which
+/// left the guard present in the one case nobody queries and absent in every
+/// case they do.
 fn locate_ranking_names_nothing(payload: &Value) -> bool {
     if !payload
         .get("query")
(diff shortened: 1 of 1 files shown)


troyjr4103:
Sequencing for whoever captains this, recorded here so it cannot be lost.

This must not land without kin#813. On main as it stands, `query_names_entity`
reads only the last dotted segment and cannot see through `::`. The parsers
store every Rust and C++ method as `Type::method`, so no row of a method query
is labelled a name match, `locate_rows_are_all_fallback` is true, and the daemon
publishes `all_fallback` even on a page whose top row IS the symbol. This change
makes the negative envelope act on that field at any limit, and its admission
gate fires on any token of three characters or more containing an underscore, so
it would fire on essentially every method query a developer types against a Rust
or C++ graph and tell the agent that no ranked entity carries the name, on a
page that does contain it. kin#813 is what makes that field true only when it
should be. The body now carries the full mechanism.

So submit kin#813 first, either landing it on main before this one or batching
both into one merge group with kin#813 ahead. The two change different files,
`crates/kin-cli/src/commands/locate.rs` against `crates/kin-mcp/src/negative.rs`,
so batching is mechanically fine. The kin-lane queue already holds kin#813 ahead
of this entry, so draining in order satisfies it. If kin#813 is ever reverted or
ejected from the queue, this has to go with it.

This is a PR comment rather than a note on the queue row because there is no way
to annotate an existing row without the captain's lease or a duplicate.
`merge readmit` calls `require_merge_authority` and also refuses an entry that
carries no deferral to clear, `merge defer` would move this entry to the tail
and requires it to be at the head, and `merge enqueue` is a bare append with no
dedup, so re-enqueueing would have created a second row for the same PR. This
lane holds no lease and changed no queue state.

One refinement after reading `bin/kin-lane` rather than trusting its help text.
`merge drain` describes itself as submitting "every green non-deferred entry in
order", but its only gate on an entry is `state == "OPEN"`; there is no check on
CI at all. A grep of the drain block finds the state read and finds no
`statusCheckRollup`, `mergeStateStatus`, or checks query, and the state read is
the positive control proving the search works. So a drain run while kin#813 is
still building will arm auto-merge on both PRs regardless.

Arming order is not entry order either. GitHub enters a PR into the merge queue
when that PR becomes eligible, not when auto-merge was armed, so if this one is
green while kin#813 is still building, this one can enter the queue first and
land alone. That is the exact outcome the dependency above rules out.

So the robust order is kin#813 landed on main first, and this one submitted
after. If both are submitted together instead, both must already be green at
submit time so eligibility cannot reorder them.

PR #842, opened 2026-08-15
Answer MCP handshake frames before the daemon binds; park a supervisor-stopped reconcile loop

troyjr4103:
An agent calling `semantic_locate` now gets what a human typing `kin locate` has always gotten. The MCP tool's default route is the fused multi-signal pipeline on every profile, the same ranking `POST /locate` serves, and the legacy cosine ranking stays reachable per call via `pipeline:"cosine"` for A/B comparison. On the 0525 battery's 23-symbol probe set the cosine default scored 0/23 at rank one where the fused pipeline scored 20/23 (non-citable dev-box measurement, `.kin-coord/reports/profile-default-decision-20260814.md`).

The shipped default profile flips from compat-v0 to accuracy-v2, a new version minted per the profile module's own rule that lever flips land as a new version rather than changing what an existing pin means. accuracy-v2 is accuracy-v1's lever set (entity fusion, lexical parity floor, calibrated 0.625 embedding seed floor, promotion-only rerank blend, 1500ms rerank budget) with the cross-encoder reranker off unconditionally. That implements the founder ruling within the module's law: the accuracy levers graduate to default, and the reranker's do-not-flip verdict is affirmed (one lateral top-1 change in 55 probes for +54% latency and +1.3GB resident). Returning false explicitly, rather than keeping accuracy-v1's cached-model condition, also protects installs that prefetched the model under earlier doctor advice from silently starting to pay for it. An accuracy-v1 pin keeps the conditional reranker it always had, and compat-v0 keeps the historical lever defaults as the A/B escape hatch.

`kin doctor` now agrees with the shipped default. A stock install reports Healthy with no remediation; the advice to set accuracy-v1 and prefetch the 1.1GB reranker model is gone; a lever-off profile someone chose still warns and points back at accuracy-v2. KIN_PROFILE stays env-only, an exported profile still wins, and the new default lands when a daemon restarts under the new binary.

Tests pin all three decisions: the default variant with every lever value including reranker-off under the cached-model-plus-resident-daemon condition, fused routing under an explicit compat-v0 environment, and doctor emitting no advice that contradicts the shipped default even with the reranker model cached.

Closes FIR-2321

### DIFF
--- a/crates/kin-cli/src/commands/mcp.rs
+++ b/crates/kin-cli/src/commands/mcp.rs
@@ -20,6 +20,16 @@ use std::pin::Pin;
 /// server starts anyway and each `tools/call` fails loud with a structured,
 /// actionable error (see `kin_mcp::daemon_delegate::daemon_unavailable_tool_result`)
 /// instead of the process silently never having started at all.
+///
+/// Answer early, load behind (FIR-2316): the stdio loop starts before the
+/// daemon binding rather than after it. Resolving the repo daemon can spawn
+/// one, and a cold start on a flagship-scale store takes minutes; awaiting it
+/// here meant zero stdout frames for that whole window, without even the
+/// `initialize` response, which no handshake probe or client timeout survives.
+/// The binding now runs on a background task that publishes into a
+/// [`kin_mcp::StartupDaemonBinding`]; `initialize` and `tools/list` answer
+/// immediately, and a `tools/call` that arrives before the binding settles is
+/// answered honestly that the daemon is still starting.
 pub async fn start(
     global: bool,
     repo: Option<PathBuf>,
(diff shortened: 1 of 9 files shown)


troyjr4103:
Provenance correction for history auditors: this PR's squash commit body (7c933f7f) is a verbatim copy of #841's and describes the profile flip, which landed separately as 1f5f9924. The actual change here is what the diff shows: the MCP stdio serving path answers initialize and tools/list before the daemon binds, tools/call returns a typed daemon-still-starting result with the phase named after a bounded grace, and a supervisor-stopped reconcile pass parks the loop instead of exiting the daemon (FIR-2316, FIR-2317). The body was wrong at submit time and the merge queue minted it verbatim; reading the diff, not the message, gives the truth for this commit.

PR #863, opened 2026-08-17
Pin kin-db 0.7.31

troyjr4103:
Carries two kin-db fixes into the kin binary. kin-db#192 judges a transaction by the exclusion rules already in force rather than the ones it publishes, which closes the livelock where admitting a new .gitignore beside newly-excluded untracked content failed wholesale on every retry (FIR-2348). kin-db#193 releases the vector index slot before saving the sidecar, which is the layer-1 half of the stall that was misattributed to graph re-derivation: the commit was waiting on a lock held by a full HNSW rebuild, not rebuilding anything itself (FIR-2351; the kin-vector half is FIR-2367).

The pin moves =0.7.29 to =0.7.31 with the registry source and checksum preserved; exactly one package moved in the lock.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3265,9 +3265,9 @@ dependencies = [
 
 [[package]]
 name = "kin-db"
-version = "0.7.29"
+version = "0.7.31"
 source = "sparse+https://kinlab.ai/registry/cargo/"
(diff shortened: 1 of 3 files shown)


troyjr4103:

PR #866, opened 2026-08-17
Read one reference collector for dead code and gate the delete list on measured edge coverage

troyjr4103:
`kin dead-code` and `find_references` answered opposite things about the same
four entities at the same graph generation, and the scan's rule never looked at
relation kind at all. `find_dead_code` asks only whether an inbound edge's source
lives in a DIFFERENT FILE (kin-db `traverse.rs` `has_cross_file_incoming`), while
`find_references` and `kin refs` keep any inbound Calls, Imports or References
edge regardless of file (`refs.rs` `collect_graph_references`,
`handlers/common.rs` `collect_graph_reference_rows`). So an entity whose only
caller sat beside it in the same file was reported unreferenced while
`find_references` named that caller. On a fresh install whose graph held
intra-file calls and nothing else, that printed 36 of 46 live source entities as
a delete list, including the declared console entry point, with no caveat.

The scan now consults `collect_graph_references`, so one collector answers for
`kin refs`, `kin refs --bulk-json`, the seeded scan and the whole-repo scan. A
test asserts the two agree entity by entity on a fixture carrying intra-file
callers, driven through the released bulk-refs builder rather than an internal
helper. The MCP whole-repo path reads the same reference kinds, which is what its
own tool description already claimed.

Entry points are never listed. A console script a project manifest declares and a
`main` are invoked from outside the graph, so no inbound edge was ever owed. The
declarations are read from the manifests the graph tracks as structured
artifacts, with bytes taken from graph-owned blob storage through the repository
authority. Nothing reads the working tree. A manifest the graph tracks but this
scan could not read is reported as unread rather than as declaring nothing.

The output is a delete list, so it now carries the strictest completeness gate in
the product, and the gate rests on a measurement that did not exist before.
Extraction already knew how many call sites and import statements it read in each
file and dropped both, so no surface could say that a graph held a fifth of its
reference edges: `graph validate` reports integrity, `graph status` reported
density, and both passed. Those counts are now recorded on every entity of the
file and read back per language against the edges the graph resolved. A file
whose parser recovery omitted calls records no call-site count rather than one it
cannot stand behind, so a reader tells unmeasured from zero, and the verdict keys
on call sites alone because a local import statement resolves to an artifact-level
edge no entity-rooted query reaches.

When a language holds no cross-file reference edge across several files whose
parser read import statements, or parsed call sites that resolved to no edge at
all, or when a tracked manifest could not be read, every affected row is labelled
unverified with the measured reason and the confident phrasing is withheld. A row
is labelled on the state of its own language, so a repository whose Rust edges
resolved does not have every Rust row labelled because two unresolved TypeScript
files sit beside them.

`graph validate` now says it checks integrity only and prints the completeness
metric beside its verdict, because "All checks passed" was read as a clean bill.
`graph status`, the structured health report, the graph command response and a new
`kin doctor` row carry the same metric. The doctor row needs attention without
failing readiness, since the graph is still telling the truth about the edges it
does hold.

Every new assertion was falsified by reverting the fix it pins: dropping the
shared-collector consult reproduces the original four-entity contradiction,
dropping the entry-point exclusion lists `main`, and forcing the gate open prints
a confident list on a graph with no cross-file edges.

`kin_core::reference_coverage` overlaps kin PR #861's `cross_file_coverage` on
the resolved cross-file count. Neither is wrong and both stay landable in either
order. FIR-2370 carries the fold to one type, one status section and one doctor
row. FIR-2371 carries the file-scoped MCP path, which keeps its documented
ignore-within-file-references semantics.

Closes FIR-2356
Closes FIR-2358


troyjr4103:

PR #867, opened 2026-08-17
Bound trace_data_flow's own response, order its fan-out by relevance, and give every step one shape

troyjr4103:
`trace_data_flow` returned a result its own client refused. Measured on a converted
`psf/requests` (777 entities), `depth: 4, limit_per_step: 8` produced 228,413 characters over
2,453 lines and was rejected for exceeding the token limit, because all 106 steps inlined a
full source body and nothing measured the payload. The caller recovered by parsing a spill
file with a script, in a tool whose stated purpose is avoiding round trips. Four defects, all
in one answer.

**The result did not fit, and there was no way to ask for less.** `include_body: false` now
returns the SHAPE of a chain: names, kinds, roles, spans, signatures, and every edge, with no
body anywhere including the focal's. On the MCP surface `compact: true` means the same thing,
because an agent that learned that spelling from `bulk_check_references` on this same surface
should not be handed a 228k-character response for using it; an explicit `include_body` wins
over the alias so the two can never disagree silently. The CLI carries `--no-bodies`.

The default stays `include_body: true` on purpose. The budget below is what makes that safe: a
default call whose bodies do not fit now degrades to a complete chain without them and says
so, where before it overflowed. Flipping the default would silently strip bodies from every
caller that fits today, which is a change with no defect behind it.

**The tool now bounds its own response, and cuts bodies before edges.**
`max_response_chars` (default 80,000, clamped to 2,000..400,000) is enforced inside the walk,
measured with the same pretty serialization both surfaces use, so the budget is charged for
the payload a caller receives rather than a compact form nobody sends. It cuts in priority
order. Step bodies go first. The focal's body goes next. Only then does it drop steps, from
the tail of the chain, bisected rather than popped one at a time. A suffix is what makes that
last stage safe, because the chain is discovery-ordered and removing the end never orphans a
surviving step's parent. Dropped bodies do NOT set `truncated`, since every edge is still
there; dropped steps do, because those are edges the caller did not receive. Either cut is
disclosed through the `degradations` channel `semantic_locate` already uses, with the counts
and the parameter that avoids it, and `bodies_included` says which mode the caller got.

Both arms of the tool are bound on one shared ceiling. The daemon special-cases the
body-inlining walk in `kin-cli`; the MCP server answers from an in-process store through the
generic walk in `kin-mcp`, which inlines no bodies and so looked safe, while 200 steps of
identity, span, and signature is a six-figure character count with no source in it. The
default, floor, ceiling, and disclosure reserve now live in `kin-mcp`'s common handlers, which
`kin-cli` already depends on, because two definitions would let one arm promise a bound the
other does not keep.

**The per-step cap now keeps the callees that continue the chain.** It used to keep whichever
neighbors the relation table listed first, which is not a choice at all: `resolve_redirects`
kept `SupportsRead.read` and `HTTPAdapter.close` while dropping `get_redirect_target` and
`rebuild_method`, the two functions that decide whether a redirect exists and what method
replays it, both in its own file and both in the graph the whole time. The same defect
inverted a caller query in the same session, where a six-wide cap on `HTTPAdapter.send` kept
five test methods and a test server and dropped `Session.send` and
`HTTPDigestAuth.handle_401`, the only two real callers. Each node now collects every candidate
before any is kept and ranks them by `kin_ranking::entity_ranking::trace_fanout_score`: a
located symbol over a file-less placeholder, source over test, Calls over Imports over
References, the expanded node's own file, then its directory, then declaration kind, then the
edge's confidence, with name and id tiebreaks so the same store returns the same chain every
run. Scored against the node being expanded rather than the focal, because at depth 3 the
focal's directory says nothing about which of a distant node's callees continue the path.

**Truncation is localized.** `truncated: true` used to sit at the top level for 106 steps,
naming no node, so a caller could not tell a complete step from a clipped one and could not
repair either. Every step now carries `fanout_truncated` and `fanout_dropped`, and the
response carries `clipped_steps`: one entry per clipped node with its step index, identity,
dropped callees, dropped callers, and the `limit_per_step` that did the clipping, which is the
number a caller raises to recover exactly what was dropped.

**Every step carries one key set.** `entity` was an optional nested record, so 16 of 106 steps
arrived with a shorter shape and broke the consumer's parser twice. The record is flattened
into each step, keeping the keys a consumer already parses and adding the span, signature,
body, `span_coherence`, `entity_role`, and `external` beside them as explicit nulls where the
graph has no answer. Nothing is invented: a symbol the graph owns no file for reports null for
its location and says so with `external: true`. The focal is reported in the same shape under
`focal_entity`, so its span survives a shape query.

**One symbol now has one identity.** `cookiejar_from_dict` arrived twice in the measured
response, once located in cookies.py and once as a file-less Module, because dedup was by
entity id and an import alias is a second entity with a second id. A name index beside
`visited` now drops a file-less candidate whose name is already in the response, and fills a
file-less step in with the located record when that arrives later, re-entering the frontier at
its own depth since the placeholder had no edges to walk and the record replacing it does.
Merging is by exact name and only when one record has no file at all, which is the narrowest
rule that fixes the measured case; `external_identities_merged` reports it.

Fifteen new tests, each falsified by restoring the behavior it guards. Three of them run
through the MCP route against a fixture whose bodies are really projected from graph-owned
truth and measure the RAW result text, which is the only unit the overflow was ever reported
in; the kin-cli fixture's authority is unopenable, so a size claim proved only there would be
vacuous. One asserts a character count, because a change that re-inlines bodies on a shape
query is invisible to every assertion about chain contents. Two more assert the other
direction: a response inside its budget is byte-identical after enforcement, and a walk that
expanded every neighbor carries none of the new clip fields at all.

Closes FIR-2361

### DIFF
--- a/crates/kin-cli/src/commands/trace_data_flow.rs
+++ b/crates/kin-cli/src/commands/trace_data_flow.rs
@@ -34,31 +34,16 @@ const DEFAULT_LIMIT_PER_STEP: usize = 5;
 const MAX_LIMIT_PER_STEP: usize = 25;
 const MAX_TOTAL_STEPS: usize = 200;
 
-/// Serialized characters one response may occupy before the tool cuts its own
-/// payload.
+/// Serialized characters one response may occupy before this walk cuts its own
+/// payload, and the floor and ceiling a caller may move it to.
 ///
-/// Every other bound here shapes the WALK, and none of them bounds the ANSWER: a
-/// walk inside all of them returned 228,413 characters over 2,453 lines from a
-/// 777-entity repository, because 106 steps each inlined a full body, and the
-/// client refused the result outright. A refused result is worse than a
-/// truncated one, since a caller gets neither the chain nor a way to ask for
-/// less, so the tool now measures what it is about to return and cuts bodies
-/// before edges to fit.
-///
-/// Sized against the tool-result ceilings agent clients actually apply, which
-/// are counted in tokens; at roughly 3.5 characters per token of JSON-wrapped
-/// source this leaves headroom under a 25k-token cap for the envelope a caller
-/// wraps around it.
-pub const DEFAULT_MAX_RESPONSE_CHARS: usize = 80_000;
-
-/// Floor for a caller-supplied budget. Below this the envelope alone does not
-/// fit, so a smaller number could only be honoured by returning nothing.
-const MIN_MAX_RESPONSE_CHARS: usize = 2_000;
-
-/// Ceiling for a caller-supplied budget. A caller with a larger window may raise
-/// the bound, but not to unbounded: the daemon serving this has other callers,
-/// and a response nothing can read is not worth building.
-const MAX_MAX_RESPONSE_CHARS: usize = 400_000;
+/// Read from `kin_mcp` rather than restated here: both arms of this one tool have
+/// to promise the same bound, and the arm that serves a generic graph store lives
+/// there. A refused result is worse than a truncated one, since a caller gets
+/// neither the chain nor a way to ask for less, which is why the number exists at
+/// all — see the definition for what it was measured against.
+pub use kin_mcp::handlers::common::TRACE_DEFAULT_MAX_RESPONSE_CHARS as DEFAULT_MAX_RESPONSE_CHARS;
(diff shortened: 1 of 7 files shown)


troyjr4103:

PR #868, opened 2026-08-17
Resolve call edges by receiver scope and role, and mark what each edge proved

troyjr4103:
A Kin call edge was a candidate rather than a fact, and nothing said which. Call
targets were matched by method name with the receiver's type discarded, so on a
`kin init` conversion of psf/requests, `Session.merge_environment_settings`
appeared to call the public `requests.get` when its only `.get(` sites are
`proxies.get("no_proxy")` on a dict and `os.environ.get(...)` on the stdlib.
`Session.send` appeared to call `RedirectSession.send`, a test double that
subclasses a redirect mixin and can never be the receiver at `adapter.send`. And
`req.copy()` fanned out to three same-named methods on unrelated types. No
response field distinguished a proven edge from a guessed one, so
`find_references`, `trace_data_flow`, `impact_analysis`, `graph_neighborhood`,
`xref` and dead-code all consumed candidates as facts.

## The resolution rules

**Receivers reach the linker.** `ExtractedRelation` gains an optional `receiver`
carrying the receiver expression as written, serde-defaulted so cached parse
output keeps loading. The Python adapter records it for every attribute call
whose owning type it could not pin. The `import_source` annotation pass now
skips receiver-bearing relations, so `obj.get(...)` is no longer pinned to
whatever module happened to export a name called `get`.

**Narrowest resolution wins.** An attribute call classifies its receiver against
the calling file's own imports.

1. A receiver bound to a repo-local module resolves inside that module, first
   bare and then as `Root.method`, and emits exactly one edge at 0.9. No
   fan-out.
2. A receiver bound to a module no file in this repository can be is external.
   No local edge is emitted. The cross-repo placeholder tier still runs. This is
   what stops `os.environ.get(...)` reaching a repo function named `get`.
3. A receiver that is a value skips the tiers that bind module-level functions,
   because a call through an object cannot reach one. This is what stops
   `proxies.get("no_proxy")` binding to the public `requests.get`. The bare-name
   method tier still runs, so recall over real methods is unchanged and a real
   ambiguity still emits every candidate.

**Role decides ties.** A Source-role call site no longer resolves to a Test-role
entity while a Source-role candidate of the same name exists. Test callers are
untouched, because a test legitimately calls its own doubles.

**Every edge says what it proved.** `RelationResolution` classifies a relation as
`type_resolved` (the destination is proven by a local binding, a declared
import, or a resolved dispatch class), `import_scoped` (the target module was
known and the symbol selected inside it), or `name_only` (a bare same-name match
across the repository). It is derived from the tier confidence the linker
already persists and merkle-binds rather than added as a stored field, so every
existing store classifies as it stands, and the relation schema, kin-model and
kin-db all stay where they are. A test asserts every tier constant maps as the
ladder declares, so the two stay in step.

## Where it is published

The field is named **`resolution`** on `find_references` rows,
`trace_data_flow` steps, `graph_neighborhood` edges, ranked impact path steps,
the CLI refs listing and cross-repo xref lines. `impact_analysis` also reports
`proven_consumer_count` beside `consumer_count`. Tool descriptions state what
each value means and that a `name_only` edge is a candidate rather than a fact.
A federated reference row reports no resolution rather than guessing one,
because the resolving edge lives in another repository's graph.

Downstream consumers that must count only proven edges, dead-code analysis in
FIR-2356 above all, should use `kin_index::RelationResolution::of(&relation).is_proven()`,
which is true for everything above `name_only`. Be aware that FIR-2356's own
`reference_edge_coverage.languages[].resolution` is a different, per-language
verdict. This one is per edge.

Both the batch and the incremental resolvers apply the rules, so a live edit
cannot re-mint an edge a full index refused. The incremental checkpoint version
moves to 4 for the added role index.

## Tests

`merge_environment_settings_does_not_call_the_public_requests_get` reproduces
the founding false edge as a permanent guard, and
`the_requests_adapter_send_shape_resolves_to_the_adapter_only` reproduces the
mixin-in-tests shape. Seven more cover one-edge import narrowing, the marked
ambiguous fan-out, the incremental path, and the ladder itself.

Every rule was reverted and the tests re-run. Reverting the role tiebreak fails
three tests. Removing the object gate fails the founding-bug guard. Disabling
module narrowing fails the one-edge test. Mis-declaring one ladder entry fails
two.

Closes FIR-2360


troyjr4103:

PR #870, opened 2026-08-17
Resolve cross-file relations on the live reconcile path, in both write orders

troyjr4103:
The indexing pipeline resolves each file only against its own entities and
pushes everything else onto `IndexedFile::unresolved_relations`, a field nothing
on the live path ever read. `IncrementalLinker` was built for exactly this job
and was never constructed in production. So a file that reached the graph after
`kin init` kept `Contains` and same-file `Calls` permanently, however well its
imports would have resolved: no cross-file `Calls`, no artifact `Imports`,
`find_references` blind across files, and `trace_data_flow` unable to leave the
file it started in.

`LiveCrossFileLinker` keeps the incremental linker current as files arrive and
binds in both directions. Each direction is tested on its own, because each one
alone leaves half the defect standing.

**Forward.** The file being reconciled resolves its destinations against every
entity already in the graph. It is installed with the identities the transaction
will commit, not the raw parse identities. A modified entity keeps the id the
graph already holds, so resolving against a parse identity would mint edges into
entities the delta is about to throw away.

**Backward.** A reference an earlier file left unbound binds the moment its
destination arrives. This is what makes "write module A, then write module B"
connect A to B. Disabling it while leaving forward resolution alone still passes
the callee-first test and fails five others, so a forward-only fix reads as
correct right up until someone writes the caller first, which is the ordinary
way people work.

**Cost bound.** One seed per process, O(graph entities), the same shape and
place as the existing LKG seed. After that a write resolves the edited file plus
the files holding a still-unbound reference to a name it defines. That set is
read out of a reverse name-to-files index rather than scanned for, and those
files are re-resolved from retained parse fragments rather than re-read from
disk, so no file is parsed twice and nothing walks the repository.
`CrossFilePass::files_resolved` reports the count. The test writes twelve
unrelated modules, asserts each resolves exactly one file, then asserts the
arrival that unblocks the importer resolves two rather than fourteen.

**The removal filter, and what each branch now covers.**

`parser_authoritative` is unchanged: parser-derived, both endpoints in this
file, and the key is absent from what this pass produced. An intra-file call
site was deleted.

`duplicate_parser_relation` is unchanged: the same, except this pass did produce
that key. It retires surplus copies so one logical edge keeps one identity.

`cross_file_source_authoritative` is new: parser-derived, sourced by an entity
of this file, destination outside it, the pass ran, and this file's freshly
parsed text no longer names that destination under either its qualified or its
bare spelling. That is the case the old filter could not reach, and it keeps a
deleted cross-file call site from leaving a permanent edge behind.

The new branch deliberately does not fire when the pass merely failed to
re-derive an edge. "The source deleted this call" and "the incremental universe
could not reach the tier the batch linker used at init" are different facts, and
only the
first belongs to a single-file reconcile. A recovered parse withdraws the
branch's authority for the same reason, since a name absent from a partial tree
proves nothing. Everything else is
still preserved, including LSP-enrichment edges, agent-created Manual edges, and
any edge this file merely receives rather than sources.

Artifact-level import edges hang off no entity, so the per-entity relation
gather cannot see them at all. They are reconciled separately by walking out of
the artifact node: add what the source declares, update a changed payload,
retire one the source dropped, and skip one the graph already holds. Retirement
also requires the destination artifact to be one the linker knows, since a
destination it never heard of is one module resolution could not have reached.

Building this surfaced hazards that reading would not have. A module path
resolving to no repository file makes the linker mint a cross-repo
external-reference placeholder whose destination is a synthetic id. The batch
path turns those into real `ExternalReference` records in the same transaction
and the live path has no such step, so admitting the edge alone names an
endpoint the store rejects, taking the whole transaction down with it. And the
linker's universe is in-memory while the graph is the authority, so a caller
that fails to apply a delta leaves the universe holding entities the graph never
admitted. Every endpoint is now checked against graph truth before its edge is
emitted, and a reconcile error drops that one file from the universe rather than
cloning the universe per write to undo it.

Tests: a new suite drives the real reconciler against a real graph, writing
three importing modules one at a time in both orders. It covers the trace
crossing a file boundary, deleting a call site removing only that edge, deleting
an import retiring only its artifact edge, an edge the reconcile did not source
surviving an unrelated edit, file removal and re-creation, and the third-party
negative. `crates/kin-index/tests/python_import_linking.rs` is extended rather
than forked, walking its own fixtures through the incremental linker in both
arrival orders, including a JavaScript pair and an ambiguity case whose
same-named twin arrives first. Confidence tiers are asserted rather than only
edges, because the blind name fallback reaches the same entity at a lower tier
and an incrementally linked edge has to be indistinguishable in kind from a
batch-linked one.

Every assertion was falsified by breaking its own mechanism. Unwiring the pass
fails ten of eleven. Disabling backward binding alone fails five. Forcing either
removal-authority branch off fails exactly the test that names it. Stubbing the
incremental per-file resolver fails four of the five new linker cases. The
third-party negative survives all of them, which is what an absence assertion
should do.

#861 has landed as `ab8cc136`, so the stacking note this body used to carry is
spent. This branch is rebased onto current main and carries only its own
commits.

The branch also carried a `Cargo.lock` captured while the local DEV-LOCAL path
patches were applied: six external `kin-*` entries had lost their registry
source and checksum and a `[[patch.unused]]` tail had been appended. Every job
that builds with `--locked` refused to proceed on it, which is what the Windows
authority matrix, the Linux daemon smoke, and the Docker image build were all
failing on. This branch changes no manifest, so the tracked lock is restored to
what main holds and the diff against main carries no lock change at all.

Closes FIR-2369


troyjr4103:
Merge-queue ejection record. This PR was removed from the hosted queue by github-merge-queue[bot] at 2026-08-18T02:18:19Z and re-added at 02:18:28Z. Nothing in the PR failed: merge-group CI run 32087672141 (group head 46635f8a) had every job green except `Check & Test (ubuntu-latest)`, whose `Install runtime guard tools` step sat in `apt-get update` against a stalled Azure mirror from 01:18:56Z until the one-hour job timeout at 02:17:46Z. The queue then dropped the entry and cancelled the run. The same hang hit PR 882's own CI (job 95566904280, 01:37 to 02:37). Hardening is in #889, which bounds every apt call in ci.yml and adds a checksummed tarball fallback for ripgrep.

PR #899, opened 2026-08-18
Never report a commit still in flight as failed, and refuse an empty successor

troyjr4103:
`kin commit` built its HTTP client with a fixed 120 second deadline
(`crates/kin-cli/src/commands/commit.rs:71` at fc906464), and a one-file commit
on a converted store outlives it. The reported run shows the daemon's own phases
running long past the client: `plan_transaction elapsed_ms=68378`,
`open_repository_authority elapsed_ms=53343`, and a
`reconcile_workspace_and_commit_authority` phase of 60837 ms whose last line
landed 213 seconds after the CLI started and 93 seconds after the CLI had already
printed `operation timed out`. The deadline expiring cancelled nothing, so the
change landed anyway. The person reading the error then ran the commit again, and
the daemon accepted that second commit on a tree it had already committed and
recorded it as `entities=0 relations=0 tree=0`. That is the pair `kin log`
carried on both repositories in the container.

The client now sets no reply deadline. A deadline can only be right if the CLI
knows how long the work should take, and it cannot, because the cost is a
property of the store. The connect timeout stays at 500 ms so a daemon that is
not listening still fails immediately, and the existing phase stream keeps
printing for the whole wait, so a commit that is working and a commit that is
wedged still look different.

When a reply is lost anyway, the CLI stops treating the socket as the authority
on what happened. It looks this invocation's operation id up in the repository
authority receipts, the same lookup the daemon's own `recover_native_commit`
performs, read from the durable store rather than from a daemon that may no
longer answer. A receipt means the commit landed, so the CLI reports the change
it recorded and exits 0. No receipt plus a fresh beat from a daemon that is still
alive means the commit is still running, and the CLI says exactly that, naming
the phase, the elapsed seconds and the operation id. Neither means the commit
failed, and the CLI reports the transport error with the daemon death note behind
it, as it did before.

A marker is trusted only while its daemon answers. A daemon killed mid-commit
never retires its open-transaction marker, and the kernel's OOM killer writes no
death note beside it, so reading the beat alone would have reported the two
attempts that died in an OOM-killed container as still running.

The daemon refuses a successor that would record nothing at all
(`refuse_a_successor_that_records_nothing`, called from
`command_commit_after_admission` before the lease gate). A plan whose change
carries no entity, relation or tree delta and no admission policy delta is
refused with HTTP 409 and a body naming the change that already holds the tree,
and the CLI renders that body as its own sentence instead of wrapping it in an
HTTP envelope. The refusal lives in the endpoint rather than in the client so it
holds for any caller, including one that has already decided its first attempt
failed. A policy-only change still commits.

`OpenTransaction`, its file name, its staleness bound and its read and write
helpers move to `kin-daemon-spawn`, which both `kin-daemon` and `kin-cli` already
depend on, and `commit_liveness` re-exports them so the supervisor's behavior is
unchanged. A CLI-side copy of those field names would have reported "no
transaction open" for every shape change the writer made, which is the same
answer a dead daemon produces.

Production behavior that is preserved: a real transport failure with no receipt
and no live daemon still reports failure and still exits non-zero, and a commit
the daemon genuinely refuses still exits non-zero with its own reason.

Still open, and deliberately not attempted here: the memory ceiling of a one-file
commit on a requests-sized store, where the daemon peaked at 10.9 GiB and was
OOM-killed under a 12 GiB cap. That is part (3) of the ticket's contract and it
is FIR-2347's and FIR-2291's write-path work. This change makes that failure
honest rather than smaller.

## Tests

- `a_commit_slower_than_a_client_deadline_still_reports_its_change` drives a stub
  daemon that answers after 600 ms. A 150 ms client is the control that proves
  the daemon really is slower than a deadline, and the production client then
  receives the reply. No test waits for the old 120 seconds; the defect needs two
  durations in the wrong order, not two minutes.
- `a_lost_reply_whose_operation_reached_authority_is_a_successful_commit` and
  `a_lost_reply_with_a_beating_daemon_is_still_running_rather_than_failed` cover
  the receipt, the fresh beat, the stale beat and the marker left behind by a
  daemon that is no longer alive.
- `a_second_commit_on_an_unchanged_tree_is_refused_and_records_nothing` drives the
  daemon endpoint and asserts the authority change count does not move.
- `a_second_commit_on_an_unchanged_tree_is_refused_and_adds_no_change` drives the
  real `kin` and `kin-daemon` binaries and asserts that the refusal names the
  landed change id and that `kin log` gains no entry.

## Falsification

Each guard was broken, watched to fail with its own message, and restored.

- Wrapping the `refuse_a_successor_that_records_nothing` call in `if false`:
  `a commit that would record nothing must be refused: {"change_id":"64bf6f0b...","entity_count":0,"relation_count":0,"file_count":0}` with `left: 200`,
  and end to end `a commit on a tree Kin already holds must be refused: stdout=Created semantic change a4948738... (0 entities, 0 relations, 0 artifacts)`.
  That is the ticket's empty change, reproduced through both surfaces.
- Forcing a 150 ms timeout in `build_commit_client`:
  `a commit with no reply deadline waits for the daemon: reqwest::Error { kind: Request, source: TimedOut }`.
- Making `classify_lost_reply` always answer `Gone`: `a receipt is a landed
  commit, not Gone` and `a beating daemon is still committing, not Gone`.
- Dropping the `is_alive(open.pid)` conjunct: `a marker belonging to a dead
  daemon is a leftover, not work in flight`.

## Gates

`cargo fmt --all -- --check` clean. CI clippy, `--all-targets --all-features`
with the `ci.yml` allowlist under `RUSTFLAGS=-D warnings`, exit 0 with no
warnings. The full local gate,
`bin/kin-lane run fir2403-commit kin -- bin/kin-dev local-test kin`, printed
`kin-dev: local-test: kin /Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/fir2403-commit-kin @ c353b1ab (chore/lane-fir2403-commit-kin)`
and finished `6027 tests run: 6027 passed`, exit 0. The branch has been rebased
since onto `320cf57a`, whose only new commit is the v0.5.39 release bump and
touches none of these files. `git ls-tree -r HEAD --name-only` matches exactly
one `.cargo/` path, `.cargo/config.toml`, and `git diff origin/main --stat --
Cargo.lock .cargo/` is empty.

Closes FIR-2403

### DIFF
--- a/crates/kin-cli/src/commands/commit.rs
+++ b/crates/kin-cli/src/commands/commit.rs
@@ -67,10 +67,7 @@ async fn run_daemon_commit(
         .await?
         .ok_or_else(|| crate::daemon_client::daemon_required_error("commit", layout))?;
 
-    let client = reqwest::Client::builder()
-        .timeout(std::time::Duration::from_secs(120))
-        .connect_timeout(std::time::Duration::from_millis(500))
-        .build()?;
+    let client = build_commit_client(commit_reply_deadline())?;
     // Create these once per CLI invocation so transport retry logic can reuse
     // the byte-identical repository transaction.
     let operation_id = kin_model::OperationId::new();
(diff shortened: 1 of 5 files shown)


troyjr4103:

troyjr4103:

PR #900, opened 2026-08-18
Report one find_references row per caller and say what the count counted

troyjr4103:
MCP `find_references` reported the number of FILES a symbol is referenced from and called it `total_upstream`. On the v0.5.38 release-candidate bytes, `to_dot` has eleven callers across two files and the tool answered `2`, with `authority_complete` and `relation_subtype_complete` both reading true and `reference_lines` empty on every row. The tool's own description tells an agent to use it to gauge blast radius before a rename or a delete, so that answer sized the work at a fifth of what it was, and nothing in the payload let the reader notice.

## Mechanism

`collect_graph_reference_rows` in `crates/kin-mcp/src/handlers/common.rs:1103` grouped rows under `reference_row_key(file_path, name)`, and that key was the bare file path whenever the referencing entity had one. Every caller sharing a file collapsed into one `ReferenceRow`, which kept the first caller's id, name and signature and dropped the rest. `handle_find_references_with_authority_source` in `crates/kin-mcp/src/handlers/entities.rs:1383` then emitted `total_upstream: references.len()`, which is the count of distinct files rather than of callers.

## The fix

Rows are keyed on the referencing entity's `EntityId`, so one caller is one row. That is the unit the shared CLI collector `collect_graph_references` in `crates/kin-cli/src/commands/refs.rs` has always counted, which is both why the two surfaces disagreed and why keying on the entity is what makes them agree. Self edges are now excluded on the MCP side as well, because the CLI collector has always excluded them and a recursive call is not an upstream caller; without that the two surfaces would have stayed one apart on every recursive function.

A `counts` object states the unit in `counted` and carries `files` and the reference-site totals beside it, so a bare number is never read against the wrong unit. `reference_sites` is emitted as a number only when every returned row could be located at a line, and otherwise it is null with `known_reference_sites` as the explicit lower bound. That is the discipline `bulk_check_references` already applies to its own totals.

Each row carries `reference_line_count` and, when `reference_lines` is empty, `reference_lines_absent_reason`: `no_evidence_span`, `span_outside_caller_file`, or `federated_xref`. A row is never dropped for a missing line, because dropping one would understate the blast radius this tool exists to size. That is FIR-2357 item 3 on this surface.

## Why reference_lines is empty, which is larger than this ticket

The empty `reference_lines` is not a Python problem and not a cross-file problem. No production ingest path in the workspace ever sets `RelationEvidence::source_span`, in any language. Every construction site in `crates/kin-index/src/pipeline.rs` and `crates/kin-index/src/linker.rs` builds evidence through `..RelationEvidence::default()`, whose `source_span` is `None`, and the root cause is that `kin_parser::ExtractedRelation` at `crates/kin-parser/src/extract.rs:176` has no span field, so the call site's position is discarded inside the language adapter before the linker sees it. The Python adapter knows the byte range and throws it away at `crates/kin-parser/src/languages/python.rs:971`.

This is already measured in the repo: `crates/kin-index/tests/relation_evidence_span_coverage.rs:658` pins `MEASURED_RELATIONS_WITH_SPAN` at 0 and asserts it, with a message telling whoever lands span emission to update the constant. So `no_evidence_span` is what every real answer reports today, and filling the lines for real means adding a span to `ExtractedRelation`, threading it through the language adapters and both resolver arms, then flipping that constant. That is a separate multi-crate change in `kin-parser` and `kin-index` and needs its own ticket.

## Response size

Rows omit the caller's body and signature unless `include_snippets=true`. One row is now one caller rather than one file, so bodies scale with callers, and each body runs to `RETRIEVAL_SNIPPET_MAX_CHARS`, which is 800 at `crates/kin-mcp/src/handlers/common.rs:2252`. This is how this lane kept its rows small alongside the response-budget work, by construction rather than by depending on that lane's code. `entity_id` still drills any row to its full body through `get_entity_source`.

## Scope note

The row unit is the caller, not the individual reference site, with every site line carried on the caller's row and the site total reported separately in `counts`. The ticket's fix contract item 1 asks for exactly that, and its ground truth of eleven is eleven distinct calling functions. It is also the only unit on which the CLI and the MCP tool can agree, because `kin refs` prints "referenced by N entities" and `refs_and_bulk_count_distinct_external_entities_not_relation_edges_or_self_edges` pins that unit. No information is lost, since a caller row lists every site line it contains.

## Falsification

Every new test was broken at the behavior it guards and failed with the message it names.

Re-collapsing the rows by file after collection, which reproduces the exact defect, failed `find_references_reports_every_caller_not_one_row_per_file` ("four callers are four rows", left 2, right 4) and the cross-surface test `cli_refs_and_mcp_find_references_agree_on_the_caller_count` (left 2, right 4).

Forcing `reference_lines_absent` to `None` failed `find_references_says_why_a_row_carries_no_site_lines` ("an empty site list must name its cause, not stay silent", left Null, right "no_evidence_span").

Forcing `reference_sites_complete` to false, which is the lazy regression FIR-2357 item 4 names, failed `find_references_reports_complete_sites_when_the_graph_carries_spans` (left Null, right 2).

Dropping the outside-file span counter failed `find_references_distinguishes_a_dropped_span_from_a_missing_one` (left "no_evidence_span", right "span_outside_caller_file").

Removing the self-edge skip failed `find_references_excludes_a_self_edge` (left 2, right 1).

The tree restored clean after each, verified with `git status --porcelain` returning nothing. The agreement test asserts the absolute number 4 before asserting the surfaces match, so both regressing the same way cannot pass it, and it asserts `counts.files` is 2 so the fixture cannot pass by having every number in it be the same.

## Gate

`bin/kin-dev local-test kin`, run through `bin/kin-lane run fir2398-refs kin --` from the lane worktree with output to a file and the status read raw. The pass printed its own resolution line, `local-test: kin /Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/fir2398-refs-kin @ 4bd6a17c (chore/lane-fir2398-refs-kin)`, which is this lane's tree and commit. Result: `6006 tests run: 6006 passed (10 slow), 11 skipped`, exit 0, and "no tracked lock contamination".

`cargo fmt --all -- --check` returned 0. Clippy ran exactly as `ci.yml` invokes it, `--all-targets --all-features` under `RUSTFLAGS=-D warnings` with the 45-lint allowlist copied out of the workflow, and returned 0.

After rebasing onto `origin/main` at 320cf57a, which moved to the v0.5.39 release and touched neither `crates/kin-mcp` nor `crates/kin-cli/src/commands/refs.rs`, the two changed crates were re-run: kin-mcp 454 passed, kin-cli 1470 passed, 0 failed.

`git ls-tree -r HEAD --name-only` matches only `.cargo/config.toml` under `.cargo/`, and `git diff origin/main --stat -- Cargo.lock .cargo/` is empty.

Closes FIR-2398
Part of FIR-2357

### DIFF
--- a/crates/kin-cli/src/commands/refs.rs
+++ b/crates/kin-cli/src/commands/refs.rs
@@ -879,6 +879,165 @@ mod tests {
         );
     }
 
+    /// `kin refs` and MCP `find_references` must return the same number for the
+    /// same entity on the same graph.
+    ///
+    /// They did not. The CLI counts distinct referencing entities; the MCP tool
+    /// keyed its rows on the caller's FILE path and so counted distinct files,
+    /// which is FIR-2398. Two surfaces answering one question with two numbers
+    /// is worse than either being wrong alone, because whichever an agent read
+    /// looked internally consistent.
+    ///
+    /// The fixture is built so the two counts cannot coincide by luck: three
+    /// callers share one file, a fourth sits in another, and the target also
+    /// calls itself. Under the old rule the MCP tool saw three keys
+    /// (two caller files plus the target's own, from the self edge) against the
+    /// CLI's four entities, so a regression separates them again.
+    #[tokio::test]
+    async fn cli_refs_and_mcp_find_references_agree_on_the_caller_count() {
+        use kin_db::InMemoryGraph;
+        use kin_model::relation::{Relation, RelationOrigin};
+        use kin_model::{
+            Entity, EntityId, EntityKind, EntityMetadata, EntityRole, EntityStore, FilePathId,
+            FingerprintAlgorithm, GraphNodeId, Hash256, LanguageId, SemanticFingerprint,
+            Visibility,
+        };
+
+        fn entity(name: &str, rel_path: &str) -> Entity {
+            Entity {
+                id: EntityId::new(),
+                kind: EntityKind::Function,
+                name: name.to_string(),
+                language: LanguageId::Rust,
+                fingerprint: SemanticFingerprint {
+                    algorithm: FingerprintAlgorithm::V1TreeSitter,
+                    ast_hash: Hash256::from_bytes([0; 32]),
+                    signature_hash: Hash256::from_bytes([0; 32]),
+                    behavior_hash: Hash256::from_bytes([0; 32]),
+                    equivalence_hash: kin_model::Hash256::from_bytes([0; 32]),
+                    stability_score: 1.0,
+                },
+                file_origin: Some(FilePathId::new(rel_path)),
+                span: None,
+                signature: name.to_string(),
+                visibility: Visibility::Public,
+                role: EntityRole::Source,
+                doc_summary: None,
+                metadata: EntityMetadata::default(),
+                lineage_parent: None,
+                created_in: None,
+                superseded_by: None,
+            }
+        }
+
+        let dir = tempfile::tempdir().unwrap();
+        let layout = kin_core::KinLayout::new(dir.path().join(".kin"));
+        let graph = InMemoryGraph::new();
+
+        let target = entity("to_dot", "linkgraph.rs");
+        let callers = [
+            entity("draws_nodes", "test_linkgraph.rs"),
+            entity("draws_edges", "test_linkgraph.rs"),
+            entity("draws_dashed", "test_linkgraph.rs"),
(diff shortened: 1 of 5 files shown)


troyjr4103:

PR #902, opened 2026-08-18
Bound every retrieval tool's MCP response on one client-sized budget

troyjr4103:
Kin's own answers were arriving as files the agent had to read. On the v0.5.38 release-candidate bytes, a `semantic_locate` with three query variants at `limit: 10` came back at 80,571 characters and a `trace_data_flow` at depth 4 with `include_body: false` came back at 79,278. Claude Code refused both, spilled them to temp files, and told the model to `Read` them with an offset. That is the file-read fallback this product exists to remove, plus a round trip, on the two most common retrieval calls an agent makes.

kin#867 had already bounded `trace_data_flow`, at 80,000 characters. The measured 79,278 sat inside that budget and was refused anyway. That is the real defect. A per-tool budget the tool alone believes in reports success while the caller gets a file to read. `semantic_locate` had no character bound at all, only a row-count `limit`.

**One constant, for every retrieval tool.** `crates/kin-mcp/src/budget.rs` defines `RESPONSE_DEFAULT_MAX_CHARS = 30_000`, and the comment says why that number and which client it is sized against. Claude Code refuses a tool result above roughly 25,000 tokens. JSON-wrapped source runs near 3.5 characters per token, so the refusal threshold sits around 87,000 characters, which is exactly why two payloads at 80,571 and 79,278 both tripped it. 30,000 is about a third of that. The margin is deliberate rather than a guess about tokenization, because the client counts a result after its own framing, an agent holds several results in one window, and a budget chosen to just fit one client becomes the next overflow the moment another one counts slightly differently. `TRACE_DEFAULT_MAX_RESPONSE_CHARS` in `handlers/common.rs:426` is now an alias for it, so the two arms of `trace_data_flow` and every other retrieval tool cannot promise different bounds. `max_chars` overrides it per call, clamped to 2,000..400,000, and `max_response_chars` keeps working, because an agent that learned that spelling from `trace_data_flow` must not be handed an unbounded response for using it.

**The envelope now reports size.** `_kin.response` carries `max_chars`, `chars_before_budget`, `bounded`, and `compact`, so a caller can see that an answer was shortened and by how much without running an unbounded call it cannot make. The accounting stanza is written before the cut as well as after it, because a note added afterwards would push a response that had just been cut to fit back over its ceiling by its own length.

**Compact by default.** Ranking explanation and per-signal breakdowns are diagnostics about a hit rather than the hit, and serving them unasked is a large part of how a ten-result ranking became 80,571 characters. `compact` defaults on and sheds `match_evidence`, `explain`, `signal_scores`, and `score_breakdown`. Setting `explain: true` or `compact: false` gets them back. Under budget pressure the ladder then sheds the per-file symbol roll-up, which repeats hits `entities[]` already carries in full, then inline source, recoverable one body at a time through `get_entity_source`, and only after both of those does it withhold hits from the tail. The primary collection always keeps at least one entry, because an empty ranking cannot be told apart from "nothing matched", and that is the one reading a size cut must never produce. Every cut is disclosed in one `degradations` entry naming what went and the parameter that recovers it. `total_ranked` and `next_cursor` are left exactly as built, so the rest of a locate ranking stays reachable through the paging the tool already had.

**Both surfaces are bound, on one ladder.** `envelope::finalize_bounded` is the last point at which the bytes a client receives are known, and the bound holds there. The `_kin` envelope and the `negative` object are part of those bytes, so a handler that cut to exactly its ceiling would go back over it the moment the result was wrapped. The daemon's raw `/mcp/tools/call` route bounds too, on the same budget less a reserve for that wrapper. That route already lacked the envelope, and serving an unbounded payload there would have made it two gaps, with the same tool bounded or not depending on which door the caller came through. The budget never touches `_kin` or `negative`, which say how far a shortened answer can be trusted and matter most on exactly the answers that were shortened.

**One call is bounded once.** The reserve is this server's arithmetic about its own default, so a budget the caller names is enforced exactly, on every arm, and the daemon's trace walk now defaults to the number the route will hold rather than to its own. kin#867's `a_trace_over_its_budget_drops_bodies_and_keeps_every_edge` caught the first version of this change doing the opposite: a caller asking for 6,000 characters had the walk cut bodies at 6,000 and the route cut steps again at 4,000, so a chain that lost only bodies came back marked truncated. When the walk and the route hold different ceilings for one call, the response gets cut to a number it never names.

**Ranking is unchanged.** Every scorer, fusion step, and ordering rule is the same, and an unbounded request returns the same result set it returned before. `max_chars: 400000` returns the same hits in the same order. What changed is the shape and the size of what gets serialized.

**Measured, before and after, on fixtures that overflow.** Each number is the final annotated text a client would count, printed by the test that asserts it:

| tool | unbounded | bounded |
|---|---|---|
| `semantic_locate` (three variants, limit 10, through the real fused serializer) | 117,183 | 12,446 |
| `graph_neighborhood` | 238,077 | 29,069 |
| `get_context_pack` | 215,465 | 28,504 |
| `find_references` | 180,034 | 28,572 |
| `trace_data_flow` (depth 4, include_body false) | 131,665 | 28,877 |
| `bulk_check_references` | 102,106 | 28,723 |
| `semantic_search` | 90,406 | 28,863 |

**Falsification.** Disabling the `enforce` call in `envelope::apply_response_budget` fails five of the six kin-mcp tests, one of them with `find_references: the response a client receives must fit its budget: 180032 chars against 30000`. The sixth, `trace_data_flow`, survives that edit because it bounds itself inside its own walk, so it also asserts the number it enforces IS the shared default. Restoring `TRACE_DEFAULT_MAX_RESPONSE_CHARS` to 80,000 fails it with `left: Number(80000), right: Number(30000)`. Disabling the `enforce` call in the daemon's `bound_mcp_tool_result` fails `the_raw_route_bounds_a_tool_the_stdio_path_would_have_bounded` at `245776 chars against 28000` and the locate test at `117183 chars against 28000`. Disabling the compact stage fails `compact_sheds_breakdowns_without_touching_the_hits` on a surviving `match_evidence`, and fails the locate test on the same key. `the_raw_route_leaves_an_unbudgeted_tool_byte_identical` stays green under every one of those edits. The budget is scoped, and a tool whose job is handing back requested source stays untouched.

Closes FIR-2396

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -8040,11 +8040,81 @@ fn resolved_entity_source_from_outcome<E: std::fmt::Display>(
     }
 }
 
+/// POST /mcp/tools/call.
+///
+/// The route bounds every retrieval response it serves, on the same budget and
+/// through the same ladder the stdio MCP path applies. This arm carries no
+/// `_kin` envelope, which was already a gap, and serving an unbounded payload
+/// here would have widened it into a second one: a client calling the daemon
+/// directly would have been handed exactly the response the stdio client is
+/// protected from, and the tool would have been bounded or not depending on
+/// which door the caller came through.
+///
+/// It cuts against the budget less the envelope reserve, because the stdio path
+/// wraps this payload in `_kin` and `negative` before a client counts it. Both
+/// arms measuring the full budget would put every enveloped response slightly
+/// over it and make the stdio pass re-cut a payload that had already been cut to
+/// fit.
 async fn mcp_tools_call(
     headers: axum::http::HeaderMap,
     State(state): State<Arc<DaemonState>>,
     Json(request): Json<McpToolCallRequest>,
 ) -> Result<impl IntoResponse, (StatusCode, String)> {
+    let budget =
+        kin_mcp::budget::ResponseBudget::from_arguments(&request.arguments).less_envelope_reserve();
+    let tool = request.name.clone();
+    let Json(result) = mcp_tools_call_inner(headers, State(state), Json(request)).await?;
+    Ok(Json(bound_mcp_tool_result(result, &tool, &budget)))
+}
+
+/// Apply the response budget to one tool result's payload text.
+///
+/// Measured and rewritten on the serialized payload rather than inside each
+/// handler because the tools this route serves build their JSON in six different
+/// places, and a bound implemented once per handler is a bound six code paths
+/// can each forget. A payload that is not JSON, or a tool the budget does not
+/// govern, is returned exactly as built.
+fn bound_mcp_tool_result(
+    result: kin_mcp::ToolCallResult,
+    tool: &str,
+    budget: &kin_mcp::budget::ResponseBudget,
+) -> kin_mcp::ToolCallResult {
+    if !kin_mcp::budget::is_budgeted(tool) {
+        return result;
+    }
+    let content = result
+        .content
+        .into_iter()
+        .map(|block| {
+            let kin_mcp::ContentBlock::Text { text } = block;
+            let Ok(mut payload) = serde_json::from_str::<serde_json::Value>(&text) else {
+                return kin_mcp::ContentBlock::Text { text };
+            };
+            if !payload.is_object() {
+                return kin_mcp::ContentBlock::Text { text };
+            }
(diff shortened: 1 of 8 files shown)


troyjr4103:

troyjr4103:

PR #903, opened 2026-08-18
Keep a server-bound MCP repository when the client's workspace roots cannot be resolved

troyjr4103:
A Kin MCP server registered as `docker exec -i -w /work/express <container> kin mcp start` refused every tool call after the first. That registration is the shape of every containerised or remote deployment, so this is not a corner case: the client's root lives in one namespace and the server's repository in another, by construction.

## Mechanism

At `crates/kin-mcp/src/server.rs`, the stdio loop asks the client for `roots/list` on `initialized`/`tools/list` while nothing is bound, and always on `notifications/roots/list_changed`. The answer went through `apply_workspace_roots`, which called the binder and got back a bare `Option<BoundRepo>`. `crates/kin-cli/src/commands/mcp.rs` `bind_first_kin_repo_against` returned `None` both when the client had moved to another repository and when nothing the client named resolved to a Kin repository at all, and the loop read every `None` as the former: with `binding.is_bound()` already true from the launch cwd, it called `binding.mark_mismatch(roots, repo_pinned)` and refused every following `tools/call`.

In a container the client's root is a host path that never exists in the server's namespace, so the per-call re-binding kin#862 added could not help: the re-check runs, fails identically, and the refusal is permanent for the life of a process the client owns. Exactly one call survived each respawn, the one that arrived before the roots response did, which is why the stranger run measured one call per restart across five restart cycles.

The refusal also reused `Envelope::daemon_unreachable()`, so it reported a reachability problem it did not have. The bound daemon answered `/health` 200 throughout. Its three remedies all needed an `.mcp.json` edit and a client restart, neither of which an agent mid-session can do.

## The fix

The binder now reports what it saw. `WorkspaceBinding` carries three verdicts: `Bound`, `OtherRepository` (the roots name Kin repositories this process can see and does not serve), and `Unresolvable` (no root resolves to a Kin repository here). `KinLayout::discover` is filesystem-rooted, so the split is decided by what the server can actually see rather than by whether a daemon happened to resolve. The comment claiming `discover` short-circuits on `KIN_DAEMON_URL` was stale and is corrected; its own doc says discovery is always filesystem-rooted.

`RepoBindingState` now records where its binding came from. A binding the server made for itself, from its launch cwd or `--repo`/`KIN_MCP_REPO`, survives roots it cannot resolve: it keeps serving and logs the disagreement at info. A binding that came from a previous `roots/list` answer still refuses when the client withdraws it, because the client's roots were its only authority. Roots naming a different Kin repository are refused either way, since answering from the repository the client left is the failure the refusal exists to prevent.

The refusal carries a distinct `degraded.workspace_mismatch` instead of `daemon_unreachable`, and names both the bound repository and the announced workspace. `kin mcp start --help` and `docs/mcp-tools.md` now state that a bound server ignores client roots it cannot resolve, so the docker-exec registration is a documented, supported shape.

## Falsification

Every new guard was broken and confirmed to fail with its own message, then restored.

1. `a_server_bound_repository_survives_roots_it_cannot_resolve`, with the provenance branch removed so `Unresolvable` always refuses while bound: `FAILED ... call 41 must not be refused for a root this server cannot resolve`. This is the red-before, green-after test for the reported defect.
2. `a_server_bound_repository_still_refuses_a_second_kin_repository` and `a_repo_bound_refusal_does_not_suggest_kin_init`, with the `OtherRepository` arm disabled: both `FAILED`, the first on `call 41 must be refused and name the tool`.
3. `a_workspace_refusal_is_not_stamped_daemon_unreachable`, with `Envelope::daemon_unreachable()` put back into `repo_mismatch_response`: `FAILED ... the refusal must carry its own degradation: {"daemon_unreachable":true}`, which is the exact payload the ticket reports.
4. `a_root_this_server_cannot_see_is_unresolvable_not_another_repository` and `roots_with_no_kin_repository_bind_nothing`, with the empty-candidates branch returning `OtherRepository`: both `FAILED`, the first on `a root that resolves to no Kin repository here must not be reported as a switch`.
5. `mcp_start_help_documents_that_a_bound_server_holds_unresolvable_client_roots`, with the help paragraph deleted: `FAILED ... help must state that a bound server ignores roots it cannot resolve`, and the printed help proves what remained.
6. `workspace_mismatch_envelope_is_not_a_reachability_failure`, with the constructor setting both flags: `FAILED ... assertion failed: env.degraded.daemon_unreachable.is_none()`.

Each test also carries its own in-test falsification: the unresolvable case is paired with a visible second repository that must still refuse, the envelope test asserts the reachability envelope never claims a workspace mismatch, and the help test asserts a phrase that was never written is absent.

## Gate

`bin/kin-lane run fir2405-roots kin -- bin/kin-dev local-test kin` passed, resolved to `/Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/fir2405-roots-kin @ 29e9bd63 (chore/lane-fir2405-roots-kin)`: 6018 tests run, 6018 passed, 11 skipped, plus doctests. `cargo fmt --all -- --check` is clean and `cargo clippy --all-targets --all-features` under `RUSTFLAGS=-D warnings` with the ci.yml allowlist exits 0. `git ls-tree -r HEAD --name-only | grep -F '.cargo/'` prints only `.cargo/config.toml`, and `git diff origin/main --stat -- Cargo.lock .cargo/` is empty. Rebased onto `origin/main` at 320cf57a before pushing.

Composes with kin#862, whose per-call re-derivation still runs and still self-heals a root that becomes bindable, and with kin#880. The `crates/kin-mcp/src/envelope.rs` change is deliberately minimal, one `Degraded` flag plus its constructor and test, so it rebases cleanly against the concurrent lane touching that file.

Closes FIR-2405

### DIFF
--- a/crates/kin-cli/src/commands/mcp.rs
+++ b/crates/kin-cli/src/commands/mcp.rs
@@ -103,7 +103,7 @@ pub async fn start(
     // the shared handle at each invocation.
     let binder_startup = std::sync::Arc::clone(&startup);
     let repo_binder: Option<kin_mcp::RepoBinder> = Some(Box::new(
-        move |roots: Vec<PathBuf>| -> Pin<Box<dyn Future<Output = Option<kin_mcp::BoundRepo>> + Send>> {
+        move |roots: Vec<PathBuf>| -> Pin<Box<dyn Future<Output = kin_mcp::WorkspaceBinding> + Send>> {
             let startup = std::sync::Arc::clone(&binder_startup);
             Box::pin(async move {
                 bind_first_kin_repo(roots, startup.pinned_by_operator()).await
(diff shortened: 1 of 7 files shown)


troyjr4103:

PR #904, opened 2026-08-18
Emit and resolve a reference edge for a Python type annotation

troyjr4103:
A Python class named only in a type annotation produced no graph edge of any class, so a parameter type, a return type and a dataclass field type each reached the graph as nothing at all. `find_references` on such a class answered with only the callers in its own file, and it answered confidently: on the reported project it returned one reference with `total_upstream: 1` and `cross_repo.authority_complete: true` while `storage.py` took the class as a parameter type. A reader who renamed on that answer would have left the old name behind in the signature. The same miss hid `WikiLink` from `ParsedNote`, whose `links: Tuple[WikiLink, ...]` field is the type it is built from, so the context pack for the class omitted the type it depends on.

## Mechanism

`crates/kin-parser/src/languages/python.rs` pruned annotations from the value-reference walk kin#877 added. `collect_python_value_refs` (python.rs:579 at 320cf57a) listed `"type"` among the node kinds it skipped outright, `extract_value_refs_from_definition` (python.rs:444 at the same sha) walked parameter defaults but never the `type` field beside them and never the `return_type` field at all, and the annotated `assignment` arm read only its `right` child. `git grep -c UsesType` on that file was 0, so annotations produced no edge of any class rather than an edge of the wrong one.

## The fix

`collect_python_type_refs` walks one annotation subtree and pushes the same `PythonValueRef` records the value walk produces. It is called from three positions: the `type` field of each parameter, which also carries `*args: T` and `**kw: T` whose names sit in a splat pattern while the annotation stays on the same field; the `return_type` field of the function; and the `type` field of an annotated assignment, which is the single shape a class-body field, a module-scope annotation and a body-local annotation all wear.

It descends the wrappers tree-sitter builds a type expression from rather than matching each one, which is what makes `Optional[Note]`, `Tuple[WikiLink, ...]`, `dict[str, Note]`, `Note | None`, `Callable[[Note], int]` and `typing.List[Note]` each reach the names they carry. A quoted forward reference is read only when the quotes hold a bare identifier.

Nothing about binding is new. Candidates go through the existing `emit_python_value_references`, so only a name the file defines or imports may bind, and the resolved edges take the linker tiers a call takes: import-based at 0.95, namespace member at 0.9.

## Why `References` rather than `UsesType`

`UsesType` exists in `kin-model` (`kin-model/src/relation.rs:216`) and the ticket allowed either kind. `References` is the one that works, and that is checkable rather than a preference. `kin_core::reference_coverage::REFERENCE_RELATION_KINDS` (`crates/kin-core/src/reference_coverage.rs:30`) is the published Calls + Imports + References triple, and its own doc says it is the set `find_references` defaults to. `kin_mcp::handlers::common::default_reference_kinds` returns exactly those three, `parse_relation_kinds("all")` in `crates/kin-cli/src/commands/refs.rs:692` returns exactly those three, and the linker gates cross-file resolution on Calls or References at `crates/kin-index/src/linker.rs:1169`, `:3090`, `:3160` and `:5072`. Emitting `UsesType` would have meant editing all four surfaces before the edge could be counted anywhere. That is the argument kin#877 made for value reads, and it holds harder here, because the whole point of this ticket is that `find_references` must count the edge.

## Falsification

Four falsifiers, each applied to the committed tree and then reverted.

Unwiring all three annotation hook sites, which restores the pre-fix behavior exactly, turns 7 of the 10 new parser unit tests red, every one of them reporting `got []`: `a parameter annotation must reference its class, got []`, `a return annotation must reference its class, got []`, `a dataclass field type must reference its element type, got []`, and four more. All 6 linking tests present at that time go red, and 3 of the 4 acceptance tests go red. The rename one is the sharpest, because without the edge `plan_rename` does not merely omit the site, it refuses the whole rename: `exhaustive rename-reference coverage disagrees for graph entity ...: graph declarations and edges require 0 'ParsedNote' occurrence(s), but repository CAS assigns 1; refusing a partial refactor`.

Dropping the `defined.contains(root) || symbol_bound.contains(root)` filter in `emit_python_value_references` turns `an_annotation_naming_a_builtin_or_an_unimported_name_emits_nothing` red with `got [("take", "Ghost"), ("take", "bool"), ("take", "int"), ("take", "str")]`, and `an_annotation_naming_a_class_this_file_never_imported_emits_no_edge_at_all` red with `left: Some(0.7)` against `right: None`, which is the blind cross-file exact-name tier binding a dependency the file never declared.

That second linking test exists because of what the falsification found. The original twin fixture, two modules defining `ParsedNote` where only one is imported, stayed GREEN when the gate was removed, because the name it probes IS imported and the linker's import tier is what holds it, while its comment claimed the extractor's gate was the mechanism. The comment was corrected and a case the gate actually holds was added beside it, in the second commit.

Dropping the shadow set from the annotation walk turns `an_annotation_naming_a_local_binding_emits_nothing` red with `got [("take", "Note")]`. `a_parameter_name_is_still_not_a_reference` needs two rules broken at once to fail, walking the whole parameter node instead of only its `type` field AND dropping the shadow set, after which it reports `got [("take", "note")]`. Breaking either alone leaves it green, which is recorded in its comment rather than left implied.

## Gate

`bin/kin-lane run fir2399-annot kin -- bin/kin-dev local-test kin` printed `local-test: kin /Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/fir2399-annot-kin @ 761a85c5 (chore/lane-fir2399-annot-kin)` and `Summary [225.097s] 6033 tests run: 6033 passed (2 slow, 9 leaky), 11 skipped`, exit status 0 read raw. `cargo fmt --all -- --check` exits 0. Clippy run exactly as `ci.yml` runs it, `--all-targets --all-features` under `RUSTFLAGS=-D warnings` with the workflow's 45-lint allowlist, exits 0. Both zero-file-search guards pass. Tracked `.cargo/` holds only `.cargo/config.toml`, and `git diff origin/main --stat -- Cargo.lock .cargo/` is empty.

`bin/kin-magic-repro` against this lane's release build reads 9 of 9 PASS plus check 0 GPU-OPTOUT PASS, exit status 0. Both binaries stamp `0.5.38 (34c4e911176d302d9478848e86eed705cd3b50d7 chore/lane-fir2399-annot-kin)`, and it ran under the daemon lock taken RC-only.

## Not claimed

Python import statements still carry no entity-level edges; they remain artifact-level, which is FIR-2354's territory. That is why the rename acceptance fixture reaches its class through `import parsing` and `parsing.ParsedNote` rather than `from parsing import ParsedNote`: `require_repository_reference_coverage` refuses outright while a graph source imports the target by name, so the named-import shape cannot reach the rename planner at all today, with or without this change. The acceptance test also does not drive the MCP handler end to end; it reads the graph under `default_reference_kinds` itself, which is the kind filter that decides whether a consumer is reported, and asserts the rendered `kin refs --kind all` answer beside it, because the handler's body projection needs a live repository authority binding an in-memory fixture does not have. Annotations on a function nested inside another function body are not covered, since such a function produces no entity of its own in this adapter.

Closes FIR-2399

### DIFF
--- a/crates/kin-cli/tests/python_type_annotation_references.rs
+++ b/crates/kin-cli/tests/python_type_annotation_references.rs
@@ -1,0 +1,375 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Firelock, LLC
+
+//! Acceptance reproduction for the type-annotation edge class, on the shape a
+//! stranger's Python project produced.
+//!
+//! `parsing.py` defines `ParsedNote` and `WikiLink`. `storage.py` imports
+//! `ParsedNote` and never calls it: it appears as a parameter type, as a return
+//! type, and as a dataclass field type, and `WikiLink` appears only as the
+//! element type `ParsedNote` is built from. Because the adapter emitted no edge
+//! of any class for an annotation, `find_references("ParsedNote")` answered with
+//! the one caller in its own file and `find_references("WikiLink")` missed
+//! `ParsedNote` entirely, so a reader who renamed `ParsedNote` on that answer
+//! would have left `storage.py` holding the old name in three places.
+//!
+//! Everything here runs through the real Python adapter, the real cross-file
+//! linker, and the collector `find_references` itself calls over the relation
+//! kinds it defaults to.
+
+use std::collections::{BTreeSet, HashMap};
+
+use kin_cli::commands::refs::{build_refs_response, RefsRequest};
+use kin_cli::commands::rename::{plan_rename, RenameRequest};
+use kin_db::InMemoryGraph;
+use kin_index::{link_cross_file_with_completeness, FileParseData};
+use kin_mcp::handlers::common::default_reference_kinds;
+use kin_model::{
+    ArtifactId, AuthorId, Entity, EntityId, EntityStore, FileLayout, FilePathId, GraphNodeId,
+    Hash256, ImportSection, OperationId, ParseCompleteness, RepoPath, ResolvedArtifact,
+    ResolvedTree, TreeEntry,
+};
+use kin_parser::{LanguageAdapter, PythonAdapter};
+
+/// The defining module. Neither class is ever called.
+const PARSING_PY: &str = r#""""Note parsing."""
+
+
+class WikiLink:
+    pass
+
+
+class ParsedNote:
+    pass
+"#;
+
+/// The consuming module. `ParsedNote` is named three times and called zero
+/// times; `WikiLink` is named once, as the element type of a field.
+const STORAGE_PY: &str = r#""""Note storage."""
+
+from typing import Optional, Tuple
+
+from parsing import ParsedNote, WikiLink
+
+
+class NoteRow:
+    note: Optional[ParsedNote] = None
+    links: Tuple[WikiLink, ...] = ()
+
+
+def upsert_note(note: ParsedNote, mtime: Optional[float] = None) -> int:
+    return 1
+
+
+def latest() -> ParsedNote:
+    raise NotImplementedError
+"#;
+
+/// A third module with no consumer at all, so a fix that rescued everything
+/// would be caught as surely as one that rescued nothing.
+const ORPHAN_PY: &str = r#""""Nothing depends on this."""
+
+
+class OrphanRecord:
+    pass
+"#;
+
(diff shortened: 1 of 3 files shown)


troyjr4103:

PR #905, opened 2026-08-18
Refuse absence authority for a language whose reference edges cannot exist

troyjr4103:
`find_references` certified an absence over express's `createApplication` with
`safe_to_conclude_absent: true` and `trust: authoritative` while the same payload's
`edge_coverage` read `imports: absent` and `references: absent`. An agent acting on that
answer deletes express.

The gate is `edge_coverage_gap` in `crates/kin-mcp/src/negative.rs`. It returned no gap as
soon as any requested class read `present`, by design: one cross-file witness was the bar.
On JavaScript the `calls` class is present because same-name bare calls resolve, so that one
witness certified an absence over a `module.exports` every consuming file reaches through
`require`.

The obvious fix does not work, and measuring is what showed it. Requiring an `imports`
witness looks principled and would report every Python absence inconclusive, because a
converted Python repository whose `from .parsing import ...` statements resolve cleanly
reports the identical `imports: absent`. Kin resolves an import statement to a cross-file
`Calls` edge plus an artifact-level edge, and mints no entity-level `Imports` edge for any
language: `kin graph status` on such a repository prints `Entity-to-entity relation kinds:
Calls: 6, Contains: 3` and `python: 4 files, calls 6/0, imports 0/2 (0%), cross-file 4`. A
build carrying that rule turned a genuinely unused Python symbol inconclusive with
`cross_file_edges_absent`, which is the failure mode opposite to FIR-2353 and no more useful.

What separates the two cases is the build, not the coverage. `crates/kin-mcp/src/edge_coverage.rs`
now publishes `reference_enrichment` beside the measured classes, derived from
`kin_core::cross_file_coverage::ENRICHABLE_LANGUAGES`. A language this build wires no
language-server adapter for cannot produce reference or override edges at any coverage, so an
absence over it is pinned inconclusive with that reason however healthy its calls look. That
is the same fact `kin graph status` already prints as "cross-file reference and override edges
unsupported for javascript: this build wires no language-server adapter". Whether an installed
server backs a wired adapter is a host fact this path does not probe, so it publishes as
`unknown` and gates nothing, leaving Rust and Python answers unchanged.

The class gate stays and is narrowed honestly. `load_bearing_classes` names `calls` as the
class a certification depends on, with a fallback to whatever the query asked for so an
`imports`-only or `references`-only query is still gated on what it read. The absent-class
reason now names the classes that were missing and the ones that were present.

`degraded_signals` carries every shortfall the answer's own `edge_coverage` names, so the
negative object can no longer read "no degraded signals" beside an observation naming two
absent classes.

Falsification, each reverted after the run:

* Removing the enrichment gate turns both express tests red and reproduces the ticket
  verbatim: `safe_to_conclude_absent: true`, `trust: authoritative`, trust_reason
  `structural_authoritative: daemon graph initialized and loaded with no degraded signals`,
  advice `Absence is authoritative: safe to treat the target as genuinely absent/unused.`
  It fails at the unit level and through the real handler and observation path, so the
  reproduction does not depend on a hand-written payload.
* Making the gate always inconclusive turns
  `a_python_graph_that_resolves_its_imports_still_certifies_an_unused_symbol`,
  `a_graph_with_cross_file_edges_still_certifies_a_genuinely_unused_symbol` and
  `find_references_certifies_absence_once_the_graph_links_calls_across_files` red, so the fix
  cannot degrade into marking everything inconclusive.
* Dropping the coverage labels from `degraded_signals` turns the disclosure assertions red
  with `left: Array []`.
* Reporting `reference_enrichment` as `unknown` for every language turns
  `a_language_with_no_adapter_reports_reference_enrichment_unsupported` and the store-backed
  express test red with `left: String("unknown") right: String("unsupported")`.

End to end on the release build this branch produced, `kin 0.5.39
(cdd01e1966bd13f164301995031335544a152af0)`, against an express-shaped JavaScript repository
whose consumers all reach the export through `require`: `find_references(createApplication)`
returns zero rows, `edge_coverage` reproduces the ticket's shape
(`calls: present, imports: absent, references: absent`, `cross_file_classes: ["calls"]`) with
`reference_enrichment: "unsupported"`, and the negative reads `safe_to_conclude_absent: false`,
`trust: inconclusive`, trust_reason leading `reference_enrichment_unsupported`, advice
`Absence is NOT authoritative`. On the Python repository the same build no longer contributes
any edge-coverage gap at all: the reason for that store's absence moved from
`cross_file_edges_absent` under the rejected rule to the pre-existing
`cross_repo_authority_incomplete`, which this change does not touch.

`bin/kin-magic-repro --only 2,7` against those binaries reports `2 pass, 0 fail, 0 unreadable`.
CHECK 2 passes by returning a reference rather than by exercising the absence path, so the
absence assertions above carry that half.

Closes FIR-2404

### DIFF
--- a/crates/kin-mcp/src/edge_coverage.rs
+++ b/crates/kin-mcp/src/edge_coverage.rs
@@ -29,6 +29,7 @@
 //!   "requested_classes": ["calls", "imports", "references"],
 //!   "classes": { "calls": "present", "imports": "present", "references": "absent" },
 //!   "cross_file_classes": ["calls", "imports"],
+//!   "reference_enrichment": "unknown",
 //!   "budget_exhausted": false,
 //!   "entities_examined": 3
 //! }
@@ -40,6 +41,10 @@
 //! before it could say. A consumer that cannot find this object must treat
 //! coverage as unknown rather than assuming either verdict.
 //!
+//! `reference_enrichment` carries the one fact the scan cannot observe from the
+//! graph: whether this BUILD can produce reference edges for the language at
+//! all. See [`reference_enrichment`].
+//!
 //! The single-focal tools attach it to an EMPTY answer only, since that is the
 //! only answer whose trust depends on it: one that returned rows proved the
 //! edges exist by returning them. The batch tool attaches it always, because its
@@ -48,9 +53,10 @@
 
 use serde_json::{json, Map, Value};
 
+use kin_core::cross_file_coverage::{ReferenceEnrichment, ENRICHABLE_LANGUAGES};
 use kin_model::entity::Entity;
 use kin_model::graph::{EntityFilter, EntityStore};
-use kin_model::ids::{EntityId, FilePathId};
+use kin_model::ids::{EntityId, FilePathId, LanguageId};
 use kin_model::relation::RelationKind;
 
 /// Reserved, additive payload key carrying the cross-file edge-class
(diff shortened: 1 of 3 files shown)


troyjr4103:

PR #906, opened 2026-08-18
Refuse a bare Python call that only a receiver could reach

troyjr4103:
A bare call to a Python builtin resolved by name alone to whatever single entity in the repository carried that name. On the six-module fixture the isolated stranger run used, `parse_file` in `parsing.py` reads a file with `open(path, "r", encoding="utf-8", errors="replace")`, and the linker's bare-name index held exactly one `open`: the `NoteStore.open` classmethod in a storage module `parsing.py` never imports. That single candidate captured the call, so `trace_data_flow` from `ingest_directory` walked into the storage module and reached `default_db_path` underneath it, and `find_references` on the classmethod listed `parse_file` as a caller.

## Mechanism

A bare `open(...)` reaches the linker as `ExtractedRelation { receiver: None, dst_name: "open" }` (`crates/kin-parser/src/languages/python.rs`, `extract_named_callee`, the `identifier` arm). It then falls past every tier that could prove a destination: same-file finds nothing, the `self`/`cls` walk needs a dotted callee, the import tiers find no `open` in the file's import map, and the exact-name bucket holds none because a method is indexed under `NoteStore.open`. It lands at the receiver-method fan-out, `crates/kin-index/src/linker.rs` tier (c2), which reads `entity_by_bare_name["open"]`, finds the one method there, and emits it at confidence 0.3, which `RelationResolution::of` classifies `name_only`.

The value-reference path already applied the right rule and the Calls path did not: `emit_python_value_references` binds only a name the file defines or imports, and its test `a_name_this_file_neither_defines_nor_imports_emits_no_reference` covers `len`. Nothing equivalent guarded a call.

## The fix

Two gates, in `resolve_one_file` and its incremental twin `resolve_one_file_incremental`.

Tier (a1), the builtin gate: a bare identifier naming a Python builtin, from a file that neither defines nor imports it, emits no edge and logs the gap. This is not a heuristic. A module-level name is reachable inside a Python module only through a definition or an import, so there is no third possibility to preserve. The table is the full public surface of CPython's `builtins` module, read off `sorted(n for n in dir(builtins) if not n.startswith("_"))` on CPython 3.13 plus `WindowsError`, 151 names, and lives in `crates/kin-parser/src/languages/python.rs` as `PYTHON_BUILTIN_NAMES` with `is_python_builtin_name` binary-searching it. A remembered dozen would have omitted `format`, `id`, `filter`, `next` and `exit`, which all read like ordinary repository functions.

Inside tier (c2), the method gate: a bare identifier never binds to a method-kind entity, because a method needs a receiver and this call shape has none. That tier exists for calls made through a receiver.

Both gates are scoped to Python. Python is the only adapter that separates the shapes at extraction time: an attribute call carries its receiver as written, a `self`/`cls` call folds its owner into the callee name, so a relation with neither is a bare identifier. Ten of the thirteen adapters record no receiver at all, so reading a missing receiver as proof of a bare call there would drop every receiver-method edge they resolve. `a_non_python_bare_call_still_reaches_its_method` pins that scoping.

A wildcard import records no specifiers, because there is no name to record, so a file carrying one cannot answer whether it binds the name and the builtin gate stands down rather than guessing. `INCREMENTAL_LINKER_CHECKPOINT_VERSION` is bumped 4 to 5, which its own doc comment requires on a linker semantics change.

## Falsification

Each gate was removed in turn and the named test went red. Both were restored; the tree carries neither edit.

Builtin gate off, method gate on: `a_bare_python_builtin_call_reaches_no_cross_module_function` FAILED, "bare `open`/`len` calls must not reach same-named module functions", two `confidence: 0.7` edges. 113 passed, 1 failed.

Method gate off, builtin gate on: `a_bare_python_call_reaches_no_method_it_has_no_receiver_for` and its incremental twin FAILED, "a bare call must not dispatch to a method", `confidence: 0.3`. 112 passed, 2 failed.

Both off: the ticket's own shape FAILED, `a_bare_python_builtin_call_reaches_no_cross_module_method`, "a bare builtin `open` must not reach a method of an unimported module", `confidence: 0.3`. 109 passed, 5 failed.

Both off, end to end: `tracing_calls_never_crosses_a_builtin_open_into_the_storage_module` FAILED with the chain `["parse_file", "NoteStore.upsert_note", "NoteStore.open", "TAG_RE", "default_db_path"]`, and the refs test FAILED with:

```
References to 'NoteStore.open' (Method) @ storage.py
referenced by 2 entities:
  open_store @ cli.py:7 [Calls] (import_scoped)
  parse_file @ parsing.py:9 [Calls] (name_only)
```

The positive control `tracing_calls_from_an_importing_caller_still_reaches_the_classmethod` stayed green in that same run, which is what says the gates refuse an edge rather than delete an entity.

## Tests

Ten unit tests in `crates/kin-index/src/linker.rs`, three acceptance tests in a new `crates/kin-cli/tests/python_bare_builtin_call_resolution.rs` that run the real Python adapter, the real cross-file linker, and a real `InMemoryGraph`, and two tests in `crates/kin-parser/src/languages/python.rs` for the table's sortedness and its coverage of the names a short list omits.

Recall controls, all green: a module defining its own `open` still binds at confidence 1.0; an imported symbol shadowing a builtin still resolves; `from storage import open_store` plus a bare `open_store(...)` still resolves; `store.open()` in a file importing `NoteStore` still resolves through the receiver scoping kin#888 added; a star import stands the builtin gate down; a TypeScript bare call still reaches its method.

## Gate

`bin/kin-lane run fir2400-builtin kin -- bin/kin-dev local-test kin` printed `local-test: kin /Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/fir2400-builtin-kin @ d21dc8b2 (chore/lane-fir2400-builtin-kin)`, then `Summary [852.889s] 6027 tests run: 6027 passed (3 slow, 2 leaky), 11 skipped`, exit status 0 read raw. `cargo fmt --all -- --check` exit 0. Clippy exactly as `.github/workflows/ci.yml` runs it, `--all-targets --all-features` under `RUSTFLAGS=-D warnings` with the workflow's 45-lint allowlist, exit 0.

Main moved two commits after that gate, a v0.5.39 release bump and the glibc floor scripts, neither touching linker or parser code. The branch was rebased onto `320cf57a` and the three affected suites re-run at the new tip: kin-index linker 114 passed, the new acceptance file 3 passed, the parser table tests 2 passed. `git ls-tree -r HEAD --name-only | grep -F '.cargo/'` prints only `.cargo/config.toml`, and `git diff origin/main --stat -- Cargo.lock .cargo/` is empty.

## Magic repro

`bin/kin-magic-repro` ran against this lane's own release build, both binaries, under the `daemon` lock taken RC-only and released in a trap. The GPU lock was never taken and the suite exports `KIN_DAEMON_AUTO_EMBED=0`, which its own check 0 asserts the daemon honored. Both binaries stamp `0.5.39 (fde606a83e71d7de7bcca0fc4dad682d047357ad chore/lane-fir2400-builtin-kin)`. Result: `10 pass, 0 fail, 0 unreadable`, exit status 0, checks 0 through 9. CHECK 5, the FIR-2360 call-edge-precision check that shares this territory, reads `PASS the ambiguous receiver in Session.drain produced 1 edge(s)`.

Closes FIR-2400

### DIFF
--- a/crates/kin-cli/tests/python_bare_builtin_call_resolution.rs
+++ b/crates/kin-cli/tests/python_bare_builtin_call_resolution.rs
@@ -1,0 +1,265 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Firelock, LLC
+
+//! Acceptance reproduction for the fabricated cross-module call edge a bare
+//! Python builtin used to mint (FIR-2400).
+//!
+//! The fixture is the shape the isolated stranger run hit: a parsing module
+//! that reads a file with the builtin `open`, and a storage module it never
+//! imports whose `NoteStore` exposes an `open` classmethod. The bare name
+//! reached the linker's bare-name index, matched the one entity in the graph
+//! carrying it, and `trace_data_flow` from `ingest_directory` walked a whole
+//! subtree of a module `parsing.py` cannot see, `default_db_path` included.
+//!
+//! Everything here runs through the real Python adapter, the real cross-file
+//! linker, and a real `InMemoryGraph`, which is what `kin init` builds over an
+//! existing tree. The positive controls sit in the same fixture on purpose: the
+//! receiver call `NoteStore.open(path)` in `cli.py` must still resolve, so a
+//! gate that simply deleted the entity from the graph would fail this file.
+
+use std::sync::Arc;
+
+use kin_cli::commands::refs::{build_refs_response, RefsRequest};
+use kin_cli::commands::repository_authority::RequestRepositoryAuthority;
+use kin_cli::commands::trace_data_flow::{
+    build_trace_data_flow_response, TraceDataFlowRequest, TraceDirection,
+};
+use kin_db::InMemoryGraph;
+use kin_index::{link_cross_file, FileParseData};
+use kin_model::{ArtifactId, Entity, EntityStore, FilePathId};
+use kin_parser::{LanguageAdapter, PythonAdapter};
+
+/// `parse_file` reads with the builtin `open` and imports nothing from storage.
+/// This is the call site the ticket names, verbatim in shape.
+const PARSING_PY: &str = r##""""Note parsing."""
+
+import os
+import re
+
+TAG_RE = re.compile(r"#(\w+)")
+
+
+def parse_file(path):
+    with open(path, "r", encoding="utf-8", errors="replace") as handle:
+        text = handle.read()
+    return {"name": os.path.basename(path), "tags": TAG_RE.findall(text)}
+"##;
+
+/// `NoteStore.open` is a classmethod, which no bare-name call can invoke, and
+/// `default_db_path` is reachable from it. In the measured response those were
+/// steps 8 and 16.
+const STORAGE_PY: &str = r#""""Note storage."""
+
+import os
+
+
+def default_db_path(root):
+    return os.path.join(root, "notes.json")
+
+
+class NoteStore:
+    def __init__(self, path):
+        self.path = path
+        self.notes = {}
+
+    @classmethod
+    def open(cls, path):
+        return cls(default_db_path(path))
+
+    def upsert_note(self, note):
+        self.notes[note["name"]] = note
+        return note
+
+    def prune_except(self, keep):
+        self.notes = {k: v for k, v in self.notes.items() if k in keep}
+        return self.notes
+"#;
+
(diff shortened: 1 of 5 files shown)


troyjr4103:

PR #907, opened 2026-08-18
Name the runtime embedding-model download on every surface

troyjr4103:
No install ships the embedding weights. `kin-db` builds its Hugging Face client with `hf_hub::api::sync::Api::new()` (`kin-db` `crates/kin-db/src/embed/mod.rs:12`, used at `mod.rs:766`), so the first local embed batch downloads about 523 MB of `nomic-ai/nomic-embed-text-v1.5` before a single vector exists. The daemon dispatches that batch inside `spawn_blocking` from the embed worker in `crates/kin-daemon/src/daemon.rs`, so the whole download sits inside one pass turn: the pass has declared itself working and has credited no units. `crates/kin-cli/src/commands/resources.rs` therefore printed `Background pass embed: working 109s, 0 units, no progress yet` and `crates/kin-cli/src/commands/graph.rs:561` printed `Embeddings: 0/1658 indexed (1658 pending)` for the entire fetch. Both lines are equally true of a wedged pass, which is how an isolated first run read them, and on a host with no egress they are the only signal there will ever be.

This adds `crates/kin-cli/src/embed_model.rs` as the one place that knows the model id, the cache root, the destination directory, the bytes already there, and what to say about all of it. It mirrors the constructor the weights actually arrive through: `Api::new()` resolves `Cache::default()`, which is `dirs::home_dir()/.cache/huggingface/hub` and never reads `HF_HOME` (`hf-hub-0.5.0/src/lib.rs:201-208`); only `ApiBuilder::from_env()` reads `HF_ENDPOINT` (`src/api/sync.rs:249`) and only `Cache::from_env()` reads `HF_HOME`, and the embedder calls neither. Progress is the bytes under that directory with `.incomplete` blobs counted and symlinked snapshot entries skipped, so the numerator moves with the fetcher rather than with a clock and a resolved snapshot is never counted twice. `retrieval_profile.rs` lends the existing cache-layout helpers rather than growing a second layout.

Four surfaces carry it. `kin init` prints an `Embedding model:` line before it returns, naming the model, the size, the source and the destination when this machine does not have the weights, and saying the first pass downloads nothing when it does; the `kin.init-result.v6` payload carries the same facts additively. `kin resources inspect` renders an `Embed model:` line in every state and describes the pass as `Background pass embed: downloading embedding model (137 of 523 MB), working 109s, 0 units, no progress yet` while it is inside the fetch. `kin graph status` appends the same cause to its embedding counters, and a fetch that has moved no bytes says exactly that with the egress it needs instead. `kin doctor` gains an `embedding_model` check: healthy when the weights are present or when the configuration fetches nothing (a remote provider, or a `KIN_EMBED_MODEL_ID` that already names a local directory), pending when they are absent and the host answers, and Missing, which blocks readiness, when a bounded three-second TCP probe of the model host does not answer. That detail names the budget it established rather than claiming the host is down, and a model this build has never measured is given no size at all rather than the default model's figure. `kin embed --help`, `docs/cli-reference.md` and `docs/quickstart.md` state the runtime fetch, its size and how to pre-seed it.

`docs/security/what-leaves-the-machine.md` claimed the fetch honors `HF_HOME`. It does not, for the reason above, and the claim is corrected rather than repeated. This is the same root split FIR-2264 reports for the reranker and nothing here changes it: the probe mirrors the loader's own root so a numerator is never taken from a directory nobody is writing to, and where `HF_HOME` points elsewhere the disagreement is reported instead of being discovered by a second download.

Falsification. Every new guard was broken, confirmed to fail with its named message, and restored, leaving `git status --porcelain` empty. Returning `Pending` instead of `Missing` for an unreachable host failed `an_unreachable_model_host_fails_loud_with_its_egress_requirement` with `a host that cannot fetch the model can never embed: Pending`. Making a cached model announce a download failed `a_cached_embedding_model_is_healthy_and_names_where_it_sits`. Making `download_phase()` always return `None` failed `a_fetch_in_flight_renders_its_measured_progress` and `the_embed_pass_names_the_model_download_it_is_inside` with `left: "Background pass embed: working 109s, 0 units, no progress yet"`. Making `status_clause()` always return `None` failed `a_pending_backlog_names_the_model_download_that_is_blocking_it`. Rendering zero bytes as a download failed `a_fetch_with_no_bytes_names_the_egress_instead_of_claiming_progress` with `left: "downloading embedding model (0 of 523 MB)"`. Dropping the size and destination from the init notice, and separately warning a cached machine about a download, both failed `init_states_the_model_download_a_fresh_machine_still_owes`. Giving an unmeasured model the default model's size failed `an_overridden_model_is_never_given_the_default_models_size` with `no measured size may be attributed to an unmeasured model: acme/private-embed ... fetches about 523 MB`. Following symlinks in the cache walk failed `cache_bytes_count_real_files_once_and_never_follow_links` with `left: 6218, right: 6144`.

Gates. `cargo fmt --all -- --check` is clean on the rebased tree. Clippy runs exactly as `ci.yml` runs it, under `RUSTFLAGS=-D warnings` with the workflow allowlist, and is clean. The full local gate `bin/kin-dev local-test kin --repo-dir kin=<lane worktree>` is rc 0 with 6036 tests run, 6036 passed and 11 skipped, and the run printed its tree as `.kin-coord/worktrees/fir2395-modeldl-kin @ 076d0cc0 (chore/lane-fir2395-modeldl-kin)`. `git ls-tree -r HEAD --name-only | grep -F '.cargo/'` prints only `.cargo/config.toml`, `git diff origin/main --stat -- Cargo.lock .cargo/` is empty, and `git diff origin/main --numstat` lists only the thirteen files this change touches.

Not claiming: no daemon was run against a real store and no model was downloaded in this lane, so the rendered strings are proven by unit tests over injected state and by source readings, not by a live fetch. The host probe establishes only that a TCP connection to port 443 did not complete inside its budget. FIR-1223 asks for a pre-seeded model path and a `kin embed --model-path` flag, and neither is added here; what this adds is the disclosure FIR-2395 asked for, including naming `KIN_EMBED_MODEL_ID` pointed at a local directory as the way to embed without egress today.

Closes FIR-2395


troyjr4103:

troyjr4103:

PR #908, opened 2026-08-18
Mark a context pack's same-file fallback rows and keep the focal body in compact mode

troyjr4103:
`get_context_pack` answered a class that has no dependency edge with six rows under `dependencies`, and nothing in the payload said they were not dependencies. `crates/kin-context/src/builder.rs:428` gathers `direct_dep_ids` over dependency edges only, and when that set is empty it queries every entity sharing the focal's file, ranks them in three coarse buckets, and keeps the first six. Each of those rows was tagged by prefixing `// same-file neighbor` to `entry.content` at builder.rs:580, and `handle_get_context_pack` in `crates/kin-mcp/src/handlers/entities.rs` builds every row from the store entity and never emits `entry.content`, so the only tag distinguishing a neighbour from an edge died at the MCP boundary. The six-row cap was silent in both modes, so a twenty-four-member class came back as six rows with no count of what was dropped. Separately, `focal_context_json_held` in `crates/kin-mcp/src/handlers/common.rs` gated the focal `body` behind `if !compact` while reading the excerpt unconditionally, so compact mode paid for the read and discarded the one thing the pack is for.

`build_context_pack_with_provenance` and `build_context_pack_with_traffic_and_provenance` now return a `DependencySelection` beside the pack, and the existing `build_context_pack` and `build_context_pack_with_traffic` stay as wrappers so no other caller had to move. The selection says which selection filled the section, how many same-file candidates there were, and how many the cap or the token budget dropped; the cap is now the named constant `SAME_FILE_FALLBACK_MAX`. `get_context_pack` stamps every dependency row with `relation: "dependency_edge"` or `relation: "same_file_neighbor"` and carries a `dependency_selection` object (`source`, `returned`, `same_file_candidates`, `same_file_dropped`) in both compact and full mode. Transitive, test and contract rows carry no relation, because the field is about the dependency section. The tool description names the fallback so an agent knows what a dependency list on a class means.

`kin context` reports the same facts in the same words. Its rendered line reads `Dependencies: 3 entries (same-file neighbors, no dependency edges; kept 3 of 3)` or `Dependencies: 1 entries (dependency edges)`, and the structured response gained a `dependency_selection` field that defaults to absent, so a daemon predating it still decodes.

Compact mode keeps the focal body, still bounded by `MCP_SOURCE_MAX_LINES` (40) and `MCP_SOURCE_MAX_CHARS` (2400), and still drops the dependency bodies and projection levels. `focal_context_json` and `focal_context_json_held` lost the `compact` parameter they no longer read. Rows grew by one short string field and the selection object is four values per response, so this stays small alongside kin#902, which adds a client-sized response budget and envelope-level `compact` semantics across the retrieval tools. The two do not fight: kin#902 sheds explain keys (for this tool, `projection`) and bodies only under budget pressure, while the `compact` argument this handler reads still shapes the dependency rows, and the focal body it keeps is the answer kin#902 is written to preserve. Nothing here reads or depends on that PR's code, and the hunks do not overlap: kin#902 touches `handlers/entities.rs` at its tool descriptions near lines 123 and 903 and adds tests at the end of that file, while this change is in `handle_get_context_pack` and adds its tests in `handlers/mod.rs`.

Falsification, each break applied to the committed tree and reverted after the run:

- Removing the `relation` insertion from both row shapes: `a_context_pack_marks_its_same_file_fallback_rows` and `a_context_pack_marks_real_dependency_edges_as_edges` both FAILED, on "a fallback row must say it is a neighbour, not an edge (compact=false)" and "an edge row must say so (compact=false)".
- Removing the `dependency_selection` object: both FAILED, on "the pack must name the selection that filled it (compact=false)" and "a pack built from edges must not report the fallback".
- Making `relation_for` return `SameFileNeighbor` for every id: the fallback test PASSED and the edge test FAILED, which is why both exist; `builder::tests::dependency_edge_rows_are_not_reported_as_same_file_neighbors` FAILED too.
- Dropping the fallback bookkeeping in the builder: `same_file_fallback_reports_its_source_and_what_the_cap_dropped` FAILED with `left: DependencyEdges, right: SameFileFallback`, and the edge test stayed green.
- Restoring the compact body gate: `a_compact_context_pack_still_carries_the_focal_body` FAILED on "a compact pack must still serve the focal body".
- Making compact stop trimming rows: the same test FAILED on its control, "compact must still drop the neighbour bodies and projections".
- Dropping the selection note from the `kin context` rendering: both CLI tests FAILED, on "the human rendering must name the fallback" and "an edge-built pack must say so".

Gates: `cargo fmt --all -- --check` exits 0. The CI clippy allowlist from `.github/workflows/ci.yml` under `RUSTFLAGS=-D warnings` exits 0. The full local gate ran through `bin/kin-lane run fir2401-pack kin -- bin/kin-dev local-test kin`, which printed the tree it gated as `.kin-coord/worktrees/fir2401-pack-kin @ 06a735aa (chore/lane-fir2401-pack-kin)` and finished `Summary [321.456s] 6019 tests run: 6019 passed (3 slow, 2 leaky), 11 skipped` at exit code 0, with the doctest pass beside it and no tracked lock contamination. `git ls-tree -r HEAD --name-only` matches only `.cargo/config.toml` under `.cargo/`, and `git diff origin/main --stat -- Cargo.lock .cargo/` is empty.

Closes FIR-2401

### DIFF
--- a/crates/kin-cli/src/commands/context.rs
+++ b/crates/kin-cli/src/commands/context.rs
@@ -69,6 +69,27 @@ pub struct ContextTarget {
     pub budget_tokens: usize,
 }
 
+/// How the pack's dependency section was filled.
+///
+/// A pack whose focal has no dependency edge falls back to neighbours from the
+/// focal's own file, and the pack carries no field saying so: the tag lives in
+/// a comment inside each entry's content. `kin context` and the
+/// `get_context_pack` MCP tool report these same facts under these same names,
+/// so a reader moving between the two surfaces is not learning two
+/// vocabularies for one selection.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ContextDependencySelection {
+    /// `dependency_edges` or `same_file_fallback`.
+    pub source: String,
+    /// Rows in `pack.dependency_signatures`.
+    pub returned: usize,
+    /// Same-file neighbours the fallback had to choose from. Zero when the
+    /// fallback did not run.
+    pub same_file_candidates: usize,
+    /// Same-file neighbours the cap or the token budget dropped.
+    pub same_file_dropped: usize,
+}
+
 /// The context response, structured half added alongside the rendered lines.
 ///
 /// `lines` stays first and required so an older client keeps decoding this, and
@@ -84,6 +105,10 @@ pub struct ContextResponse {
     /// rendering answers with guidance rather than a pack.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub pack: Option<kin_model::context::ContextPack>,
+    /// Absent for the same reason as `pack`, and from a daemon that predates
+    /// the field.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub dependency_selection: Option<ContextDependencySelection>,
 }
 
 pub async fn run(
@@ -161,6 +186,7 @@ pub fn build_context_response(
             schema_version: CONTEXT_RESPONSE_SCHEMA_VERSION.to_string(),
             target: None,
             pack: None,
+            dependency_selection: None,
         });
     };
     let opts = kin_context::ContextOptions {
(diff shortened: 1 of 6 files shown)


troyjr4103:

PR #909, opened 2026-08-18
Probe the kin-vfs driver, and stop doctor and setup overclaiming

troyjr4103:
Three surfaces reported work they had not done. Each is fixed where it decided, and each state is now something a user can act on.

**`kin doctor` could not tell a broken kin-vfs driver from an absent one (FIR-2394).** The projection check inferred presence from one path: `projection_installed_under` (crates/kin-cli/src/commands/health.rs:665 on c4c26d65) tested `~/.kin/bin/kin-vfs` and nothing else, so an archive or Homebrew install, which puts the driver beside the `kin` binary, read as absent. The row then stated that guess as fact: "Neither is present here, so nothing is missing" (health.rs:776). On the 2026-08-18 stranger container both `kin-vfs` and `libkin_vfs_shim.so` sat in `/opt/kin` beside the binary doctor had just resolved by full path, and the driver could not load at all. Doctor now resolves the driver beside the running binary, in `~/.kin/bin`, and on PATH, runs the one it finds, and reports three distinct states: absent keeps the green n/a and now names where it looked, present and loadable says the driver runs, present and unloadable is a failure carrying the loader's literal message.

The probe runs `--help`, not `--version`, and judges on whether the process reached its own code rather than on a zero exit. The shipped `kin-vfs` takes a subcommand and declares no version, so `kin-vfs --version` exits 2 with a clap usage error on a healthy install: probing that way would have failed every install that works, which is the same shape of wrong answer this check exists to remove. A loader refusal looks different. `ld.so` prints its complaint and exits 127, dyld kills the process, and a wrong-architecture or unreadable file fails at spawn.

**`kin doctor --fix` wrote a PATH export for a directory it never created (FIR-2402).** `plan_rc_update` appended the `~/.kin/bin` export unconditionally (setup.rs:2345 on c4c26d65), and only the launcher-provisioned layout populates that directory, so an archive install ended up with a `.bashrc` naming a path that did not exist before the repair or after it. The repair now writes that line only when the directory exists, says out loud when it skips it, and the doctor row states which of the two cases the install is in instead of promising a PATH entry for nothing.

**`kin setup` reported a client as configured that doctor had just called absent (FIR-2406).** Doctor's AI-client row read config files only (health.rs:1612), so "no AI client config files detected, nothing to configure" was an answer about configuration, not about what is installed; `kin setup` uses a different detection and configured Claude Code seconds later on the same host. Both surfaces now read one detection, so they cannot disagree. Setup also stops calling a client "configured" when it is not installed, which is reachable through the advanced picker: that case now reads "pre-configured for a client that is not installed" and names the file. The discovery reminder, which is a standing directive in a file every future session of that client reads, is announced with its block heading and path before setup appends it, not only after.

Falsification, each guard broken in turn and restored:

- The unloadable arm removed: `a driver that cannot run must need attention, got Unsupported`, and `the_three_driver_states_are_three_different_rows` fails on `matches!(unloadable.status, HealthStatus::Misconfigured)`.
- The exe-directory candidate removed: `a driver beside the running binary must be probed: ["/tmp/.../kin-home/bin/kin-vfs", "/Users/.../.kin/bin/kin-vfs"]`.
- `driver_ran` replaced by `status.success()`: `a driver that ran and complained about arguments is installed and loadable`, left `Unloadable { message: "error: unexpected argument found" }`.
- The PATH skip removed: `no PATH line may reference a directory that does not exist`, and the repair test fails with `a repair may not write a PATH reference to a directory it did not install`.
- The row's directory test removed: `a hook installed by an archive install is not broken: Misconfigured`.
- Detection ignored in the client row: `a detected client must be named: no AI client detected and no client config files present, so there is nothing to configure`, and the end-to-end row fails with `an installed client must reach the row`.
- `client_write_summary` forced to the detected branch: `an absent client must be named as absent: Claude Code configured (/home/u/.claude.json)`.
- The reminder marker changed: the new test fails, and the pre-existing `discovery_reminder_injection_is_idempotent` appends the block twice.

Checked against the real binary, built from this branch. With an unloadable `kin-vfs` beside it and no shim, `kin doctor` prints:

```
  ✗ VFS projection    MISCONFIGURED  the kin-vfs driver at /tmp/dp-e2e/opt/kin-vfs is installed but will not run: kin-vfs: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by kin-vfs)
      fix:  reinstall kin for this platform: curl -fsSL https://get.kinlab.dev/install | sh. If the message names a system library version this host does not have, projection cannot run here, and the CLI and daemon are fully functional without it.
```

With the driver removed the row returns to the green n/a and names where it looked. `kin doctor --fix` against a home with no `~/.kin/bin` prints:

```
  Skipped the PATH line: /tmp/dp-e2e/kin-home/bin does not exist, and this install put nothing there
  Appended to /tmp/dp-e2e/home/.bashrc
  ✓ Shell integration    ok    bash hook installed and sourced from /tmp/dp-e2e/home/.bashrc; no PATH line was added because /tmp/dp-e2e/kin-home/bin does not exist on this host
```

The rc carries only the hook source line and `~/.kin/bin` is still not created. On that same run the client row reads `Cursor, Google Antigravity detected with no client config file yet`, where before it read "nothing to configure". On the developer machine, where a real driver is installed, the row reads `shim installed (928928 bytes, ~/.kin/lib/libkin_vfs_shim.dylib); kin-vfs driver at ~/.kin/bin/kin-vfs runs`.

Gate: `bin/kin-lane run doctorpaper kin -- bin/kin-dev local-test kin` on `/Users/.../.kin-coord/worktrees/doctorpaper-kin @ e1830c61 (chore/lane-doctorpaper-kin)`, 6010 tests run, 6010 passed, 11 skipped, exit 0, plus the doctest pass. `cargo fmt --all -- --check` clean and clippy clean under `RUSTFLAGS=-D warnings` with the ci.yml allowlist. `python3 scripts/test-release-workflow-authority.py` exits 0. That run predates the rebase onto 320cf57a (v0.5.39), which touched only release version strings and applied with no conflicts; `git diff origin/main --numstat` lists only the three files here, and `git diff origin/main -- Cargo.lock .cargo/` is empty.

Closes FIR-2394
Closes FIR-2402
Closes FIR-2406

### DIFF
--- a/crates/kin-cli/src/commands/health.rs
+++ b/crates/kin-cli/src/commands/health.rs
@@ -14,8 +14,9 @@ use serde_json::Value;
 
 use crate::commands::auth::default_base_url_for_health;
 use crate::commands::setup::{
-    check_binary_in_path, configured_mcp_launcher, detect_shell, home_dir, hook_filename, kin_dir,
-    shell_rc, shim_filename, CANONICAL_NPM_MCP_COMMAND, CANONICAL_NPM_MCP_PACKAGE,
+    check_binary_in_path, configured_mcp_launcher, detect_shell, detected_ai_client_names,
+    home_dir, hook_filename, kin_dir, shell_rc, shim_filename, CANONICAL_NPM_MCP_COMMAND,
+    CANONICAL_NPM_MCP_PACKAGE,
 };
 use crate::daemon_client::{InstalledStartupProtocol, SupervisorStartupSentinel};
 
@@ -637,8 +638,12 @@ fn check_vfs_projection() -> HealthCheck {
         }
     };
     let lib_path = kin_home.join("lib").join(shim_filename());
+    let driver = resolve_vfs_driver(&vfs_driver_candidates(
+        &kin_home,
+        env::current_exe().ok().as_deref(),
+    ));
 
-    vfs_projection_check_for(&lib_path, projection_installed_under(&kin_home))
+    vfs_projection_check_for(&lib_path, &driver)
 }
 
 /// Name of the projection driver binary the installer places beside `kin`.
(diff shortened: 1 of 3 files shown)


troyjr4103:

troyjr4103:

PR #913, opened 2026-08-18
Land the stranger-run fixes: commit deadline and empty duplicates, reference rows, MCP budget and roots, absence gate, annotation edges, builtin calls, model download, context-pack rows, doctor and setup, phase-capture flake

troyjr4103:
This is one composed landing of the eleven open stranger-run fix branches, kin#899 through kin#909, merged in order onto `origin/main` at `320cf57a` (v0.5.39). Each merge commit on the branch names the PR it carries, so the branch history is the composition record; the hosted queue squashes on landing and this body is the squash message. Every branch's own PR body remains the detailed record of its mechanism, falsification and gate; what follows is one paragraph per fix and the proof that they compose.

**Commit deadline and empty duplicates (kin#899, FIR-2403).** `kin commit` no longer sets a reply deadline on its HTTP client, because the cost of a commit is a property of the store and a fixed 120 second deadline made a landed commit read as `operation timed out`. When a reply is lost anyway the CLI reads the operation's receipt from the durable repository authority: a receipt means the commit landed and the CLI reports it and exits 0; no receipt plus a fresh beat from a live daemon means the commit is still running, and the CLI says so with the phase, elapsed seconds and operation id; a marker whose daemon is dead is a leftover, not work in flight. The daemon refuses a successor that would record nothing (no entity, relation, tree or admission-policy delta) with HTTP 409 naming the change that already holds the tree, so a retry after a misreported failure cannot mint an `entities=0 relations=0 tree=0` change. `OpenTransaction` and its helpers move to `kin-daemon-spawn`, which both binaries already depend on.

**Reference rows (kin#900, FIR-2398, part of FIR-2357).** MCP `find_references` keyed its rows on the referencing file, so eleven callers in two files came back as two rows and `total_upstream: 2`. Rows are now keyed on the referencing entity, self edges are excluded as the CLI collector always did, and a `counts` object states the unit (`counted`), the file count and the reference-site totals, with `reference_sites` a number only when every row could be located at a line and `known_reference_sites` the explicit lower bound otherwise. Each row carries `reference_line_count` and, when its lines are empty, `reference_lines_absent_reason`. A cross-surface test pins `kin refs` and the MCP tool to the same caller count.

**Phase-capture flake (kin#901, FIR-2407).** `the_cli_commit_path_names_every_phase_it_spends` could capture two phases out of ten because tracing-core resolves a callsite's cached `Interest` against the calling thread's dispatcher whenever exactly one dispatcher is registered, so a parallel test reaching `timed_commit_phase` first poisoned the callsite as `never` for the capturing thread. `capture_events_on_this_thread` now registers a second dispatcher for the life of the capture, so interests union to `sometimes` and each emitting thread is asked once per event. The reproduction stays in the test permanently, and `a_repeat_deferral_escalates_from_debug_to_warn` moves to the same seam. Production is unchanged; the seam is `cfg(test)`.

**MCP response budget (kin#902, FIR-2396).** A `semantic_locate` at 80,571 characters and a `trace_data_flow` at 79,278 were both refused by Claude Code and spilled to files, the second while sitting inside its own 80,000-character per-tool budget. `crates/kin-mcp/src/budget.rs` defines one `RESPONSE_DEFAULT_MAX_CHARS = 30_000` for every retrieval tool, `max_chars` overrides it per call within 2,000..400,000, and the bound is enforced at `envelope::finalize_bounded`, the last point where the bytes a client receives are known, and on the daemon's raw `/mcp/tools/call` route with the same budget less a wrapper reserve. `_kin.response` reports `max_chars`, `chars_before_budget`, `bounded` and `compact`. Compact defaults on and sheds ranking diagnostics; under pressure the ladder sheds the per-file roll-up, then inline source, then tail hits, always keeping at least one primary entry and disclosing every cut in `degradations`. Ranking is unchanged.

**Server-bound MCP roots (kin#903, FIR-2405).** A server registered as `docker exec -i -w /work/express <container> kin mcp start` refused every tool call after the first, because the client's `roots/list` answer named a host path the server's namespace cannot resolve and the loop read that as a move to another repository. The binder now reports `Bound`, `OtherRepository` or `Unresolvable`, and a binding the server made for itself (launch cwd, `--repo`, `KIN_MCP_REPO`) survives roots it cannot resolve while a binding that came from a previous `roots/list` still refuses when withdrawn. The refusal carries `degraded.workspace_mismatch` instead of `daemon_unreachable` and names both repositories. `kin mcp start --help` and `docs/mcp-tools.md` document the shape.

**Python type annotations (kin#904, FIR-2399).** A class named only in a parameter type, a return type or a dataclass field type produced no graph edge of any class, so `find_references` on it answered from its own file with `authority_complete: true`. `collect_python_type_refs` walks the `type` field of each parameter, the `return_type` field, and the `type` field of an annotated assignment, descending the wrappers tree-sitter builds a type expression from, and feeds the same `emit_python_value_references` gate so only a name the file defines or imports may bind. The edge is `References`, the kind every counting surface already reads.

**Absence gate for unenrichable languages (kin#905, FIR-2404).** `find_references` certified an absence over express's `createApplication` as `authoritative` while its own `edge_coverage` read `imports: absent, references: absent`, because one present class (`calls`) was the bar. `edge_coverage` now publishes `reference_enrichment` from `ENRICHABLE_LANGUAGES`; a language this build wires no language-server adapter for pins an absence inconclusive with that reason however healthy its calls look, and `degraded_signals` carries every shortfall the observation names. Requiring an `imports` witness was measured and rejected because it turned a genuinely unused Python symbol inconclusive too.

**Bare Python builtin calls (kin#906, FIR-2400).** A bare `open(...)` in a file that neither defines nor imports `open` resolved at confidence 0.3 to the one `NoteStore.open` method in the repository, so `trace_data_flow` walked into an unimported storage module. Two Python-scoped gates in `resolve_one_file` and its incremental twin: a bare identifier naming one of the 151 CPython builtins from a file that neither defines nor imports it emits no edge, and a bare identifier never binds to a method-kind entity because a method needs a receiver. A wildcard import stands the builtin gate down. `INCREMENTAL_LINKER_CHECKPOINT_VERSION` moves 4 to 5.

**Embedding-model download (kin#907, FIR-2395).** No install ships the weights, and the first embed batch fetched about 523 MB inside one pass turn, so `kin resources inspect` and `kin graph status` read exactly like a wedged pass. `crates/kin-cli/src/embed_model.rs` mirrors the loader's own cache root and measures bytes as they arrive; `kin init` names the download before it returns, `kin resources inspect` and `kin graph status` say `downloading embedding model (N of 523 MB)` while inside the fetch, `kin doctor` gains an `embedding_model` check that reads Missing when a bounded probe of the model host does not answer, and the `HF_HOME` claim in `docs/security/what-leaves-the-machine.md` is corrected.

**Context pack fallback rows (kin#908, FIR-2401).** `get_context_pack` filled `dependencies` for a class with no dependency edge from six same-file neighbours and said nothing, and compact mode paid for the focal body and discarded it. `build_context_pack_with_provenance` returns a `DependencySelection`; every dependency row is stamped `relation: "dependency_edge"` or `"same_file_neighbor"`, a `dependency_selection` object reports the source, the count returned and what the cap dropped, `kin context` renders the same facts, and compact mode keeps the focal body bounded by `MCP_SOURCE_MAX_LINES` and `MCP_SOURCE_MAX_CHARS`.

**Doctor and setup overclaiming (kin#909, FIR-2394, FIR-2402, FIR-2406).** `kin doctor` resolves the kin-vfs driver beside the running binary, in `~/.kin/bin` and on PATH, runs `--help`, and reports absent, loadable, or unloadable with the loader's literal message, judging on whether the process reached its own code rather than on exit 0. `kin doctor --fix` writes the `~/.kin/bin` PATH line only when that directory exists and says when it skips it. Doctor's AI-client row and `kin setup` read one detection, and setup no longer calls an uninstalled client configured.

## Composition proof

Every merge ran with `-c rerere.enabled=false`. All eleven merges auto-merged with no textual conflict. `cargo check --workspace --all-targets` ran through the lane after each merge so a semantic break is attributed to the branch that introduced it. Two were found and are folded into the merge commit that introduced them: after kin#903, `Envelope::workspace_mismatch()` lacked the `response: None` field kin#902 added to `Envelope` (a compile error, E0063); after kin#909, two statements in `health.rs` exceeded rustfmt's chain and call widths at that branch's own tip and `cargo fmt --all` reflowed them, no token changed. Every other intermediate check was clean.

Four of the individual PRs were CI-red at their own tips, reproduced locally on a `git archive` export of each tip with `origin/main` as the passing control, and repaired here in three commits named for the PR each repairs. kin#899: the Windows authority leg found `authority_change_count` unused under `-D warnings` because its only caller is a `#[cfg(unix)]` test, so the helper now carries the same gate; and the OpenTransaction marker's fs reads and writes had moved to `kin-daemon-spawn` without their zero-file-search pins, so the pins now follow the code (`3c553cc7`). kin#902: `check_runtime_boundaries.sh` refused a `SessionRegistry::new()` in a `handlers/entities.rs` test, so `session.rs` gains a `cfg(test)` constructor and the test uses it (`7c71f03c`). kin#907 and kin#909: the zero-file-search guard had no entry for `embed_model.rs` and a stale pin for the driver probe in `health.rs`; both are pinned with their reasons (`7e6a2f65`). Every script guard `ci.yml` runs (`verify-zero-file-search.py`, `zero_file_search_guard.sh`, `check_runtime_boundaries.sh`, `check_private_repo_coupling.sh`, `verify-hydration-semantics.py`, `test-private-repo-coupling.py`, `test-assertion-reachability.py`, `test-homebrew-release-gate.py`, `test-install-checksum.py`) exits 0 on the composition.

`git diff origin/main --numstat` lists 45 files, 9,318 insertions and 483 deletions. Of the 43 files the branches touch, 31 are touched by exactly one branch and are byte-identical to that branch's version. Twelve are touched by two or more branches; for each, the multiset of added and removed lines against `origin/main` equals the union of the branches' own added and removed lines exactly, except for the one `response: None,` line above (the check reports that line, which is how it shows it can say no). The two files outside that set are `crates/kin-mcp/src/session.rs` (the eight-line test constructor) and `scripts/zero-file-search-allowlist.json`. `git ls-tree -r HEAD --name-only` matches only `.cargo/config.toml` under `.cargo/`, and `git diff origin/main --stat -- Cargo.lock .cargo/` is empty; no branch bumped a dependency and the lockfile did not move.

`cargo fmt --all -- --check` exits 0. Clippy exactly as `ci.yml` runs it, `--all-targets --all-features` under `RUSTFLAGS=-D warnings` with the 45-lint allowlist, exits 0. The full local gate, `bin/kin-lane run landwave kin -- bin/kin-dev local-test kin`, printed `local-test: kin /Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/landwave-kin @ 7e6a2f65 (chore/lane-landwave-kin)` and finished `Summary [271.659s] 6140 tests run: 6140 passed (3 slow, 4 leaky), 11 skipped` with 22 doctest runs all ok, exit status 0 read raw. `bin/kin-magic-repro` and `bin/kin-release-preflight` against the release build of this tip were still running when this PR opened; their result lines are added to this body before the PR is enqueued.

Left out on purpose: kin#910, kin#911 and kin#912 opened while this lane ran; they are CI-workflow changes rather than stranger-run fixes, touch only `.github/workflows/*` and `scripts/test-release-workflow-authority.py`, share no file with this branch, and ride separately.

Closes FIR-2403
Closes FIR-2398
Closes FIR-2407
Closes FIR-2396
Closes FIR-2405
Closes FIR-2399
Closes FIR-2404
Closes FIR-2400
Closes FIR-2395
Closes FIR-2401
Closes FIR-2394
Closes FIR-2402
Closes FIR-2406
Part of FIR-2357


troyjr4103:
Composition record, for anyone reading this after the squash.

Four of the eleven branches merged here were CI-red on their own tips, and none of their pull request bodies said so. Each was verified on a `git archive` export of the branch with `origin/main` as the control, and repaired as a separate commit named for the pull request it repairs.

| PR | failing check | cause |
|---|---|---|
| 899 | Check Zero File-Search Invariant | `kin-daemon-spawn/lib.rs` gained the moved `OpenTransaction` filesystem calls with no allowlist pins, and the `commit_liveness.rs` pins went stale |
| 899 | Windows authority tests | `fn authority_change_count` in the `api.rs` tests is unused on Windows because its only caller is `#[cfg(unix)]`, so it is dead code under `-D warnings` |
| 902 | Check Runtime Boundaries | `SessionRegistry::new()` called from a new test in `handlers/entities.rs`, which is not an allowed file |
| 907 | Check Zero File-Search Invariant | `embed_model.rs`, five filesystem probes, no allowlist entry |
| 909 | Check formatting, then the zero-file-search guard once formatting was fixed | `health.rs` subprocess probe plus `is_file`/`is_dir`, and a stale pin |

Two more breaks existed only in the union and were folded into the merge commit that introduced each: after 903, `Envelope::workspace_mismatch()` lacked the `response: None` field 902 added (E0063); after 909, `health.rs` failed rustfmt at that branch's own tip. Every merge was taken with `rerere.enabled=false` so no recorded resolution could apply silently, and a multiset-union check over the twelve files touched by more than one branch came back exact.

Full local gate on the repaired tip: 6140 tests run, 6140 passed, 11 skipped, doctests ok, exit 0, resolved against this worktree at 7e6a2f65 and run twice. Formatting, the CI clippy allowlist, and every `ci.yml` script guard green.

Full record at `.kin-coord/reports/landwave-20260818.md`.

PR #914, opened 2026-08-18
DO NOT MERGE: falsify the PR-time install proof against the v0.5.38 coverage shape

troyjr4103:
Falsification branch for FIR-2412, opened only to make CI answer one question and closed as soon as it does. Do not merge, do not enqueue.

It takes the head of #911 and adds one step to the semantic retrieval proof that rewrites the locate capture into the shape v0.5.38 produced, `semantic_coverage.complete` false with `graph_bodies.with_body` 0 of 1, right after the proof captures it. Everything else is identical to #911.

If `Install Proof (PR) / ubuntu-latest` and `Install Proof (PR) Gate` both go red here, the pull-request proof can fail on the exact shape that burned v0.5.38. If they stay green, the job in #911 is decoration and must not be trusted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1385,3 +1385,157 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           slug: firelock-ai/kin
           fail_ci_if_error: false
+
+  # The public install proof is what caught v0.5.38, and it caught it on the
+  # tag, after the version was cut and the archives were built and published,
+  # because nothing before the tag ran it. These three jobs run that same
+  # reusable proof against the binaries this pull request builds, so the break
+  # is a red pull request instead of a burned tag.
+  install-proof-pr-build:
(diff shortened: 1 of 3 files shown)


troyjr4103:
Answered. The pull-request install proof does fail on the v0.5.38 shape.

Run 32153257872. `Install Proof (PR) / ubuntu-latest` failed at the step `Validate installed capability proof` with `Error: semantic kin locate did not prove complete coverage and return probe.py`, on this capture:

    {"supported":true,"indexed":4,"total":4,"pending":0,"complete":false,
     "graph_bodies":{"source_paths":1,"with_body":0,"gap_paths":1}}   files: ["probe.py"]

`probe.py` is still present, so the refusal came from `complete`, which is what a graph body gap clears. `Install Proof (PR) Gate` then failed with `the install proof failed against this pull request's binaries (failure); the same proof gates every release tag`.

Closing unmerged. The mechanism it proves lives in #911.

PR #917, opened 2026-08-18
Add kin agent, a Kin-native provider-neutral agent loop

troyjr4103:
Kin has no agent of its own today. Every agent story is "point someone else's CLI at our MCP server", and the strongest claim in the product, that an agent physically cannot answer from raw file search, is built out of one vendor's permission system: `crates/kin-cli/src/commands/assistant_adapter.rs:29` declares an `EnforcementTier`, and only the Claude adapter reaches `Enforced`, because it writes a settings file denying `Grep`, `Glob` and `Read` and installs a `PreToolUse` hook; the Codex adapter refuses outright rather than overclaim. That makes the guarantee rented rather than owned, and it cannot follow us onto a local model. This adds `crates/kin-agent`, a provider-neutral agent loop, plus a `kin agent` subcommand in kin-cli, so Kin ships its own agent and enforces its own rule in its own process. It speaks the OpenAI chat-completions wire format, which is what LM Studio, Ollama, llama.cpp, vLLM, OpenRouter and OpenAI all speak, so one adapter reaches a laptop model and a hosted one from the same two flags, and it drives Kin over MCP stdio rather than a translated CLI bridge, so it inherits the real tool registry, the `_kin` freshness envelope and the `negative` verdict instead of reimplementing them.

The policy is enforced by absence and by refusal. The belt is Kin's tools plus exactly two local tools, `edit_file` and `write_file`; there is no shell tool, no file-search tool and no file-read tool anywhere in the belt or the router, so there is nothing to fall back to, and a tool the model invents is refused by name with a refusal that says why. Both halves matter, because a hallucinated `bash` call that silently returned nothing would read to the model as a command that produced no output. When a result reports `safe_to_conclude_absent` false, the loop appends Kin's verdict and its named limiting factor to what the model sees and tells it to treat the answer as unknown, so an untrusted absence cannot become a claim that something does not exist; a degraded-coverage note is surfaced once per run rather than on every call. Every local edit is bracketed by `kin_transaction_begin` and `kin_transaction_commit` under a `kin_session_start`, and those lifecycle tools are stripped from the model's belt so the harness owns them. Envelope and verdict are read by piercing `content[0].text`, because the wrapper never carries them and a reader that looks at the wrapper's top level comes back empty for every key; a payload that cannot be parsed is reported unreadable rather than collapsed to a default.

Runs emit the Claude Code stream-json transcript the fleet's analyzers already read, with an RFC3339 timestamp on every record because per-call latency is derived by subtracting a `tool_use` stamp from its matching `tool_result`, and Kin tools named `mcp__kin__<tool>` so `usage.py` classifies them as MCP calls. A `kin-trace.jsonl` sidecar carries the envelope summary, the negative verdict, the policy decision and each edit's provenance, joinable on `tool_use_id`, so everything Kin-specific stays out of the stream and existing readers keep working untouched. Exit codes are the run's own taxonomy: 0 a final answer, 1 a harness error, 2 the tool-call budget spent, 3 the deadline expired, 4 the endpoint unreachable, 5 the MCP server failed, with a finer `stop_reason` separating an endpoint that was not there from a model that could not hold the tool-calling protocol, because a run that died the second way must never score as a task the toolset failed.

Provenance. The text-shaped tool-call readers for Qwen and Gemma follow the ones proven in kin-bench at `crates/kin-bench-engine/src/live/agent/openai_compat.rs:1547` and `:1378`, including Gemma's special quote tokens. kin-bench is founder-owned and stays private; this is a lift of that design into Apache-2.0 code that ships. They earn their place because a loop that cannot read those shapes stalls on every local model that answers in prose instead of filling the protocol field.

Because this is a subcommand of the `kin` binary, it ships in every release archive and through npm with no release-workflow change at all. Nothing in `.github/workflows/` is touched.

Evidence from a real run of the built binary, against a scripted OpenAI-compatible endpoint and a scripted MCP server on a two-file fixture repository. It exited 0 with `stop_reason` `final_answer`, 5 tool calls of which 3 reached Kin, 1 local edit and 1 refusal, over 6 turns. The edit landed: `src/backoff.py` gained the docstring directly under `def compute_retry_backoff`. The refusal held: the model asked for `bash` with `grep -rn compute_retry_backoff .` and got back "There is no tool named `bash` and it will not be run. This agent has no shell, no file-search and no file-read tool on purpose", with the trace row recording `surface: policy, policy: refused, is_error: true` and nothing executed. The untrusted absence was handed through as unknown: the model saw the empty payload followed by "[kin] This result is empty and Kin reports the absence CANNOT be trusted: markdown bodies are not indexed. Treat this as unknown rather than absent", and the result record counts `unsafe_absence_events: 1`. The edit was bracketed: the trace runs `kin_session_start`, the two reads, `kin_transaction_begin`, `kin_transaction_commit`, then the `edit_file` row carrying `provenance: {bracketed: true, closed_with: "kin_transaction_commit", closed_cleanly: true}`, then `kin_session_end`. Both fleet analyzers read that transcript unmodified and exited 0: `usage.py` reported `5 tool calls: mcp=3, other=2` with rows for `mcp/kin/semantic_locate n=2`, `mcp/kin/get_entity_source n=1`, `other/bash n=1 err=1`, `other/edit_file n=1`, and `analyze.py` reported 13 events, `mcp_calls: 3`, `errors: {bash: 1}` and per-tool latency percentiles. There is no `fallback` row in either, which is the policy showing up in the measurement rather than in a claim, because `usage.py`'s fallback classifier fires on rg, grep, git, find, ctags and ag inside a Bash call and on the Read, Grep and Glob tools, and none of those exist here to be called.

The two local-model runs the design calls for are not claimed here, and they are now the sibling lane's. `kin agent doctor` reaches LM Studio and reports "answered with 8 model(s)" with `qwen/qwen3.6-35b-a3b@q8_0` among them, then exits 5 on a deliberately absent MCP command, which is the contract's endpoint-good, MCP-bad case. The runs themselves are being driven by lane `kinagent-stranger`, which holds the GPU lock against this binary with reason "kin agent stranger arms, qwen then gemma", so its results are its own deliverable rather than a duplicate. What follows from that for this pull request: the Qwen and Gemma text-shape readers are proven by unit tests against captured shapes, not against those two models live, and that is a real gap rather than a formality.

Falsification. Every new guard was broken in turn, run, and restored, and each failed with its named assertion: the router returning `Route::Kin` for unknown names broke "the refusal must say why"; dropping the belt check in the Qwen reader let prose mint a call; disabling the negative branch broke "the model must be told the absence is untrusted"; making `begin_transaction` return no bracket dropped the observed MCP sequence to `["kin_session_start", "semantic_locate", "kin_session_end"]` against the expected five; removing the transcript timestamp broke "record without a timestamp"; swapping the Qwen and Gemma readers for each other broke each other's test; passing tool specs on the forced final turn broke "the forced final turn must offer no tools"; dropping the required-argument check let a malformed call through; letting `..` through let a path escape the repository; forcing `isError` false lost an MCP error. The restored tree re-ran green. That last step is not a formality: the first pass of the falsification harness restored each file from a backup, which gave it the backup's older mtime, so cargo skipped the rebuild and the restored run passed against a stale binary. The suite was re-run with a `touch` after every restore.

Two bugs were found and fixed in this branch, neither visible to the tests. `kin agent doctor` aborted with exit 101 and "Cannot drop a runtime in a context where blocking is not allowed", because the CLI dispatches inside its tokio runtime and a blocking HTTP client drops its own runtime there; the subcommand now runs on a plain thread outside it. And `KIN_AGENT_TEST_KIN_BIN`, the ignored integration test's binary variable, was missing from `crates/kin-core/env_var_allowlist.txt`, which the registry-completeness guard caught.

Working with Claude Code, Codex, Cursor and Gemini through `kin setup` and MCP stays first class. `kin agent` is the premier path, not the only one, and the README and CLI reference say so in those words.

Closes FIR-2413

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3045,6 +3045,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "kin-agent"
+version = "0.5.39"
+dependencies = [
+ "anyhow",
+ "chrono",
+ "reqwest 0.12.28",
+ "serde",
+ "serde_json",
+ "tempfile",
+ "thiserror 2.0.18",
+ "uuid",
+]
+
+[[package]]
 name = "kin-blobs"
 version = "0.1.5"
 source = "sparse+https://kinlab.ai/registry/cargo/"
@@ -3088,6 +3102,7 @@ dependencies = [
  "fs2",
  "hex",
  "keyring",
+ "kin-agent",
  "kin-blobs",
  "kin-buildinfo",
  "kin-context",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ members = [
     "crates/kin-buildinfo",
     "crates/kin-notify",
     "crates/kin-notifier-macos",
+    "crates/kin-agent",
     "crates/kin-cli",
     "crates/kin-daemon-spawn",
     "crates/kin-daemon",
@@ -141,6 +142,7 @@ kin-migrate = { path = "crates/kin-migrate" }
 kin-runtime = { path = "crates/kin-runtime" }
 kin-buildinfo = { path = "crates/kin-buildinfo" }
 kin-notify = { path = "crates/kin-notify" }
+kin-agent = { path = "crates/kin-agent" }
 kin-cli = { path = "crates/kin-cli", default-features = false }
 kin-daemon = { path = "crates/kin-daemon" }
 kin-daemon-spawn = { path = "crates/kin-daemon-spawn" }
--- a/README.md
+++ b/README.md
@@ -283,9 +283,32 @@ add local vector similarity over them, and confirm coverage with
 
 ## Works with your agent
 
-Kin ships an MCP server, so any agent that speaks MCP can read the graph.
-`kin setup --intent agent` configures every client it detects in one pass. These
-are the per-client one-liners when you would rather install Kin directly.
+Kin ships its own agent, and it is the path we recommend for agent work. `kin
+agent run` drives any OpenAI-compatible endpoint, so a local model in LM Studio,
+Ollama, llama.cpp or vLLM works from the same flags as a hosted one, and it
+reaches the graph over the same MCP server every other client uses.
+
+```sh
+kin agent run --task "Find where the retry backoff is computed and document it" \
+  --model qwen/qwen3.6-35b-a3b --base-url http://127.0.0.1:1234/v1
+```
+
+What makes it different from pointing another agent at the MCP server is that the
+rule is enforced inside the agent rather than borrowed from a vendor's permission
+layer. It has Kin's tools plus exactly two local ones, `edit_file` and
+`write_file`. There is no shell, no grep and no file-reading tool, so it cannot
+answer a repository question from raw file search, and a tool it invents is
+refused by name. When Kin reports that an empty result cannot be trusted, the
+agent is told the answer is unknown and given the named gap instead of concluding
+the thing does not exist. Every edit runs inside a Kin transaction under a Kin
+session, so the change carries provenance naming the agent. Run `kin agent doctor
+--base-url <url>` first to check both halves answer. See
+[the CLI reference](docs/cli-reference.md#kin-agent) for the full surface.
+
+Working with Claude Code, Codex, Cursor, Gemini and anything else that speaks MCP
+stays first class. `kin setup --intent agent` configures every client it detects
+in one pass. These are the per-client one-liners when you would rather install Kin
+directly.
 
 Claude Code, from inside a session:
 
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ reaches the graph over the same MCP server every other client uses.
 
 ```sh
 kin agent run --task "Find where the retry backoff is computed and document it" \
-  --model qwen/qwen3.6-35b-a3b --base-url http://127.0.0.1:1234/v1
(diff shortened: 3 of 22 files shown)


troyjr4103:
Two results from the sibling lane `kinagent-stranger`, which drove the local models against this belt, worth attaching here because they bear on the design rather than on the driver.

The policy was measured rather than asserted. Its 25-minute run of `qwen3.6-35b-a3b@q8_0` recorded raw-search fallbacks at **0**, noted in its own table as "impossible by belt", across 49 tool calls of which 33 went to Kin. Its report is at `.kin-coord/reports/kinagent-stranger-20260818.md`.

The transaction gap this branch names as a follow-up is real and has now been observed on a live run, not reasoned about. That run opened eight transactions, attempted two commits, had both refused with "Cannot commit transaction ...: exact repository commits reject empty transactions", aborted five, and landed zero changes in `kin log`. That is exactly the shape this branch predicts: the harness brackets a local edit with begin and commit so the session and the run are named, but it stages nothing, so the file change never becomes a graph-owned entity update. The fix is for `edit_file` to stage `{verb: "update", target: <entity>, body: <new source>}` through `kin_transaction_stage`, which the MCP schema already accepts. That lane tracks it as FIR-2417.

Also worth recording: that lane independently hit the tokio runtime panic and independently reached the same one-site remedy, a plain thread outside the CLI's runtime, which is commit `cb1a680e` on this branch. Their local patch can be dropped in favour of this head.

PR #918, opened 2026-08-18
Probe the documentation-only fast path context names

troyjr4103:
Throwaway probe, do not merge, will be closed once read.

This exists to answer one question with evidence rather than by reading the workflow: on a real documentation-only diff, do `Falsify guards`, `Feature permutation tests (ubuntu-latest)` and `Feature permutation tests (macos-latest)` appear in the check set with a skipped or successful conclusion, or do they not appear at all? A required context that never appears waits forever rather than passing, so the answer decides whether those three names can safely be added to the main ruleset.

It touches one markdown file and nothing else, so `Classify diff scope` should report `docs_only=true`.

### DIFF
--- a/docs/docs-only-probe.md
+++ b/docs/docs-only-probe.md
@@ -1,0 +1,4 @@
+# Docs-only classification probe
+
+Throwaway probe verifying that the documentation-only fast path still reports every
+required context name. Delete with the branch.


troyjr4103:
Answered. On this documentation-only diff all ten required contexts appear and are satisfied: `Feature permutation tests (ubuntu-latest)` and `(macos-latest)` report SUCCESS from the documentation-only twin, and `Falsify guards` reports SKIPPED, which satisfies a required check. Zero required contexts were absent. Closing; the probe file never lands.

PR #919, opened 2026-08-18
Register a repository with kin init so cross-repo commands can see it

troyjr4103:
Cross-repo commands returned nothing on every fresh install, and this makes them answer.

`kin_migrate::finalize::update_registry` was the only function that adds a repo to
`~/.kin/registry.toml`, and it had zero production call sites. An unfiltered grep across the
whole repository found only its own definition and its own test, plus the crate's re-export
of the function name. Nothing else touched it. No other production writer existed either:
`kin registry` has no add subcommand, `registry.rs` only cleans stale entries, and
`kin-core/src/registry.rs` only writes an empty default file.

The consequence ran the whole chain. `pin_registered_local_repository_authorities` loads the
registry, iterates `registry.repos`, and pins each sibling as a repository authority, which
is then registered in the spine. On a fresh install the registry is empty, zero siblings pin,
the spine holds only the current repo, and both `kin xref` and `kin deps` answer with nothing.

This was invisible on the maintainer's machine because a leftover `registry.toml` from an old
migration made broken cross-repo look like working cross-repo.

## What changed

`kin init` now registers the repository it just admitted, warn-and-continue on failure, in the
same shape as the existing `exclude_store_from_git` side effect so a registration failure can
never fail `kin init`. It reuses the existing `update_registry` rather than adding a second
writer, because a second writer is how this class of bug returns.

`kin deps` and `kin registry` text was updated in the same change. Five user-facing strings
and three test assertions asserted that nothing registers a repository, which this change
makes false. Copy that contradicts what this PR does is not a follow-up.

## Proof

A test under a scratch `KIN_REGISTRY_PATH` calls the real async `run()` entrypoint and reads
the registry back. It panics with `registry entries: []` against the unfixed code and passes
after. The registry path comes from `KIN_REGISTRY_PATH`, not `KIN_HOME`, so a test isolating
only `KIN_HOME` would have passed for the wrong reason.

End to end, two fixture repos each admitted by a separate `kin init` under a scratch home:
`kin deps` from the consumer reports `depends on (1) demo-provider-repo (cargo manifest)`, and
from the provider reports `consumed by (1) demo-consumer-repo`. `kin deps` only reads the
registry file locally, so the demo never touches the daemon or the embedding pipeline.

## What this PR does not claim

The local gate was green at 6141/6141 on the pre-rebase tip. This branch has since been
rebased onto a main that added `kin agent` (#917), 4854 insertions across 23 files, cleanly
with no conflicts. The full workspace gate has not been re-run against this exact tip, but
`kin-cli`'s own lib suite was, standalone: 1512/1512 passed, 0 failed, in 793s, on this exact
tip after the rebase. CI, which builds the PR merged into main, is still the first full
workspace re-run against the merged content.

`kin migrate` is a second live admission path with the identical gap. It uses a different
pipeline and calls none of `finalize.rs`'s three functions, all of which have zero production
callers. A repository admitted through `kin migrate` still will not register itself. That is
deliberately out of scope here and needs its own change.

`crates/kin-cli/src/commands/mcp.rs` carries the same stale claim in its `kin mcp` startup
message, the same shape this PR fixes elsewhere. Left untouched here since it sits next to
MCP surfaces another branch owns right now.

The registry id is the directory's basename, not a generated repository identity. Two on-disk
repos sharing a basename collide on that id and overwrite each other's entry on the second
`kin init`. This is how `update_registry` already worked before this change. Redesigning its
identity scheme is a separate piece of work.

Opened as a draft and handed to the `kin` merge captain. No merge-authority lock was taken and
nothing was enqueued.

### DIFF
--- a/crates/kin-cli/src/commands/deps.rs
+++ b/crates/kin-cli/src/commands/deps.rs
@@ -102,10 +102,11 @@ pub fn registered_id_for_path<'a>(
                 String::new()
             };
             anyhow::bail!(
-                "no registry entry records the repository at {};{named} no shipped command writes \
-                 one, so the scoped view cannot answer here: `kin registry` only lists and cleans, \
-                 and the migration writer that records an entry has no caller. Use `kin deps \
-                 --all` to list the repositories that are registered",
+                "no registry entry records the repository at {};{named} `kin init` registers a \
+                 repository when it runs, so this one either predates that registration or was \
+                 never initialized with `kin init` here; `kin registry` only lists and cleans \
+                 rather than adding an entry after the fact. Use `kin deps --all` to list the \
+                 repositories that are registered",
                 working_dir.display()
             )
         }
@@ -137,9 +138,11 @@ pub fn repo_dependency_view(registry: &KinRegistry, repo_id: &str) -> Result<Rep
         .ok_or_else(|| {
             anyhow::anyhow!(
                 "repository '{repo_id}' is not in the Kin registry, so no cross-repo dependency \
-                 records exist for it, and no shipped command writes an entry that would create \
-                 them: `kin registry` only lists and cleans. Use `kin deps --all` to list the \
-                 repositories that are registered"
+                 records exist for it. `kin init` registers a repository when it runs; this one \
+                 either predates that registration or was never initialized with `kin init` \
+                 here, and `kin registry` only lists and cleans rather than adding an entry \
+                 after the fact. Use `kin deps --all` to list the repositories that are \
+                 registered"
             )
         })?;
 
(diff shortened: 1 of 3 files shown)


troyjr4103:
Taken by the kin merge captain (session 856a484a, lease captain-finishkin2). All 33 checks green, 4 skipped, CLEAN. Enqueued as lane crossrepo3; it lands after the v0.5.40 release PR clears the queue, in the wave-2 landing with the other fix branches. No ticket line in the body: if a Linear ticket exists for the cross-repo empty-answer finding from the canon audit, add `Closes FIR-XXXX` before it is submitted so the queue's squash carries it.

PR #935, opened 2026-08-18
Route every absence through one gate, so a name filter cannot certify what a reference query refused

troyjr4103:
`semantic_search` could certify an absence because the absence gate was entered only by tools that declared a cross-file edge class, and `semantic_search` declares none. Every negative-capable tool whose answer carries a payload now passes through one gate that reads two independent declarations: the edge classes an absence is read off, and whether the absence is a claim about one language's extracted graph at all.

## Mechanism

`crates/kin-mcp/src/negative.rs` `edge_coverage_gap` opened with `if requested.is_empty() { return None; }`, and `absence_cross_file_classes` (`crates/kin-mcp/src/negative.rs:217`) returns an empty vector for `semantic_search`. So the whole gate, including the `reference_enrichment` half kin#905 added for FIR-2404, was skipped, and the verdict fell through to `Envelope::negative_trust(NegativeClass::Structural)` (`crates/kin-mcp/src/envelope.rs:939`), which answers `structural_authoritative` from daemon health alone.

That is the entire distance between the two calls the stranger run made minutes apart on expressjs/express in one session. `find_references` on `createETagGenerator` refused to certify with `reference_enrichment: "unsupported"` and `trust: "inconclusive"`, correctly. `semantic_search(query: "utils", kind: "module")` certified `safe_to_conclude_absent: true`, `trust: "authoritative"`, `semantic_coverage: null`, while `lib/utils.js` sat in the tree holding nine entities.

It was not a missing case in a table. It was the gate's entry condition, which made "declares no edge class" mean "needs no evidence" rather than "reads a different substrate".

## The fix

`absence_is_language_scoped` (`crates/kin-mcp/src/negative.rs:261`) is the second declaration, with a per-tool table and its exclusions in the doc comment. `absence_coverage_gap` (`:376`) is the one gate, renamed from `edge_coverage_gap` because it now gates more than edges: load-bearing classes (wording unchanged), a filter region the index never populated, and language resolvability. The enrichment reason splits on what the tool actually read, so a name filter is never handed the reference-edge sentence for a mechanism it never used.

`edge_coverage::observe_absence_scope` (`crates/kin-mcp/src/edge_coverage.rs:327`) is the observation for a claim that traverses no edge: the languages it spans, their enrichment, and the entity count of the filter's own region. It pays no witness scan, because that scan measures nothing those verdicts rest on. `semantic_search` (`crates/kin-mcp/src/handlers/entities.rs:99`), `find_dead_code_seeded` (`:2774`) and `graph_neighborhood` (`:3580`) publish it on the empty path only.

A negative also names the coverage its verdict rests on and recites that one (`coverage_basis` at `crates/kin-mcp/src/negative.rs:918`, `structural_coverage_clause` at `:946`).

## Tool by tool

| tool | negative kind | edge classes declared | gate before | gate after |
|---|---|---|---|---|
| `semantic_search` | `no_entity_match` | none | **no** | **yes** |
| `find_dead_code_seeded` | `no_seed_match` | none | **no** | **yes** |
| `graph_neighborhood` | `no_neighbors`, `focal_not_in_graph`, `no_traversal` | none | **no** | **yes** |
| `find_references` | `no_references` | the query's `relation_kinds` | yes | yes, unchanged |
| `bulk_check_references` | `reachability_verdicts` | the query's `relation_kinds` | yes | yes, unchanged |
| `trace_data_flow` | `no_flow` | calls, imports, references | yes | yes, unchanged |
| `semantic_locate` | `no_ranked_match`, `no_named_match` | none | no | no, by declaration |
| `dead_code` | `no_dead_code` | none | no | no, deliberately |
| `entity_history` | `no_history` | none | no | no, deliberately |

`semantic_locate` is the one that could not be routed honestly here. Its payload is built by the daemon's own locate route in `crates/kin-daemon/src/api.rs`, not by `handlers/entities.rs`, so it publishes no observation, and declaring a dependency it cannot back would make every empty ranking inconclusive on evidence nothing collected. It stays gated on complete embedding coverage, and its unnamed-ranking arm is never certifiable at any coverage. `dead_code`'s empty result is the inverse claim, so a gate there would be wrong-signed. `entity_history` reads change history.

`resolution_miss_for` (`crates/kin-mcp/src/negative.rs:1557`), which qualifies `impact_analysis` / `semantic_diff` / `semantic_review` `scope_not_resolved`, the `focal_not_resolved` arms of `find_references` and `trace_data_flow`, and `kin_provenance_query` `entity_not_resolved`, is unchanged. A resolution miss carries no payload at all, so there is nowhere to publish an observation without changing the error wire shape, and making the error body JSON would route it to `negative_for` instead, which returns `None` for it and would delete the negative entirely. That family needs an envelope-level language fact and is a separate change.

## Contract item 3

The ticket asks that `semantic_coverage: null` and `trust: "authoritative"` never co-occur. This PR implements the justification, not the literal rule, and the difference is worth stating.

`envelope.semantic_coverage` is null for every `semantic_search` call on the MCP path by construction: `Envelope::with_health` (`crates/kin-mcp/src/envelope.rs:811`) never sets it, and `with_payload_metadata` (`:856`) lifts it from a key the search payload does not carry. A blanket rule would make `semantic_search` unable to certify any absence ever, which makes contract item 4 unsatisfiable in production and would leave the Python fixture asserting behavior the product cannot produce.

Instead a negative names its basis and recites that coverage. The express sentence is gone in both directions:

- JavaScript: `...with graph coverage for JavaScript (this answer reads no edge class; no language-server adapter for it in this build) and degraded signals [edge_coverage:reference_enrichment_unsupported]. Absence is NOT authoritative...`
- Python: `...with graph coverage for Python (this answer reads no edge class; its language-server availability unprobed) and no degraded signals. Absence is authoritative...`

Under the old code that same Python payload read `...with semantic coverage unknown and no degraded signals. Absence is authoritative: safe to treat the target as genuinely absent/unused.` That exact string is what falsification F4 printed.

If the literal rule is wanted instead, it is a one-line change that turns `semantic_search` absences permanently inconclusive.

## Falsification

Eight falsifications, each applied to the committed code, run to completion, and reverted; full output and exit codes read from files, never through a pipe. Restored tree re-verified green before commit.

| # | what was broken | observed |
|---|---|---|
| F1 | enrichment gate no longer reaches existence claims | EXIT=101, the 4 enrichment guards failed |
| F2 | empty-region gate disabled | EXIT=101, exactly the 2 scope guards failed |
| F3 | gate entry reverted to the literal pre-fix condition | EXIT=101, 7 failed |
| F4 | a structural verdict recites embedding coverage again | EXIT=101, 1 failed, printing the express advice string verbatim |
| F5 | `semantic_search` publishes no observation | EXIT=101, exactly the 2 handler guards failed |
| F6 | `graph_neighborhood` publishes no observation | EXIT=101, 2 failed, including a pre-existing neighborhood test |
| F7 | the two siblings dropped from the language-scope map | EXIT=101, 4 failed |
| F8 | gate fires on every unprobed language, the over-firing degradation | EXIT=101, 27 failed, including the pre-existing FIR-2404 Python fixture |

F8 is the regression bar: a fix that degraded into marking everything inconclusive would show up as those 27.

## Gate

`cargo fmt --all -- --check` exit 0. Clippy exactly as `ci.yml` runs it, 45-lint allowlist under `RUSTFLAGS=-D warnings`, exit 0. `scripts/verify-zero-file-search.py`, `scripts/zero_file_search_guard.sh`, `scripts/check_runtime_boundaries.sh`, `scripts/check_private_repo_coupling.sh` and `scripts/test-private-repo-coupling.py` all exit 0.

`bin/kin-lane run absencegate kin -- bin/kin-dev local-test kin` exit 0, run after rebasing onto the `origin/main` carrying #931 and #932. `bin/kin-dev` does not honour `--no-fail-fast` (its `local-test` usage takes only `[--repo-dir <repo>=<path>]... [repo...]` and builds `DEV_TEST_ARGV=(nextest run --workspace)` with no pass-through), so the run count is the evidence nothing was hidden:

```
Summary [ 332.486s] 6302 tests run: 6302 passed (3 slow, 3 leaky), 12 skipped
kin-dev: local-test: kin  .kin-coord/worktrees/absencegate-kin @ bce63342 (chore/lane-absencegate-kin)
```

After the gate exited, `git diff --stat HEAD -- Cargo.lock .cargo/` is empty and `git ls-tree -r HEAD --name-only | grep -F '.cargo/'` prints only `.cargo/config.toml`.

Closes FIR-2430

### DIFF
--- a/crates/kin-mcp/src/edge_coverage.rs
+++ b/crates/kin-mcp/src/edge_coverage.rs
@@ -303,6 +303,71 @@ pub fn observe_cross_file_reference_coverage_for_languages_witnessed<S: EntitySt
     })
 }
 
+/// Observe the language scope an absence claim covers, for a tool that
+/// traverses no edge to make it.
+///
+/// `semantic_search`, `find_dead_code_seeded` and `graph_neighborhood` answer
+/// from the entity index and the walk, not from a cross-file reference class, so
+/// the witness scan above measures nothing their verdict rests on. What their
+/// verdict does rest on is which languages the claim spans and whether this
+/// build can resolve their programs, which is the one fact
+/// [`reference_enrichment`] already answers. Publishing it under the same key
+/// the reference tools publish means one gate reads one observation rather than
+/// two shapes drifting apart.
+///
+/// `classes` is deliberately empty rather than absent: this observation asserts
+/// nothing about cross-file edges, and a class map full of `unknown` would read
+/// as a scan that failed instead of one that was never the question.
+///
+/// `scope_entities` is the count of entities the query's own filter selects with
+/// its name pattern removed, so a kind-filtered absence can state the coverage
+/// of that kind. `None` leaves it out entirely, because a region nothing counted
+/// is unknown rather than empty. A resolved count of zero says the filter
+/// selected a region the index never populated, which is a fact about the index.
+pub fn observe_absence_scope(languages: &[LanguageId], scope_entities: Option<usize>) -> Value {
+    let mut observed: Vec<LanguageId> = Vec::new();
+    for language in languages {
+        if !observed.contains(language) {
+            observed.push(*language);
+        }
+    }
+    let language = if observed.is_empty() {
+        "no resolved language".to_string()
+    } else {
+        observed
+            .iter()
+            .map(|language| format!("{language:?}"))
+            .collect::<Vec<_>>()
+            .join(", ")
+    };
+    let mut observation = json!({
+        "scope": "absence_scope",
+        "language": language,
+        "requested_classes": Vec::<&str>::new(),
+        "classes": Map::new(),
+        "cross_file_classes": Vec::<&str>::new(),
+        "reference_enrichment": reference_enrichment(&observed),
+        "budget_exhausted": false,
+        "entities_examined": 0,
+        "scan": "skipped_no_edge_dependency",
+    });
+    if let Some(count) = scope_entities {
+        observation["scope_entities"] = json!(count);
+    }
+    observation
+}
+
+/// The distinct languages a resolved entity set spans, in first-seen order.
+pub fn languages_of(entities: &[Entity]) -> Vec<LanguageId> {
+    let mut languages: Vec<LanguageId> = Vec::new();
+    for entity in entities {
+        if !languages.contains(&entity.language) {
+            languages.push(entity.language);
+        }
+    }
+    languages
+}
+
 /// Whether every class the absence verdict rests on is already `present`.
 ///
 /// Narrowed by [`crate::negative::load_bearing_classes`] rather than by a rule
--- a/crates/kin-mcp/src/handlers/entities.rs
+++ b/crates/kin-mcp/src/handlers/entities.rs
@@ -36,7 +36,13 @@ On an empty result the response carries an additive `negative` object: its \
 (daemon-owned graph, initialized and loaded, holding entities, no degraded signals) \
 or merely \"not indexed yet\" — check it before treating \"none found\" as ground \
 truth. This filter reads the entity index rather than the vector index, so its \
-absence is gated on the graph being complete, not on embedding coverage.";
+absence is gated on the graph being complete, not on embedding coverage. An empty \
+answer also carries `edge_coverage`, naming the languages the filter's own scope \
+spans and how many entities it holds with the name pattern removed, and the absence \
+is NOT certified when that scope is empty or when its language is one this build \
+wires no language-server adapter for: nothing then resolves the program behind the \
+declarations, so a miss cannot separate a symbol the repository lacks from one the \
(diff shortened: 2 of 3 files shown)


troyjr4103:
Queue note, recorded by the captain so this PR carries its own history: this entry left the merge queue once at 00:27Z and re-entered immediately. The cause was benign batch invalidation rather than a check failure. Its first speculative batch (run 32200191536) was built on kin#933's head and was CANCELLED when kin#934 landed ahead of it; the batch was re-formed and the entry now builds on 38269dab1905e50e3939cbeed187945b9575d74b, kin#934's squash. No check on this PR failed at any point: 31 SUCCESS and 4 SKIPPED, all concluded.

PR #940, opened 2026-08-19
Build release-candidate archives without a tag

troyjr4103:
The release proof loop runs `bin/kin-release-preflight` on the composed build, then `bin/kin-stranger run --archive` on the release candidate in fresh containers, and only then lands and tags. The second step could not run before a tag existed.

Release-shaped archives came only from `release.yml`, which runs only on a `v*` tag push (`.github/workflows/release.yml:6`). So the archive stranger for v0.5.40 ran on the shipped linux-aarch64 archive after the cut rather than on a candidate, and `bin/kin-release-preflight` skipped its Linux leg on every run, printing "no Linux archive was given and this host cannot cross-build one, so the Linux leg (Debian 12 container + glibc floor) was skipped". A macOS host cannot cross-build a linux-aarch64 archive, so the Debian 12 container leg and the glibc floor guard that v0.5.38 shipped past never saw candidate bytes.

## What this adds

`.github/workflows/rc-build.yml` is a `workflow_dispatch` build of the same archive layouts `release.yml` builds, for an arbitrary `ref` input, uploaded as artifacts named like the release assets (`kin-linux-aarch64` carrying `kin-linux-aarch64.tar.gz`, its `.sha256` and its provenance manifest) with a 7 day retention, plus a job summary naming the built sha, the version, and each archive's sha256.

It creates no tag, publishes no release, publishes to no registry, and touches no promotion state. It declares no environment and reads no secret: `grep -c 'secrets\.'` over the file returns 0, and its `permissions:` block is `contents: read`. That is deliberate and load bearing. The note above `config` in `release.yml:36` says manual dispatch is disabled there because a dispatch can target branch-controlled workflow code, which must never inherit Apple or package-publish secrets. This workflow is dispatchable precisely because it can reach none of them.

Targets are linux-x86_64, linux-aarch64 and macos-aarch64. windows-x86_64 and macos-x86_64 are deliberately absent: neither host tool consumes them, so a candidate says nothing about either archive.

## Why a copy and not a workflow_call refactor

The DRY shape would be factoring `release.yml`'s `build` job into a reusable workflow. That job is 880 lines (`release.yml:376` to `release.yml:1243`) inside the one workflow whose mistakes cannot be repaired afterwards, because a tag run resolves its workflows from the tag. The refactor would also have to turn `environment: release` into an input and route six Apple secrets through `secrets:`, and none of it could be proven before it landed. The blast radius of getting that wrong is the next release.

So `rc-build.yml` copies the steps, and `scripts/check-rc-build-drift.mjs` holds the copy to the original. It does not compare loosely. It rebuilds what the shared block must be by taking `release.yml`'s build steps, dropping the six macOS signing and notarization steps, applying six declared textual deltas, and requiring byte equality, then compares the shared matrix rows. Each delta carries a `DELTA(` comment in `rc-build.yml` naming the difference and why: the `ref` input on checkout, the bounded `scripts/ci-apt-install.sh` call in place of an unbounded `apt-get` (FIR-2391), the version source for the KinNotifier bundle and the provenance manifest, a candidate manifest that carries no `release_tag`, and the artifact retention.

Wired into `ci.yml` as "Check the RC build mirrors the release build", with the test suite added to the release policy step's `node --test` list. `rc-build.yml` is also registered in `scripts/test-release-workflow-authority.py`, whose census is exact equality over every workflow path, job id and display name, plus the sha256 of its dynamic job name context.

## Falsification

`scripts/check-rc-build-drift.test.mjs` breaks the relationship 12 ways and requires the named complaint each time. All 12 pass, and the first case is the control that the check can pass on the tree as it stands:

```
# tests 12
# pass 12
# fail 0
```

The cases: a step added to `release.yml` is reported and names it; a shared step edited in either file is reported and names the step; an omitted signing step appearing in `rc-build.yml` is reported; an omission-list entry `release.yml` no longer carries is reported as stale; a candidate-only step missing from `rc-build.yml` is reported; a delta whose anchor stopped matching is reported by label; a matrix row that moved in `release.yml` is reported; and an rc-build target `release.yml` does not publish is reported.

Two parser bugs were found by those cases and fixed. The build job body used to absorb the comment block introducing `notarize_linux`, and a comment block above a step used to attach to the previous step. Both are now covered by their own tests.

The census registration was falsified by construction: before registering, `python3 scripts/test-release-workflow-authority.py` failed with "workflow-wide required-check producer census requires the exact reviewed workflow paths, job ids, and display names".

## Gates run locally

`actionlint` clean on `rc-build.yml` and `ci.yml`. It caught two real defects on the first pass: three shared steps read `matrix.skip_vfs`, which `release.yml` only gets from its windows row, and backticks inside single-quoted `printf` formats tripped SC2016. Both are fixed, the first by declaring `skip_vfs` on every row with the value those steps already saw.

- `node scripts/check-rc-build-drift.mjs` passes: 6 steps omitted, 2 candidate-only, 6 deltas applied
- `node --test` over the seven release scripts: 92 tests, 92 pass, 0 fail
- `python3 scripts/test-release-workflow-authority.py`: exit 0
- `python3 scripts/test-assertion-reachability.py`: exit 0
- `python3 scripts/test-homebrew-release-gate.py`: 49 passed
- `python3 scripts/verify-zero-file-search.py` and `bash scripts/zero_file_search_guard.sh`: exit 0

No Rust changed.

## First live dispatch

A `workflow_dispatch` workflow is dispatchable only once it exists on the default branch, so this one cannot be dispatched from this branch and the first live run happens after merge. The captain runs:

```
gh workflow run rc-build.yml --repo firelock-ai/kin -f ref=<sha>
gh run watch --repo firelock-ai/kin <run-id>
```

Then feeds the result to the preflight with `bin/kin-release-preflight --rc-run <run-id>`, which lands in a separate kin-ecosystem pull request. Until that first dispatch runs green, this workflow is proven only by `actionlint` and by the drift check that ties its steps to steps `release.yml` already runs on every tag.

Closes FIR-2444

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -964,6 +964,7 @@ jobs:
           node --test \
             scripts/check-glibc-floor.test.mjs \
             scripts/check-kin-vfs-compat.test.mjs \
+            scripts/check-rc-build-drift.test.mjs \
             scripts/check-release-version.test.mjs \
             scripts/extract-release-notes.test.mjs \
             scripts/prepare-release.test.mjs \
@@ -971,6 +972,15 @@ jobs:
           node --check scripts/check-release-version.mjs
           node --check scripts/prepare-release.mjs
           node --check scripts/release-intent.mjs
+
+      # rc-build.yml builds release-candidate archives from a copy of
+      # release.yml's build job, so the archive stranger and the preflight Linux
+      # leg can run before a tag exists. A copy nothing checks keeps reporting
+      # success while the original moves, and release.yml is the one workflow
+      # whose mistakes a later commit cannot repair.
+      - name: Check the RC build mirrors the release build
+        if: runner.os == 'Linux'
+        run: node scripts/check-rc-build-drift.mjs
 
       # release.yml refuses a release whose Kin lock resolves a different
       # kin-vfs-core than the pinned kin-vfs checkout builds, and that
--- a/.github/workflows/rc-build.yml
+++ b/.github/workflows/rc-build.yml
@@ -1,0 +1,847 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 Firelock, LLC
+
+name: RC Build
+
+# A release-candidate archive build that needs no tag.
+#
+# The release proof loop says: preflight the composed build, run the archive
+# stranger on the candidate, then land and tag. The second step could not run
+# before a tag existed. Release-shaped archives were produced only by
+# release.yml, release.yml runs only on a `v*` tag push, and a tag run resolves
+# its workflows from the tag, so nothing landed afterwards can repair it. The
+# archive stranger therefore ran on shipped bytes for v0.5.40, after the cut,
+# and bin/kin-release-preflight skipped its Linux leg with "no Linux archive was
+# given and this host cannot cross-build one" because a macOS host cannot
+# produce a linux-aarch64 archive.
+#
+# This workflow closes that gap. It builds the same archive layouts release.yml
+# builds, for an arbitrary ref, and uploads them as artifacts. It creates no
+# tag, publishes no release, publishes to no registry, and touches no promotion
+# state. It holds no credential: it declares no environment and reads no secret,
+# which is what keeps a dispatch of branch-controlled workflow code away from the
+# Apple and package-publish credentials release.yml gates behind `environment:
+# release` (see the note above `config` in release.yml).
+#
+# SOURCE OF TRUTH: .github/workflows/release.yml, job `build`. Every step below
+# is a byte-identical copy of that job's step of the same name, except where a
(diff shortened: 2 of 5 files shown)


troyjr4103:
Captain note: both queue ejections were the drift guard working. kin#942 moved release.yml's kin-vfs checkout to 3df11fcb inside the block rc-build.yml mirrors, so the merge-group build failed "the workflows in the tree agree" while the PR's own checks, built against the older main, stayed green. Rebased onto current main and re-copied the changed step plus the provenance expectation (2 occurrences of the old ref). Drift check now reports agreement (7 steps omitted, 2 candidate-only, 5 deltas applied), the drift suite passes 17 of 17, all policy suites 151 of 151, release-workflow authority exit 0, actionlint clean. The force push is the rebase plus one repair commit 3975d3a7.

PR #941, opened 2026-08-19
Make ten first-run and retrieval surfaces report what is actually true

troyjr4103:
Ten papercuts from the isolated stranger runs on the v0.5.40 candidate and the shipped
v0.5.39 bytes. Eight are fixed here, one was already fixed on main and is closed with
evidence, and two carry a named remainder stated below rather than quietly dropped.

Two of the tickets named a mechanism that turned out not to be the one producing the
defect, so those are worth reading first.

## Provenance does not survive a spawn (FIR-2434)

`crates/kin-cli/src/commands/resources.rs:781` already had the correct branch and a test
for it, so the render was never the bug. A kin binary that selects the default resource
profile writes it into its own environment, because that is where kin-infer and kin-db
read it, and every process it spawns inherits a set variable indistinguishable from an
export the operator typed. The CLI spawns the daemon, the daemon captures
`ActualResources`, and on a host where nobody set anything the daemon reported the ship
default as an operator override, eight lines above its own "Env overrides: none".

`crates/kin-cli/src/resource_profile.rs:44` now records the VALUE the default was written
for, not a bare flag. An inherited default is recognisable, and a value a caller
deliberately overrode on the way down (a benchmark harness pinning `proof`) no longer
looks like kin's own choice, because the marker and the active value disagree.

## Only the files with nothing to count reported a count (FIR-2427)

kin#924 fixed the header and the zero-denominator render and both held. The residual is
the number itself, and it is not a missing counter or a first-entity read. A probe through
the real `PythonAdapter` settles it: a Python file WITH calls emits the
call-extraction-incomplete marker, so `attach_file_reference_parse_counts` removes its
count; a Python file with NO calls emits no marker and records a truthful zero. Across a
repository the only files contributing a denominator are the ones that had nothing to
contribute, and the sum over them is a measured zero. Hence `calls 238/0` beside a
JavaScript control reading `calls 242/609 (39%)`.

`crates/kin-core/src/reference_coverage.rs` now names the three states through
`CallSiteMeasurement`, publishes no percentage for a partial measurement, and renders
"parse side measured on 1 of 3 files (5 sites there)".

## The rest

**FIR-2435.** `crates/kin-cli/src/commands/health.rs:868` offers the shim copy
`find_shim()` resolves as `cp <source> <dest>` and names the network installer only when
no local copy exists, saying which arm the reader got. It still never names
`kin doctor --fix`, which that constant's doc comment bars because the text is reprinted
in the post-fix manual-steps list.

**FIR-2436, part of.** `crates/kin-cli/src/commands/graph.rs:345` states the answer once
as `Repository coverage: N of M admitted files produced entities`, the supported-input
line says it is an upper bound rather than a second answer, and
`crates/kin-mcp/src/envelope.rs:210` carries `entity_count_scope` so an agent can
reconcile the daemon count against the CLI's without arithmetic. The green verdict beside
pending enrichment is unchanged, for the reason `graph_health.rs:487` already gives:
failing on absence would fail every early repository.

**FIR-2437.** `ResponseBudget` carries `envelope_reserve`, so a default budget's cut reads
"28000-character budget (the 30000-character default less a 2000-character envelope
reserve)". A caller-named budget still names one number, which is the control.

**FIR-2438.** Adding words to the archive touches three surfaces, because adding a member
without them breaks every update. `scripts/release-archive-shape.cjs` admits `README.md`,
`INSTALL.md` and `checksums-sha256.txt` on all three families,
`scripts/write-release-archive-docs.mjs` generates them and hashes every root file, both
packaging legs of the release workflow call it and assert the members, and
`crates/kin-cli/src/commands/update.rs` skips them by name while still aborting on
anything else. `scripts/install.sh` needed nothing, because it moves a fixed list of
binaries.

**FIR-2439, part of.** Two of three. `find_references` reports
`focal_resolution.same_name_candidates` and adds an `ambiguous_name` degradation when a
bare name resolved among several, copying the shape `trace_data_flow` already used.
`crates/kin-parser/src/adapter.rs:206` stops slicing a comment between a declaration and
its body into the extracted signature, in every language. The context-pack bookkeeping
hoist is not attempted: `source_provenance_fields` is a shared seam with callers across
four tools whose doc comment treats the exact per-row spelling as a contract, so hoisting
it changes the wire shape of every tool that uses it. It should be its own lane.

**FIR-2425.** `crates/kin-daemon-spawn/src/lib.rs` records how long a daemon took to
publish its port, and `daemon_startup_patience` starts at the CLI's own 300s default,
scales to four times a recorded cost, caps at 1800s, and honours
`KIN_DAEMON_READY_TIMEOUT_SECS`. `crates/kin-mcp/src/daemon_delegate.rs:1211` uses it for
both the port wait and the health poll, and a still-starting daemon is answered with
"retry this tool in a moment; do not restart `kin mcp start`", because restarting mid-boot
discards the startup in progress.

**FIR-2418.** No code change. kin#925 (d45598ca) already replaced the directory poll with
`wait_for_leased_capture_directory`, which waits for the lease itself, and the separate
lease assertion is gone. 40 consecutive passes against the reported 5 failures in 23.

**FIR-2408.** `crates/kin-mcp/src/budget.rs:271` named `impacted_entities`, a key an
`ImpactReport` has never serialized, so every rung of the ladder walked an absent array
and cut nothing on a 416,567-character response. It now names the four real buckets and
sheds `fingerprint` and `metadata` in compact mode through a new `bulk_keys` rung.

## Falsification

Every guard was broken and confirmed to fail with its named message, then restored.
Thirteen cases: the impact shape reverted to `impacted_entities`; `bulk_keys` emptied; the
reserve dropped from the disclosure; the percent guard removed; the partial render arm
removed; the measured-files counter widened; the signature comment clamp removed; the
archive skip list disabled; the profile marker read as a bare flag; the patience policy
pinned back to 15s; the coverage label renamed; the supported-input wording reverted; and
`same_name_entity_count` pinned to 1.

Two of those found real gaps in my own work rather than confirming it. The first pass of
the FIR-2427 percent guard did not fail when removed, because that fixture's partial sum
was zero and the percent declined for the wrong reason, so a second case with a nonzero
partial sum was added. The FIR-2439 focal report and the FIR-2436 coverage line had no
guard at all until falsification showed nothing failed, so both gained one.

The zero-file-search allowlist entry is falsified by its own history: the guard failed on
the two new boot-cost expressions before the pin and passes after.

## Measurement

`impact_analysis` driven through the real handler and the real budget on a 120-caller
fixture: 318,056 characters before, 28,838 after, against a 30,000 budget. This is not a
live MCP session against a converted repository; it proves the budget now reaches the
payload.

Closes FIR-2434
Closes FIR-2435
Closes FIR-2437
Closes FIR-2438
Closes FIR-2427
Closes FIR-2425
Closes FIR-2418
Closes FIR-2408
Part of FIR-2436
Part of FIR-2439


troyjr4103:
Captain note on the four silent ejections: every queue build failed the rc-build drift guard ("the workflows in the tree agree"), because this branch edits release.yml's packaging steps inside the block rc-build.yml mirrors and was rebased from before kin#940 landed the guard, so the PR's own checks never ran it. Rebased onto current main and mirrored the change: the archive documentation step and the doc-contents assertion now exist in rc-build.yml's Unix and Windows packaging steps too, which is where they belong, since a candidate archive is what the stranger scores; the one real difference, a candidate having no tag to read a version from, is recorded as two RC_VERSION deltas in the drift checker. Drift check reports agreement with 7 deltas, the drift suite passes 17 of 17, all policy suites 151 of 151, release-workflow authority exit 0, actionlint clean on both workflows. kin#949 had the same disease and the same cure (its ref move mirrored, pushed as c79e993e).

PR #951, opened 2026-08-19
Record call sites in relation evidence and paint them in kin refs

troyjr4103:
find_references answered with entity rows and no line numbers, so an agent that needed the call sites had to fall back to grep, which the brownfield stranger did in every A/B comparison. This records the call site of every parsed call at extraction time, threads it through the linker into relation evidence, and paints the sites in kin refs and the MCP reference surfaces.

Every language adapter records the call expression's span when it extracts a call relation (Python including decorator call sites, Rust, JavaScript, TypeScript, Go, Java, Kotlin, PHP, C, C++, Swift, HCL and the shallow-backed languages), the linker carries the span into relation evidence rather than dropping it, and kin refs prints a reference-sites clause per caller row. Tests cover the call-site line across both ingest arms, the language matrix, and span expectations; the order-independence assertion is restated as shape equality so it cannot pass by accident of ordering.

## Gates

Run by the captain from the lane worktree after taking the lane over, every status read raw. cargo fmt --check exit 0. Both guard scripts and verify-zero-file-search pass. Clippy exactly as ci.yml runs it under RUSTFLAGS=-D warnings, exit 0. Feature permutations pass: cargo nextest run -p kin-daemon --features gcs (908 tests passed) and cargo test -p kin-cli --no-default-features (12 passed). Full workspace gate before the rebase: 6294 tests run, 6293 passed, 1 failed; the one failure was commands::auth::tests::default_cli_actor_id_prefers_saved_credential_email, which is untouched by this diff, passed in isolation immediately afterward, and passed 10 of 10 isolated repeat runs, so it is a pre-existing concurrency flake and not this change. Rebased onto origin/main across 938 and 946 (two import-list conflicts resolved as unions, keeping FILE_PARSED_EXTERNAL_MODULE_IMPORTS_KEY beside RelationSite), then the full workspace gate re-run clean: 6405 tests run, 6405 passed, 12 skipped, doctests exit 0.

Closes FIR-1825

### DIFF
--- a/crates/kin-cli/src/commands/graph.rs
+++ b/crates/kin-cli/src/commands/graph.rs
@@ -1517,7 +1517,7 @@ mod tests {
             file_path: "src/app.rs".to_string(),
             entities: vec![caller.clone()],
             relations: vec![kin_parser::ExtractedRelation {
-            site: None,
+                site: None,
                 receiver: None,
                 call_shape: None,
                 kind,
--- a/crates/kin-cli/src/commands/graph.rs
+++ b/crates/kin-cli/src/commands/graph.rs
@@ -1517,6 +1517,7 @@ mod tests {
             file_path: "src/app.rs".to_string(),
             entities: vec![caller.clone()],
             relations: vec![kin_parser::ExtractedRelation {
+            site: None,
                 receiver: None,
                 call_shape: None,
                 kind,
--- a/crates/kin-cli/src/commands/refs.rs
+++ b/crates/kin-cli/src/commands/refs.rs
@@ -3,6 +3,7 @@
 
 use anyhow::{Context, Result};
 use kin_index::RelationResolution;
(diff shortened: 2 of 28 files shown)


troyjr4103:
Rebased across kin#950, which landed while this PR's checks ran and conflicted on the shared reference collector and the refs painter fixture. Both conflicts were resolved by composition rather than by taking a side: the collector now carries 950's receiver_name_guesses map beside this PR's edges_by_caller map with both insert sites, and the painter fixture's spanless caller keeps 950's fan-out confidence (so its row sits under the candidate heading) while carrying this PR's named-absence site clause, with the spanned caller carrying real site spans. Full workspace gate on the composed head: 6428 tests run, 6428 passed, 12 skipped, fmt clean. The earlier ubuntu failure on this PR was the FIR-2449 flake (pre-948 head); this push includes 948.

PR #954, opened 2026-08-19
Abandon v0.5.41: its frozen install proof fails on a projection recording the tag cannot receive the fix for

troyjr4103:
v0.5.41's release run 32237324302 failed its Public Install Proof on windows-latest, macos-latest and macos-15-intel, both attempts, same signature, before anything published. The cause is frozen into the tag: setup records nfs on macOS and projfs on Windows without engaging them, and the tag's own doctor reads the recording as projection_mode=misconfigured, which the proof's aggregate-health rule treats as fatal. The failure is deterministic on every fresh install of these bytes, so further retries cannot pass, and the repair (kin#953, merged as 0b8767c8 with a falsified guard) resolves from main while a tag run resolves everything from the tag.

The release train currently holds with marker tag_not_finalized ("highest tag v0.5.41 is not finalized GitHub Latest v0.5.40; release recovery owns this state"), and its own policy steps past a tag only on a reviewed abandonment record. This adds that record with all five required fields, and the entry is proven with the repo's own selector: `python3 scripts/select-admissible-release-tag.py` over the updated manifest skips v0.5.41 with the notice line and names v0.5.40 the highest admissible base, exit 0.

Channel state verified before writing the record: the only GitHub Release object for v0.5.41 is the unpromoted prerelease the workflow publishes before proof, `npm view @kinlab/kin version` prints 0.5.40, and the Homebrew formula reads 0.5.40. The ubuntu proof legs passed on both attempts, so the abandonment is scoped to what the tag actually cannot do.

Release policy suites on this branch: node --test over the eight release scripts, 151 pass, 0 fail; release-workflow authority exit 0.

Closes FIR-2454 is deliberately absent: that ticket closed with kin#953. This record is the sequencing half, and v0.5.42 supersedes the dead tag.


troyjr4103:
Superseded by #955, which carries the same v0.5.41 abandonment record and is already in the hosted merge queue with auto-merge armed. Exactly one record may land: scripts/select-admissible-release-tag.py refuses duplicate tag entries, and a refusal hard-stops the Release Train, the tag mint, and recovery. Closing this one to keep the file single-record; no content dispute, both PRs cite run 32237324302 and supersede to v0.5.42.

PR #957, opened 2026-08-19
Never certify absence of a thing from absence of a match

troyjr4103:
`semantic_search` reported `safe_to_conclude_absent: true` with "safe to treat the target as genuinely absent/unused" for a name the graph resolves, and `impact_analysis`, whose whole output is used/unused verdicts, carried no `negative` object at all. Both are fixed here.

## The call, and why nothing caught it

The v0.5.41 candidate stranger run recorded its first Kin call as `semantic_search(query: "request", kind: "method")` on psf/requests. Zero results, `trust: "authoritative"`, about `Session.request`, which is a method, is named `request`, and is in the graph.

kin#935 routed every empty path through one gate, and on this payload that gate worked correctly. `semantic_search` declares no cross-file edge class, so the class arm is skipped by design. The kind filter's scope held every method in the repository, so `absence_scope_empty` could not fire. Python is one of the two entries in `ENRICHABLE_LANGUAGES` (`crates/kin-core/src/reference_coverage.rs:42`), so `reference_enrichment` read `unknown` and the enrichment arm could not fire either. Every gate read a healthy substrate and every gate was right. This is not the unenrichable-language case the family's first three tickets were about.

The matcher is what missed, and no observation described it. `by_name_pattern` (kin-db `crates/kin-db/src/engine/index.rs:130`) collects exact-name and token-index hits and returns EARLY on any hit, never reaching its substring fallback; the filter predicate (`crates/kin-db/src/engine/graph.rs:10168`) then drops what survives by kind, language and role. A query whose name pattern did select candidates can answer zero because the narrowing filters removed all of them, and it certified because nothing in the payload said the name had matched anything.

## The fix

`crates/kin-mcp/src/handlers/entities.rs:117` asks the name's own side: on the empty path only, and only when a narrowing filter was applied, it re-runs the filter with `kinds`, `languages` and `roles` cleared and the name kept, publishing the count and the filters applied through `attach_name_filter_scope` (`crates/kin-mcp/src/edge_coverage.rs:361`). With no narrowing filter that query is this query, so nothing extra runs and nothing is published.

`absence_coverage_gap` (`crates/kin-mcp/src/negative.rs:497`) reads it. A nonzero count means the query observed absence of a MATCH, so `name_filter_narrowed_to_zero` fires, names the count and the filter that removed the candidates, and the answer reads inconclusive. `absence_coverage:name_filter_narrowed` joins `degraded_signals` on the same terms as the two gaps beside it.

The certified case stopped overclaiming too. A name filter reads the entity index and traverses no edge, which is exactly why `absence_cross_file_classes` gives it no class, so it has no basis for the word "unused". Its authoritative advice now speaks about declarations and sends the caller to `find_references` for the question about use. `find_references` keeps the original sentence, because it reads the edges that sentence is about.

`impact_analysis` gained a `RetrievalSpec` row (`field: "entity_impacts"`, `always: true`, `Structural`), so the generic annotator at `crates/kin-mcp/src/envelope.rs:1271` attaches the envelope on every response, and `crates/kin-mcp/src/handlers/review.rs:168` publishes the cross-file reference coverage its verdicts are read off, over the changed entities' languages with the weakest governing. Its declarations in `absence_cross_file_classes` and `absence_is_language_scoped` were already written and were waiting for the spec row. `always: true` matches `bulk_check_references`: an `entity_impacts` row carrying `consumer_count: 0` is the verdict a caller acts on whether or not the blast radius elsewhere is populated.

## Falsification

Six guards, each broken in turn, the named test re-run, full output and exit code read from a file rather than through a pipe. Every case exited 101 with the named test FAILED.

1. Gate arm disabled: `a_name_filter_narrowed_to_zero_certifies_nothing` failed reading `safe_to_conclude_absent: true`, `trust: "authoritative"`, which is the pre-fix bug reproduced.
2. Handler observation not published: the handler test failed with the `name_filter` key absent (`left: Null`, `right: 1`).
3. Consequence unscoped: the advice failed carrying the literal string the stranger quoted, "Absence is authoritative: safe to treat the target as genuinely absent/unused."
4. `impact_analysis` spec row removed: `negative_for` returned `None`, panicking on "impact verdicts are always qualified".
5. `impact_analysis` coverage not published: the handler test failed on a payload carrying no `edge_coverage`.
6. Disclosure label renamed: the test failed on the missing `absence_coverage:name_filter_narrowed`.

Controls in the other direction, which is the FIR-2404 regression bar: the pre-existing `a_resolvable_language_still_certifies_a_genuinely_absent_declaration` passed before and after, a kind-filtered query whose name matches nothing at all still certifies, and a query with no narrowing filter publishes no second count.

## Gates

All run from the lane worktree through `bin/kin-lane run matchabsence kin`, statuses read raw rather than through a pipe.

- `cargo fmt --all -- --check` rc=0, re-run rc=0 after the rebase.
- `cargo clippy --all-targets --all-features` with the exact 45-lint allowlist read out of ci.yml, under `RUSTFLAGS=-D warnings`: rc=0. The one warning in the log is the pre-existing `block v0.1.6` future-incompat note.
- All four ci.yml guard scripts rc=0: `verify-zero-file-search.py`, `zero_file_search_guard.sh`, `check_runtime_boundaries.sh`, `check_private_repo_coupling.sh`.
- `cargo nextest run --workspace --no-fail-fast` rc=0: 6459 tests run, 6459 passed, 12 skipped, zero FAIL lines, so the run never cancelled early. `cargo test --doc` rc=0.
- Feature permutations on the changed crate, `--all-targets` under `-D warnings`: `--no-default-features` rc=0, `--all-features` rc=0.
- Rebased onto origin/main at 64c7c8c5; `cargo nextest run -p kin-mcp` rc=0 afterwards, 526 tests run, 526 passed.
- `git ls-tree -r HEAD --name-only | grep -F '.cargo/'` prints only `.cargo/config.toml`, and `git diff origin/main --stat -- Cargo.lock .cargo/` is empty. `git diff HEAD~1..HEAD --numstat` lists only the five files above.

One leg could not run here and is not being claimed. `cargo check --locked --target x86_64-pc-windows-msvc -p kin-cli -p kin-daemon --no-default-features` fails on this host inside `ring`'s C build for want of an MSVC toolchain, before reaching any Rust, so it is a host limit rather than a result. The check that leg catches is platform-gated dead code, and this diff adds none: it introduces zero `cfg(` lines, against a positive control that finds the one already in the file.

## Post-967 rebase

The merge queue ejected this branch (run 32297411043): the speculative merge compiled #967's four-argument `negative_for` against this branch's three-argument test calls at entities.rs:4856/4884 and handlers/mod.rs:6257. Rebased onto main carrying #967 and passed the empty response-gap list at those three sites, exactly as #967's own test call sites do. kin-mcp's 542 tests pass beside the new language gates, including this branch's positive control that an unnarrowed nothing-matched absence stays certifiable; fmt and all six guard scripts pass on the rebased tip.

Closes FIR-2452

### DIFF
--- a/crates/kin-mcp/src/edge_coverage.rs
+++ b/crates/kin-mcp/src/edge_coverage.rs
@@ -74,6 +74,11 @@ pub const EDGE_COVERAGE_KEY: &str = "edge_coverage";
 /// instead of an unqualified "1".
 pub const REFERENCE_RESOLUTION_KEY: &str = "reference_resolution";
 
+/// Key inside the observation carrying what the query's NAME FILTER selected on
+/// its own, for an answer whose narrowing filters could have emptied it. Read by
+/// [`crate::negative`].
+pub const NAME_FILTER_KEY: &str = "name_filter";
+
 /// How many entities may have their relations read while looking for a witness.
 ///
 /// A healthy graph answers in single digits, so this bound is only reached on a
@@ -359,6 +364,38 @@ pub fn observe_absence_scope(languages: &[LanguageId], scope_entities: Option<us
         observation["scope_entities"] = json!(count);
     }
     observation
+}
+
+/// Record what the query's NAME FILTER selected on its own, beside the scope
+/// [`observe_absence_scope`] already measured.
+///
+/// The two counts answer different questions and an absence needs both. The
+/// scope count removes the NAME and keeps the narrowing filters, so it says
+/// whether the region the caller asked about is populated at all. This one
+/// removes the NARROWING FILTERS and keeps the name, so it says whether the name
+/// resolved to anything before those filters were applied.
+///
+/// Only the second can see the case FIR-2452 was filed for. On psf/requests,
+/// `semantic_search(query: "request", kind: "method")` answered zero while the
+/// scope held every method in the repository and Python is a language this build
+/// enriches, so every gate that existed read healthy and the answer certified
+/// `safe_to_conclude_absent: true` about a name the graph resolves. What
+/// actually happened is visible only from the name's own side: the store's
+/// pattern index returns its exact-name and token hits and returns EARLY on any
+/// hit, never reaching its substring fallback, and the kind predicate then
+/// removed every candidate it had returned. That is absence of a MATCH, and the
+/// observation has to carry it before the gate can refuse to call it absence of
+/// a THING.
+///
+/// `narrowed_by` names the filters that were applied, so the verdict can say
+/// which ones removed the candidates rather than reporting an unattributed miss.
+/// Attached only when a narrowing filter was actually applied: with none, this
+/// query IS the name query and a second count would restate the first.
+pub fn attach_name_filter_scope(observation: &mut Value, narrowed_by: &[&str], candidates: usize) {
+    observation[NAME_FILTER_KEY] = json!({
+        "narrowed_by": narrowed_by,
+        "candidates": candidates,
+    });
 }
 
 /// The distinct languages a resolved entity set spans, in first-seen order.
--- a/crates/kin-mcp/src/handlers/entities.rs
+++ b/crates/kin-mcp/src/handlers/entities.rs
@@ -4852,9 +4852,13 @@ mod tests {
             serde_json::json!(["kind"])
         );
 
-        let negative =
-            crate::negative::negative_for("semantic_search", &empty, &ready_daemon_envelope(2))
-                .expect("an empty search carries a negative");
+        let negative = crate::negative::negative_for(
+            "semantic_search",
+            &empty,
+            &ready_daemon_envelope(2),
+            &[],
+        )
+        .expect("an empty search carries a negative");
         assert_eq!(negative["safe_to_conclude_absent"], false);
         assert_eq!(negative["trust"], "inconclusive");
         assert!(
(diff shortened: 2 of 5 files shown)


troyjr4103:
Merge queue ejection record: this PR's merge group (CI run 32277567311, base d06e2b97) failed at Check & Test (ubuntu-latest) on a single test, mcp_status_and_locate_disclose_live_only_entities_and_then_stop_once_a_commit_records_them (live_graph_durability_disclosure.rs:331, 61.48s), and GitHub removed the entry at 17:00:54Z with nothing recorded here. The PR's own CI ran the same suite green, and the only base delta was the kin-actions pin bump (#963), which this test cannot observe, so this is read as the settle-race flake class pending a second observation. Resubmitting once; a repeat failure on the same test stops the line and reopens this as a real interaction.

troyjr4103:
Second ejection, same test, same line: run 32284566284 failed Check & Test (ubuntu-latest) on mcp_status_and_locate_disclose_live_only_entities_and_then_stop_once_a_commit_records_them (live_graph_durability_disclosure.rs:331) at 62.27s against 61.48s on the first ejection. Per the stop-the-line rule recorded above, this PR is NOT being resubmitted until the interaction is named. Working hypothesis under investigation: this PR's absence-gate change interacts with the post-#964 test-execution environment on a settle-race assertion; the PR's own green CI ran pre-#964. FIR-2465 is updated from flake to interaction.

troyjr4103:
Ejection root cause and fix: the failing merge group (run 32297411043) compiled #967's widened negative_for signature against this branch's three pre-967 test call sites. Rebased onto main and passed the empty response-gap list at entities.rs and handlers/mod.rs test sites, mirroring #967's own tests. kin-mcp suite 542/542, fmt and all six ci.yml guard scripts pass locally on the new tip 699a644f. Ready to resubmit after #962 lands per the agreed order.

PR #967, opened 2026-08-19
Wire JavaScript and TypeScript reference enrichment, provision the servers, and stop certifying an absence over a language nothing resolved

troyjr4103:
JavaScript and TypeScript reference enrichment is wired, the language servers behind Python and JavaScript enrichment can be installed from `kin setup` and `kin doctor --fix`, and both halves are proved end to end against real language servers on fixtures shaped like the losses the npm-0541 brownfield verdict recorded.

## The discrepancy the ticket asked to name first

The Aug-19 reading of `ENRICHABLE_LANGUAGES` was correct: it was `[Rust, Python]`, and Python was genuinely wired on both daemon enrichment paths. Enrichment simply never started. At `crates/kin-daemon/src/daemon.rs:1375` the daemon calls `kin_lsp::discovery::discover_servers()` once during startup and, on an empty result, leaves `state.lsp_enrichment_tx` as `None` for the life of the process, so no enrichment message is ever sent and no adapter is ever consulted. The container the stranger run used carried no language server at all. The Python adapter did not fail to start, it was never reached. So the Python losses were a provisioning gap and the JavaScript losses were a wiring gap on top of it.

## Wiring

`ENRICHABLE_LANGUAGES` now names Rust, Python, TypeScript and JavaScript. The daemon held two copies of its adapter map about three hundred lines apart, which is how a language could be missing from both without either site looking wrong; one function, `lsp_adapter_for`, now serves the incremental and sweep paths. Both TypeScript and JavaScript resolve to kin-lsp's `TypeScriptAdapter`, which is complete rather than a stub, and kin-lsp's own `KNOWN_SERVERS` already mapped both names onto `typescript-language-server`. kin-lsp needed no change.

The doc comment claiming the daemon wires an adapter for exactly those languages is now an assertion: a test compares the adapter map against the constant in both directions.

## Provisioning

`crates/kin-cli/src/commands/language_servers.rs` holds the install recipes. `kin setup` and `kin doctor --fix` detect the missing servers and install them with consent, prompting once per install command with the download disclosed, and doing nothing non-interactively unless `--install-language-servers` is passed. The doctor rows name the exact command instead of describing one. The installer re-probes `PATH` after a command exits zero, because an npm global prefix that is not on `PATH` installs successfully and leaves the binary unreachable. The restart advice names `kin daemon stop` rather than the obvious `kin daemon restart`, because `kin daemon` has only `status` and `stop`.

## The consequential part: this change broke the absence-trust gate, and the fix reaches Python

FIR-2404's third gate keyed on `reference_enrichment == "unsupported"`, the BUILD limit, and the doc on `load_bearing_classes` says why it was needed: the measured edge classes cannot separate an express-shaped JavaScript graph from a healthy Python one. Wiring JavaScript removes that discriminator, and the first full gate run caught the consequence. The express fixture went straight back to `"safe_to_conclude_absent": true`, `"trust": "authoritative"`, advice "safe to treat the target as genuinely absent/unused" over `createApplication`. That is the FIR-2430 defect.

The build limit was only ever one cause of the gate's own stated reason, that nothing resolves the program behind the parsed declarations. A wired adapter with no server installed leaves it exactly as unresolved, so the same hole was already open for Python: on a host with no pyright, a Python absence was being certified as authoritative on a graph that could not have produced a single reference edge. Wiring JavaScript made that visible rather than introducing it.

The observation now reports the host as well as the build, through the same `PATH` lookup and the same binaries table the doctor row and the install recipes read, and the gate blocks on both `unsupported` and `no_language_server` with wording that says which. It stays silent on `unknown`, which is what an answer whose focal never resolved reports; firing there produced "nothing established that a language server resolved no resolved language on this host" and displaced the real limiting factor.

**Behavior change worth a deliberate look:** on a host with the language server installed, nothing changes. On a host without it, Kin stops calling an absence authoritative for that language, Python and Rust included, where it previously did. That is stricter and it is the honest reading, but it is a change to a shipped trust surface. The alternative was shipping a known-false certification for JavaScript.

## Proof

`crates/kin-daemon/tests/lsp_reference_enrichment.rs` builds two fixtures and drives the real enrichment against real servers. Python: a call reaches `HTTPAdapter.send` through `self.connection`, with a second method named `send` on the mixin. JavaScript: a call reaches `router.handle` across a `require` chain, with a second `handle` in the calling file. Each fixture holds two entities with the same name and the assertion is that the edge lands on the right one, so a bare-name matcher has a fifty-fifty choice and no information to make it with. Both assert the relation classifies as `type_resolved`.

`crates/kin-cli/src/commands/refs.rs` closes the last hop: an `RelationOrigin::Lsp` edge reaches the reference surface as a resolved caller while a receiver-name guess beside it stays a candidate. It also pins a gap rather than papering over it. kin-lsp constructs every enrichment relation with `evidence: Vec::new()` (`src/enrichment.rs:218`, `:320`, `:426`, `:503`), so an LSP-resolved caller arrives with no source span and its `reference_lines` come back empty with `NoEvidenceSpan`. The test asserts that, and will fail in the right place when kin-lsp starts carrying spans.

When no server is on `PATH` the live tests skip loudly, naming the binary and the install command. `scripts/ci-install-language-servers.sh` installs pinned versions into a local prefix before the test step on Linux, so the proof actually runs; it warns and exits zero on a registry failure rather than failing the gate, and the skip stays visible as a run-summary annotation and a per-test line.

## Falsification

Five, each run to completion with full output and exit status read directly.

1. Removing JS and TS from `ENRICHABLE_LANGUAGES` with the adapter still wired: `typescript: daemon wires an adapter = true, ENRICHABLE_LANGUAGES says false`, exit 101.
2. Removing the JS/TS arm from `lsp_adapter_for` with the constant intact: the same test in the opposite direction, `daemon wires an adapter = false, ENRICHABLE_LANGUAGES says true`, plus two more, exit 101.
3. The same removal against the live JavaScript proof: `javascript has no adapter in this build, which contradicts ENRICHABLE_LANGUAGES`, exit 101.
4. Inverting the JavaScript proof to demand the same-named local `handle`: `FALSIFICATION: demanding the bare-name answer; got [Entity(EntityId(76037a27-...))]`, exit 101. Exactly one target came back and it was the cross-file one.
5. Removing the `no_language_server` arm from the trust gate: three tests fail and the express one prints the restored answer verbatim, `"safe_to_conclude_absent": true`, `"trust": "authoritative"`, exit 101.

All five reverted and the tree confirmed clean.

## Gates

`cargo fmt --all -- --check` exits 0. Clippy exactly as ci.yml runs it, `--all-targets --all-features` under `RUSTFLAGS=-D warnings` with the 45-lint allowlist, exits 0. `cargo nextest run --workspace --no-fail-fast --locked` reports `6493 tests run: 6493 passed (4 slow, 1 leaky), 12 skipped`, exit 0, with the live proof inside that total. `cargo test --doc --locked` exits 0. Six guard scripts pass: zero-file-search, the answer-modules variant, runtime boundaries, private repo coupling, hydration semantics, and the kin-vfs compat pin check.

`git ls-tree -r HEAD --name-only | grep -F '.cargo/'` prints only `.cargo/config.toml`. `Cargo.lock` carries one added line, `which 8.0.4` in kin-core's dependency list, reusing the crates.io entry kin-cli and kin-daemon already pull; no new package and no path source.

## Also corrected

`docs/language-support.md` promised LSP enrichment for eight languages when the daemon wired two. Go, Java, C and C++ have adapters in kin-lsp the runtime has never reached, so the table now says which are wired and which are not, and the section names the install commands and the startup-discovery behavior.

Not claiming: the brownfield A/B has not been re-run. That belongs to the first build carrying this. What is proved here is the mechanism the verdict blamed.

Closes FIR-2464

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1202,6 +1202,15 @@ jobs:
         with:
           tool: nextest
 
+      # Without a language server on PATH the reference-enrichment proof in
+      # crates/kin-daemon/tests/lsp_reference_enrichment.rs skips, so this step
+      # is what makes it run. It warns and exits 0 on a registry failure rather
+      # than failing the gate, and the skip is annotated on the run summary and
+      # printed per test, so it is never silent.
+      - name: Install language servers for the enrichment proof
+        if: runner.os == 'Linux'
+        run: bash scripts/ci-install-language-servers.sh
+
       - name: Test
         run: cargo nextest run --locked
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3205,6 +3205,7 @@ dependencies = [
  "tracing",
  "unicode-normalization",
  "uuid",
+ "which 8.0.4",
  "winapi-util",
  "windows-sys 0.61.2",
 ]
--- a/crates/kin-cli/src/commands/graph.rs
+++ b/crates/kin-cli/src/commands/graph.rs
@@ -277,31 +277,16 @@ fn embedding_coverage_is_measured(_graph: &kin_db::InMemoryGraph) -> bool {
     true
 }
 
-/// Language servers this build can actually enrich with, and the binaries that
-/// provide them.
-///
-/// Two facts have to hold together for a reference edge to exist: the daemon
-/// wires an adapter for the language, and a server binary is installed. The
-/// binary names mirror `kin_lsp::discovery::KNOWN_SERVERS` for exactly the
-/// languages `kin_core::reference_coverage::ENRICHABLE_LANGUAGES` names;
-/// probing them here rather than through `discover_servers` keeps the daemon's
-/// status path off a subprocess, since discovery runs `--version` on every
-/// server it finds.
-pub(crate) const LANGUAGE_SERVER_BINARIES: &[(kin_model::LanguageId, &[&str])] = &[
-    (kin_model::LanguageId::Rust, &["rust-analyzer"]),
-    (
-        kin_model::LanguageId::Python,
-        &["pyright-langserver", "pylsp"],
-    ),
-];
-
 /// Languages whose enrichment server is installed on this host.
+///
+/// The table this probes lives in `crate::commands::language_servers`, beside
+/// the command that installs each one, because a name that drifts between the
+/// probe and the advice produces a doctor row that reports a gap and a fix that
+/// does not close it. Probing here rather than through
+/// `kin_lsp::discovery::discover_servers` keeps the status path off a
+/// subprocess, since discovery runs `--version` on every server it finds.
 pub(crate) fn installed_language_servers() -> HashSet<kin_model::LanguageId> {
-    LANGUAGE_SERVER_BINARIES
-        .iter()
-        .filter(|(_, binaries)| binaries.iter().any(|binary| which::which(binary).is_ok()))
-        .map(|(language, _)| *language)
-        .collect()
+    crate::commands::language_servers::installed_language_servers()
 }
 
 /// The whole-graph relation totals the completeness section reports beside its
--- a/crates/kin-cli/src/commands/health.rs
+++ b/crates/kin-cli/src/commands/health.rs
@@ -2532,7 +2532,7 @@ fn coverage_unreadable(
             language_server_gap_detail(missing_servers)
         ),
     )
-    .with_manual_fix(LANGUAGE_SERVER_FIX)
+    .with_manual_fix(language_server_fix())
 }
 
 /// Turn the measurement into a verdict, split from its fetch so the rule is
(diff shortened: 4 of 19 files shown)


troyjr4103:
The macos-latest Feature-permutation red on head fefc49e5 (run 32293643724) is the known FIR-2449/FIR-2465 flake: mcp_status_and_locate_disclose_live_only_entities_and_then_stop_once_a_commit_records_them panicked at live_graph_durability_disclosure.rs:331 after the fixed 60s ambient-admission bound raced runner contention. 9 of 10 tests in the binary passed; nothing in this branch touches that path. Failed legs rerun; the bound itself is fixed by #969 (wait on the daemon's own record), which is in checks now.

PR #981, opened 2026-08-20
Decide every risk finding on the changed entity's own evidence

troyjr4103:
`assess_risk` walks the diff one change at a time, and every condition inside that loop read a
DIFF-GLOBAL bucket off the impact report. So a finding about one entity fired on another
entity's evidence. `shadow.rs` already states the rule this should have followed:

> another entity's consumers do not make this entity's surface change risky; the per-entity
> inbound attribution decides.

Six sites, one mistake, and it fails in BOTH directions, which is why this is one audit pass
rather than one bug.

The consumer conditions (signature change, contract modification, removal) fire when their
bucket is NON-empty, so diff-global INVENTED findings. That is the reported case: a removal
finding on an entity nothing depends on, because something else in the diff had dependents.

The coverage conditions (modified entity, new public entity) fire when their bucket is EMPTY,
so diff-global SUPPRESSED them. A modified entity with no tests of its own read as covered
because a neighbour in the same diff had one. That half hides risk instead of inventing it, and
nothing in the output says so, which makes it the worse of the two. It was not in the ticket.

Two of the six were spelled `impact.affected_x.iter().any(|_| true)`, which is `!is_empty()`
written so it does not read like the whole-bucket test it is. That spelling is part of why this
survived.

## What decides each condition now

Five conditions take the changed entity's own record. The all-zero fallback for a missing record
is conservative in both directions at once, which is what lets one accessor serve every
condition: zero consumers does not raise a finding on absent evidence, and zero covering tests
does report the gap on absent evidence. `analyze_impact` records a row for every changed entity,
so that fallback is reached only on paths that computed no impact at all, and those already
carry their own blocking evidence gap.

Removal gates on the UNION of two per-entity sources, because it has two and neither alone
covers the path the other serves. `removed_entity_dependents` reads the relations this diff
removed, which are exactly the edges that pointed at the entity, and it also builds the finding's
message, so gate and message can no longer disagree. The recorded per-entity count covers what
that cannot see: `analyze_impact` walks the live graph, where a removed entity is already gone,
so its dependents may arrive only through `shadow.rs`'s base-side overlay. Gating on the
relation walk alone silenced the unresolvable-record case, which is the one a reviewer can least
afford to lose. Neither source is the diff-global bucket, which is the point.

## After #978, the pre-fix defect is visible in a report

#978 rewrote the removal message without touching the gate. A spurious fire therefore takes this
path: the diff-global gate is true because another entity has dependents, `removed_entity_
dependents` correctly finds none for THIS entity, and the finding reads "Removed entity `X` still
has dependents" while naming none. This branch's own pre-fix run printed exactly that:

```
Removed entity `orphan_helper` (Function) at src/orphan.rs:10 still has dependents
```

with no dependents listed. A finding that asserts dependents and can list zero is the defect
wearing a label, and an auditor can spot it from a report alone.

## Fixtures, and why this is not fitting tests to the change

Every new case is a TWO-entity diff. A one-entity diff cannot tell diff-global attribution from
per-entity attribution, which is exactly why the existing tests passed under both readings.

Four new cases fail on the pre-fix code, in both directions, and two new controls pass and must
keep passing, so the set cannot be satisfied by a patch that makes the checks fire never or
always.

Seven existing fixtures gained the per-entity row that a real report always carries. Every
assertion in them is unchanged. Six went green on the fixture completion alone, which is what
shows they were about the entity's own evidence all along and only their fixtures were written in
the form that let a neighbour answer for it. `analyze_impact` records a row for every changed
entity, so a fixture naming a consumer in a global bucket with no per-entity trace is not a
smaller report, it is a shape the analyzer cannot produce.

The other two are the removal tests, and they are the reason the gate is a union rather than a
single source. They failed against the relation-only gate and they were right to.

One direction is pinned rather than left to be discovered. A removal consumed ONLY by a test
fires. `shadow.rs`'s BLOCKING gate keys on non-test consumers, and that is right for a gate,
since a test is not what decides whether a deletion is safe to merge. This is not that gate. It
is a risk finding, deleting a function a test calls does break the test, and #978 made the same
call in the message, where a test consumer is listed but sorts after production ones. The two
surfaces disagree on purpose and the disagreement is legible: the finding names the test, the
gate does not block on it. An unpinned choice there would be the same class of defect as the one
this change fixes.

`EntityImpact` deliberately does not derive `Default`, so the test helper writes every field out:
each is a measured count, and a field added later should force a decision rather than arriving
silently as zero in tests that are about which entity a count belongs to.

Closes FIR-2485

### DIFF
--- a/crates/kin-review/src/risk.rs
+++ b/crates/kin-review/src/risk.rs
@@ -4,11 +4,52 @@
 use std::collections::BTreeMap;
 
 use crate::diff::{EntityChangeKind, RelationChangeKind, SemanticDiff};
-use crate::impact::ImpactReport;
+use crate::impact::{EntityImpact, ImpactReport};
 use kin_model::entity::{Entity, EntityKind, EntityRole, Visibility};
 use kin_model::ids::EntityId;
 use kin_model::relation::GraphNodeId;
 use kin_model::review::{RiskLevel, RiskSummary};
+
+/// This entity's own inbound attribution, or an all-zero record when the report
+/// carries none for it.
+///
+/// FIR-2485. Every condition in [`assess_risk`]'s per-change loop used to read a
+/// DIFF-GLOBAL bucket off the report and attribute it to whichever change was in
+/// hand, so a finding about one entity fired on another entity's evidence. The
+/// principle is `shadow.rs`'s and is stated there: another entity's consumers do
+/// not make this entity's surface change risky, the per-entity inbound
+/// attribution decides.
+///
+/// It fails in BOTH directions, which is why one pass fixes six sites rather
+/// than five. The consumer conditions fire when their bucket is non-empty, so
+/// diff-global INVENTED findings. The coverage conditions fire when their bucket
+/// is empty, so diff-global SUPPRESSED them: a modified entity with no tests of
+/// its own read as covered because a different entity in the diff had one. The
+/// suppressing half hides risk rather than inventing it, and nothing in the
+/// output said so.
+///
+/// The all-zero fallback is conservative in both directions at once, which is
+/// what lets one accessor serve every condition: zero consumers does not raise a
+/// finding on absent evidence, and zero covering tests does report the gap on
+/// absent evidence. `analyze_impact` records every changed entity, so the
+/// fallback is reached only where a report was built without computing impact,
+/// and those paths already carry their own blocking evidence gap.
+fn per_entity(impact: &ImpactReport, entity_id: &EntityId) -> EntityImpact {
+    impact
+        .entity_impact(entity_id)
+        .cloned()
+        .unwrap_or_else(|| EntityImpact {
+            entity_id: *entity_id,
+            consumer_count: 0,
+            strong_consumer_count: 0,
+            proven_consumer_count: 0,
+            contract_consumer_count: 0,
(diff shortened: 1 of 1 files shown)


troyjr4103:
Ejected from the merge queue at 05:38 by an unrelated flake. Recording it here because an
ejection marks nothing on the PR it drops, so this one would otherwise read CLEAN and
MERGEABLE and simply never merge.

Merge-group run `32336337185`, job "Feature permutation tests (ubuntu-latest)", step "Test
featureless kin-cli":

```
commands::health::tests::a_driver_that_will_not_load_is_reported_as_broken_rather_than_absent
panicked at crates/kin-cli/src/commands/health.rs:4494:
  the loader's own words must reach the report: Text file busy (os error 26)
test result: FAILED. 1596 passed; 1 failed
```

Not this branch. Against its true merge base this PR touches one file,
`crates/kin-review/src/risk.rs`, and nothing in `kin-cli`. The failing test writes a driver
script, chmods it `0o755`, and executes it; `Text file busy` is `ETXTBSY`, the exec-after-write
race, which is environmental rather than a behavior change. It is the only failure among the
last 25 `ci.yml` runs.

Needs re-submission by the captain. No change required here.

PR #983, opened 2026-08-20
Record the receiver a JavaScript member call was written on

troyjr4103:
## The mechanism

`crates/kin-parser/src/languages/javascript.rs` constructed every extracted call with
`receiver: None`, unconditionally. Python and Rust populate that field; JavaScript never
has. Two things follow from it.

The linker loses the only signal that separates a call through an imported module from a
call through a value nothing here types, so a member call arrives as a bare leaf and the
name-matching tiers claim it. That is how `JSON.stringify(code)` at
`lib/response.js:68` acquired a `Calls` edge to express's own `stringify` at line 1026,
how `console.error(err.stack)` acquired one to a local `error` handler, how
`http.createServer(...)` acquired one to a test file's own `createServer`, and how
`arr.forEach(...)` acquired one to axios's exported `forEach`.

And FIR-2486's unresolved-receiver tier, which this branch's first three commits build,
cannot fire for JavaScript at all. A call with no receiver gives the tier nothing to
record, so `this.router.handle(req, res, done)` at `lib/application.js:177` was dropped
rather than carried as an unproven step, and `trace_data_flow` on `app.handle` returned
a walk that simply did not contain the line the function exists for.

## The fix

`extract_calls_from_context` now returns the callee name together with a receiver.

An owner-folded call keeps no receiver. `this.m()` inside a method and `Owner.m()`
naming the enclosing owner are already recorded as `Owner.m`, so the owner is in the
name, and repeating it as a receiver would make the linker skip the same-file tier that
name exists to reach. This is the split Python's adapter makes for `self` and `cls`.

Every other member call carries the object's source text, but only when that text is one
unbroken dot-separated name chain. The field's consumers read it as a name: the linker
splits it at the first `.` and asks the calling file's imports about the root, and the
placeholder tier rejects anything holding whitespace. A receiver that is itself a call,
a subscript, a parenthesized expression or a chain broken across lines answers neither
question, so it is declined rather than stored as raw source text. `this` is kept rather
than dropped, because it is the root of the property chain the destination hangs off and
`this.router` is the receiver the express hand-off is actually written on.

TypeScript needs no separate change and gains the same behavior: `typescript.rs` imports
this function from the JavaScript adapter and calls it at every site that extracts a
call body. The `receiver: None` sites remaining in `typescript.rs` are `Contains`,
`Extends` and `Implements` relations, which have no receiver. One of the new tests pins
the shared path so the two languages cannot drift apart silently.

## Measurement, two corpora asked for and a third for breadth

`receiver_is_object` gates several linker tiers, so populating the field changes which
tier resolves every JavaScript member call. Both arms ran at binaries identical but for
this one file, each sha256-checked. Corpora are pinned by commit:

- `expressjs/express` `a3714473feb3d2908add734d340e7755fd85e0a3` (tree `134de344`), the
  brownfield bar's own pin
- `expressjs/body-parser` `b1f3d588c5dca4b3a4ca7d5d3765c9129118926a` (tree `b77e37b2`)
- `axios/axios` `84a9f3b9a4f3244b8c8e818f557d64c7b964fb25` (tree `53995167`), added
  because body-parser is 13 JavaScript files and axios is 192 plus 25 TypeScript

Call edges by resolution:

| corpus | type_resolved | import_scoped | name_only | total |
|---|---|---|---|---|
| express | 78 to 70 | 0 to 4 | 188 to 190 | 266 to 264 |
| body-parser | 24 to 20 | 0 to 0 | 26 to 45 | 50 to 65 |
| axios | 230 to 227 | 0 to 110 | 387 to 682 | 617 to 1019 |

Call edges by tier, read off the confidence each tier stamps:

| tier | express | body-parser | axios |
|---|---|---|---|
| 1.00 same-file | 72 to 64 | 16 to 12 | 188 to 185 |
| 0.95 import-based cross-file | 6 to 6 | 8 to 8 | 41 to 41 |
| 0.90 receiver-module | 0 to 4 | 0 to 0 | 0 to 110 |
| 0.85 inherited method | 0 to 0 | 0 to 0 | 1 to 1 |
| 0.70 cross-file name match | 35 to 1 | 16 to 0 | 187 to 49 |
| 0.30 bare-name fan-out | 107 to 25 | 0 to 0 | 169 to 49 |
| 0.20 external import reference | 46 to 46 | 10 to 10 | 31 to 31 |
| 0.10 unresolved-receiver placeholder | 0 to 118 | 0 to 35 | 0 to 553 |

Tier migration, counting each `(caller, destination)` call edge once:

| corpus | edge gone entirely | edge new entirely | resolution changed |
|---|---|---|---|
| express | 123 | 121 | 2 |
| body-parser | 20 | 35 | 0 |
| axios | 141 | 555 | 110 |

All 112 migrations are upgrades from `name_only` to `import_scoped`. Nothing migrated
downward.

## Finding: receiver shadowing, a fabrication class that was never filed

Every proven edge this change removes belongs to one unnamed defect, and naming it is
worth more than the fifteen instances. When a call is written on a builtin, a Node core
module or any foreign object, and this repository happens to define a function with the
same leaf name, the bare-leaf tiers bound the call to the local definition. The result
is a false map with no disclosure on it: `res.status` reads as a caller of express's
`stringify`, and `stringify` reads as calling itself, when the source says
`JSON.stringify`. Nothing in the answer said the receiver was never consulted, because
nothing recorded the receiver at all.

Recursion is the sharpest shape of it. Nine of the fifteen are self-edges, where a
function's own definition captured a call it makes on something else entirely, so a
reader tracing the function saw a recursive call the source does not contain.

All fifteen, with the receiver that was ignored and the line it was written on:

| corpus | caller | bound to | actual call site | receiver ignored |
|---|---|---|---|---|
| express | `res.status@lib/response.js` | `stringify@lib/response.js` | `JSON.stringify(code)` at `lib/response.js:68` and `:72` | `JSON` |
| express | `res.cookie@lib/response.js` | `stringify@lib/response.js` | `JSON.stringify(value)` at `lib/response.js:755` | `JSON` |
| express | `stringify@lib/response.js` | itself, at `:1026` | `JSON.stringify(...)` at `lib/response.js:1030` and `:1031` | `JSON` |
| express | `error@examples/error/index.js` | itself, at `:20` | `console.error(err.stack)` at `:22` | `console` |
| express | `format@examples/content-negotiation/index.js` | itself, at `:33` | `res.format(obj)` at `:36` | `res` |
| express | `json@examples/content-negotiation/users.js` | itself, at `:17` | `res.json(users)` at `:18` | `res` |
| express | `count@examples/view-locals/index.js` | itself, at `:48` | `User.count(...)` at `:49` | `User` |
| express | `count2@examples/view-locals/index.js` | `count@...index.js:48` | `User.count(...)` at `:87` | `User` |
| body-parser | `createServer@test/json.js` | itself, at `:766` | `http.createServer(...)` at `:771` | `http` |
| body-parser | `createServer@test/raw.js` | itself, at `:503` | `http.createServer(...)` at `:508` | `http` |
| body-parser | `createServer@test/text.js` | itself, at `:573` | `http.createServer(...)` at `:578` | `http` |
| body-parser | `createServer@test/urlencoded.js` | itself, at `:972` | `http.createServer(...)` at `:977` | `http` |
| axios | `isBuffer@lib/utils.js` | itself, at `:221` | `val.constructor.isBuffer(val)` at `:228` | `val.constructor` |
| axios | `trim@lib/utils.js` | itself, at `:472` | `str.trim()` at `lib/utils.js:473` | `str` |
| axios | `toObjectSet@lib/utils.js` | `forEach@lib/utils.js` | `arr.forEach(...)` at `:916` | `arr` |

Every line above was read in the pinned tree rather than inferred. The four
body-parser helpers sit at different offsets in their files, which is why they are
listed separately rather than collapsed to one line number.

Python already had the defence: `is_unbound_python_builtin_call` refuses to let a bare
call to a name in the builtins table be answered by matching that name elsewhere in the
repository. JavaScript had no equivalent and no receiver to build one from. Recording
the receiver removes the whole class rather than the fifteen cases, because the
receiver-scoped tier now answers before any bare-leaf tier is reached: a receiver bound
to a module outside the tree yields no local edge at all, and a receiver whose type
nothing here settles skips the tiers that match on leaf name alone.

The four new express edges at 0.90 are correct and new: `count`, `count2`, `users` and
`users2` now bind `User.count` and `User.all` in `examples/view-locals/user.js` through
the `require('./user')` the file writes. The 110 axios ones are the same shape at scale,
binding `lib/utils.js` exports through `import utils from "../utils.js"`.

## What unproven recall costs

express loses 123 call edges and gains 121. Of the 123: 51 have a destination in a
different example or test file than the caller, and those files import nothing from each
other, so each is a bare-name fabrication; 8 are the same-file ones above; 64 land in
`lib/`. Of those 64, 45 have a caller in `test/` or `examples/` calling `res.send`,
`app.use`, `res.status` and friends, which are semantically right and now unbound, and
that is the honest cost. The remaining 22 run `lib/` to `lib/` and split four correct
from eighteen fabricated; the eighteen include seven `req.*` getters bound to `res.get`
for `this.app.get('trust proxy fn')` and `logerror` bound to both `req.get` and
`res.get` for `this.get('env')`, which is exactly the fabrication family FIR-2464 named.

The four correct losses are `createApplication` calling `app.handle` and `app.init`,
`res.format` calling `req.accepts`, and `res.sendFile` calling `this.app.enabled`. All
four have a receiver that is a local variable or property holding an in-repo export,
which needs a type answer this repository does not carry. That is FIR-2487's route.

Kin does not count `name_only` as evidence a destination is used, so nothing that counts
moved for the worse, and the proven set got strictly more correct.

## Fabrication controls

706 unresolved-receiver placeholders appear across the three corpora. Every one reads
`name_only` at confidence 0.10, and none has a destination id colliding with a real
entity in its corpus, checked directly rather than assumed.

## Falsification

Four tests were added to `crates/kin-parser/tests/js_ts_call_resolution.rs`, with the
shared fixtures extended by an `Application` class carrying the express hand-off shape, a
subscript receiver and a call receiver. Each was falsified by breaking the behavior it
guards, reading full output and exit codes rather than a pipe:

1. Reverting the change: `js_member_call_records_the_receiver_it_was_written_on` and
   `ts_member_calls_record_receivers_through_the_same_constructor` fail with
   ``` `console.log(...)` must record `console` as its receiver ```, left `None`.
2. Giving the owner-folded branch a receiver:
   `js_owner_folded_and_bare_calls_record_no_receiver` and the TypeScript test fail with
   ``` `this.sayHi()` is recorded owner-qualified, so it carries no receiver ```, left
   `Some("this")`.
3. Dropping the name-chain filter: `js_receiver_that_is_not_a_name_chain_is_declined` and
   the TypeScript test fail with ``` `a.b().c()` is written on a call, which names no
   binding ```, left `Some("a.b()")`.

Falsification 1 alone leaves the two negative tests green, because with no receivers
anywhere "no receiver" is trivially true. That is why 2 and 3 exist: a negative assertion
has to be falsified against a wrong version of the change, not against its absence.

The four pre-existing tests asserting an unresolvable receiver binds nothing were not
touched and still pass.

## Replay semantics

`HYDRATION_SEMANTICS_VERSION` moves with this change. The branch's third commit carries
8 to 9 with the manifest digest, and the parser change lands in the same squash, so one
bump records one semantic epoch rather than two. The guarded surface is `history.rs` and
`semantic_import.rs` functions pinned by digest, none of which this change touches, so
`verify-hydration-semantics.py` passes unchanged.

## Acceptance, measured end to end and falsified

`bin/kin-brownfield-repro` was run twice against the same pinned fixtures, on binaries
that differ only in `crates/kin-parser/src/languages/javascript.rs`. Nothing else about
the two runs differs, and only one check moves.

With the parser file at its pre-change state, check 4 fails:

> CHECK 4 FIR-2464 FAIL the hand-off this.router.handle at lib/application.js:177, the
> last line of app.handle and the edge the question is about, is absent from the walk;
> steps are ['app.enabled', 'req.get', 'res.get', 'create', 'users.get', 'logerror',
> 'app', 'finalhandler', 'app.set', 'req', 'res', 'message']

With the change, check 4 passes on all three of its assertions:

> CHECK 4 FIR-2464 PASS the hand-off router.handle is in the walk
> - no fabricated callee is counted (8 of 18 steps counted)
> - both real callees app.enabled, finalhandler are present
> - the hand-off router.handle is in the walk

The bar's `counted_step` excludes any step whose resolution reads `name_only`, so the
placeholder is in the walk and outside the count, which is the shape the ticket asks
for: present with an honest label, never claimed as resolved.

Every other check is identical across the two arms. Diffing the verdict lines returns
exactly one difference, check 4. Checks 2 and 3, the fabrication controls, pass in
both: "no known bare-name impostor counted as a caller" and
"HTTPDigestAuth.handle_401 counted with reference_lines [312]". Check 7 fails in both
and is FIR-2487's ticket, untouched here. Full run: 8 pass, 1 fail, 0 unreadable.

## Gates

- `cargo nextest run --workspace --no-fail-fast --locked`: 6587 tests run, 6587 passed,
  3 slow, 2 leaky, 12 skipped. The full suite ran; nothing was cancelled early.
- `cargo test --doc --locked`: pass.
- `cargo fmt --all -- --check`: pass.
- `cargo clippy --all-targets --all-features` under `RUSTFLAGS=-D warnings` with the
  45-lint debt allow-list from `.github/clippy-allowed-lints.txt`: pass.
- All six `ci.yml` guard scripts pass: `verify-zero-file-search.py`,
  `zero_file_search_guard.sh`, `verify-hydration-semantics.py`,
  `check_runtime_boundaries.sh`, `check_private_repo_coupling.sh` and
  `test-private-repo-coupling.py`.
- `git diff origin/main --stat -- Cargo.lock .cargo/` is empty and
  `git ls-tree -r HEAD --name-only` lists only `.cargo/config.toml` under `.cargo/`.

Run from `.kin-coord/worktrees/jsreceiver-kin` on branch `chore/lane-jsreceiver-kin`,
four commits on `origin/main` at 93f31671.

Closes FIR-2486

### DIFF
--- a/crates/kin-core/src/ref_view.rs
+++ b/crates/kin-core/src/ref_view.rs
@@ -661,6 +661,26 @@ where
 
     parsed_relations.extend(projection_relations);
     for relation in parsed_relations {
+        // A placeholder edge names a destination this repository does not
+        // define, so nothing in the snapshot backs it and admission rejects the
+        // whole view with "unadmitted destination endpoint". Materialize the
+        // target the same way historical delta derivation does, from the
+        // importing entity's language, since a target defined elsewhere has
+        // none of its own.
+        if let Some(destination) = relation.dst.as_entity() {
+            if !snapshot.entities.contains_key(&destination) {
+                let language = relation
+                    .src
+                    .as_entity()
+                    .and_then(|id| snapshot.entities.get(&id))
+                    .map(|entity| entity.language);
+                if let Some(target) = language
+                    .and_then(|language| kin_index::placeholder_target_entity(&relation, language))
+                {
+                    snapshot.entities.insert(destination, target);
+                }
+            }
+        }
         snapshot.relations.insert(relation.id, relation);
     }
 
--- a/crates/kin-index/src/history.rs
+++ b/crates/kin-index/src/history.rs
@@ -24,8 +24,9 @@ use sha2::{Digest, Sha256};
 use crate::classifier::{FileClassification, FileClassifier};
 use crate::error::{IndexError, Result};
 use crate::linker::{
-    is_external_import_placeholder, link_cross_file_borrowed_with_completeness,
-    ArtifactIdentityMap, FileParseCompletenessMap, FileParseData,
+    is_external_import_placeholder, is_unresolved_receiver_placeholder,
+    link_cross_file_borrowed_with_completeness, split_unresolved_receiver_token,
+    unresolved_receiver_display_name, ArtifactIdentityMap, FileParseCompletenessMap, FileParseData,
 };
 use crate::pipeline::IndexPipeline;
 
@@ -448,7 +449,59 @@ fn external_reference_targets(
             .push(source.language);
     }
 
+    // The second placeholder class: a member call whose receiver resolves to
+    // nothing this repository defines. It carries no import source, because
+    // there is no module to name, so its identity and its display name come
+    // from the member expression the evidence recorded as written.
+    let mut receivers: BTreeMap<EntityId, (String, Vec<LanguageId>)> = BTreeMap::new();
+    for relation in linked {
+        if !is_unresolved_receiver_placeholder(relation) {
+            continue;
+        }
+        let Some(destination) = relation.dst.as_entity() else {
+            continue;
+        };
+        if entities.contains_key(&destination) {
+            continue;
+        }
+        let (Some(source), Some(token)) = (
+            relation.src.as_entity().and_then(|id| entities.get(&id)),
+            relation
+                .evidence
+                .first()
+                .and_then(|evidence| evidence.token.as_deref()),
+        ) else {
+            continue;
+        };
+        let Some((receiver, symbol)) = split_unresolved_receiver_token(token) else {
+            continue;
+        };
+        receivers
+            .entry(destination)
+            .or_insert_with(|| {
+                (
+                    unresolved_receiver_display_name(receiver, symbol),
+                    Vec::new(),
+                )
+            })
+            .1
+            .push(source.language);
+    }
+
     let mut targets = BTreeMap::new();
+    for (destination, (display_name, languages)) in receivers {
+        let fingerprint = fingerprints
+            .entry(destination)
+            .or_insert_with(|| external_reference_fingerprint("", &display_name))
+            .clone();
+        let Some(language) = lowest_language(&languages) else {
+            continue;
+        };
+        targets.insert(
+            destination,
+            external_reference_entity(destination, &display_name, language, fingerprint),
+        );
+    }
     for (destination, (import_source, symbol, languages)) in importers {
         let fingerprint = fingerprints
             .entry(destination)
--- a/crates/kin-index/src/history.rs
+++ b/crates/kin-index/src/history.rs
@@ -49,7 +49,7 @@ use crate::pipeline::IndexPipeline;
 /// authored to contain until it is admitted again, and reports nothing about
 /// which version authored it. Coupling the dial to graph authority so a
 /// version gap can be detected and refused is open follow-up work.
-pub const HYDRATION_SEMANTICS_VERSION: u32 = 8;
+pub const HYDRATION_SEMANTICS_VERSION: u32 = 9;
 
 /// Semantic graph delta derived for one pre-enrichment change identity.
 ///
--- a/crates/kin-index/src/history.rs
+++ b/crates/kin-index/src/history.rs
@@ -548,6 +548,45 @@ pub fn is_external_reference_target(entity: &Entity) -> bool {
     entity.role == EntityRole::External && entity.file_origin.is_none()
 }
 
+/// The placeholder entity one placeholder relation's destination stands for,
+/// or `None` when the relation is not a placeholder.
+///
+/// Both placeholder classes are handled here so a caller cannot learn one and
+/// miss the other. That is not hypothetical: the historical ref view re-links
+/// source and inserts the result straight into a snapshot, and admission fails
+/// closed on an endpoint no entity backs, so a synthesis that knew only the
+/// cross-repo class turned a new class of edge into a hard error the moment one
+/// appeared.
+///
+/// The language comes from the caller because a target defined elsewhere has
+/// none of its own; the importing side is the only thing that observed it.
+pub fn placeholder_target_entity(relation: &Relation, language: LanguageId) -> Option<Entity> {
+    let destination = relation.dst.as_entity()?;
+    let token = relation.evidence.first()?.token.as_deref()?;
+    if is_external_import_placeholder(relation) {
+        let import_source = relation.import_source.as_deref()?;
+        let fingerprint = external_reference_fingerprint(import_source, token);
+        return Some(external_reference_entity(
+            destination,
+            token,
+            language,
+            fingerprint,
+        ));
+    }
+    if is_unresolved_receiver_placeholder(relation) {
+        let (receiver, symbol) = split_unresolved_receiver_token(token)?;
+        let name = unresolved_receiver_display_name(receiver, symbol);
+        let fingerprint = external_reference_fingerprint("", &name);
+        return Some(external_reference_entity(
+            destination,
+            &name,
+            language,
+            fingerprint,
+        ));
+    }
+    None
+}
+
 /// Build the external target a cross-repo reference resolves against.
 fn external_reference_entity(
     id: EntityId,
--- a/crates/kin-index/src/lib.rs
+++ b/crates/kin-index/src/lib.rs
@@ -61,7 +61,11 @@ pub use linker::{
     CALL_SHAPE_PARSE_COVERAGE_FULL_V1, CALL_SHAPE_PARSE_COVERAGE_INCOMPLETE_V1,
     INCREMENTAL_LINKER_CHECKPOINT_VERSION, KIN_INDEX_CRATE_VERSION,
 };
-pub use linker::{is_external_import_placeholder, EXTERNAL_IMPORT_REFERENCE_RULE};
+pub use linker::{
+    is_external_import_placeholder, is_unresolved_receiver_placeholder,
+    split_unresolved_receiver_token, unresolved_receiver_display_name,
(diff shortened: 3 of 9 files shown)


troyjr4103:
Ejected from the merge queue by a timing flake that is not in this diff. Recording it here because a merge-group failure leaves no mark on the pull request it removes, so without this the PR reads CLEAN and simply never merges.

What happened. Merge group `gh-readonly-queue/main/pr-983-93f31671...`, head `c8dc90a806fd`. Its CI run 32350571468 concluded `failure` with 14 jobs success, 4 skipped and 2 failure: `Check & Test macOS shard (1)` at the `Test` step, and `Check & Test (macos-latest)` at `Require every macOS shard to have succeeded`, which is the aggregate reporting the first one.

The single failing test is `kin-cli daemon_client::probe_process::tests::bounded_probe_runtime_timeout_starts_after_parseable_pid_readiness`, panicking at `crates/kin-cli/src/daemon_client/probe_process.rs:1458` with:

```
called `Result::unwrap()` on an `Err` value: Custom { kind: TimedOut,
error: "delayed-ready daemon probe fixture timed out after 400ms; cleanup=ok; capture=ok;
stdout=running 1 test ... test result: ok. 1 passed; 0 failed ... finished in 1.12s" }
```

The captured stdout shows the child fixture succeeding and taking 1.12s against a 400ms runtime budget, so this is a load-dependent timeout on a busy runner rather than a logic failure. The shard summary reads 747 of 3319 tests run, 746 passed, 1 failed, and nextest then cancelled the remaining 2572.

Why it is not this change:

- This branch touches no file under `crates/kin-cli`. Its nine changed files are in `kin-core`, `kin-index`, `kin-parser`, one script manifest and two adapter fixtures.
- The same test passed on the pull-request CI run for this exact commit `0d97e530`, run 32348425188, where `Check & Test macOS shard (1)` and `(2)` both concluded success.
- It passed in the local full-workspace gate on this branch at 0.894s, inside a run of 6587 tests with 6587 passing.

No code change is being made here in response. Widening a timing budget in another area's test from a pull request about JavaScript receiver extraction would mix an unrelated change into a measured diff and could mask a real regression. Re-submission is the captain's call.

PR #993, opened 2026-08-20
Gate the release train on the proof loop's own records

troyjr4103:
Run against v0.5.44's actual tag commit, this gate prints `Resolved candidate a5a316c8e6320bb25a37f2a3d2c84cb6d276fa63 from landed commit a4ffe620...` and then refuses it for having no proof records. That is this gate catching the exact incident that filed the ticket, on real GitHub data rather than a fixture.

The release train now refuses to merge a version bump or promote a tag unless the proof loop's own records exist for that exact candidate. On 2026-08-20 it did all three with nobody at the button: merged its own bump kin#986, tagged v0.5.44, and promoted to Latest, with no preflight record for the candidate and no stranger run on its bytes.

## The chain rule is the spine of this design

Two independent existence checks would prove only that a preflight happened and that a stranger happened. Requiring the stranger's `archive_sha256` to appear among the preflight legs for the gated commit proves the stranger ran on the bytes the preflight judged FOR THIS RELEASE. That is the sentence the whole ticket exists to make true, and every other rule here is in service of it.

## Why this is not just a missing assertion

The root cause is that the automation had no knowledge the proof loop exists. The only thing that had ever held an unproofed cut was a session-local holder monitor, which was correctly retired when its pause-era purpose ended. A gate that lives in a monitor can always be retired correctly again, so this one lives in the train.

Two facts turned up while building it that the ticket does not cover, and both changed the design.

The evidence was not reachable from CI at all. `.kin-coord/preflight/<stamp>/preflight.json` is an umbrella coordination path: `git ls-files` finds nothing under `.kin-coord` in this repo, and neither `bin/kin-release-preflight` nor `bin/kin-stranger` publishes its record anywhere. Both only consume from GitHub. So the records live solely on the operator's disk, and no Actions job can see them. This change therefore defines a CI-readable, sha-keyed evidence surface, an orphan `release-evidence` branch carrying `evidence/<sha>/preflight.json` and `evidence/<sha>/stranger.env`, read through the contents API. That is the same shape `scripts/check-kin-vfs-compat.mjs` already uses to read another repo's lock at a pinned ref, including its fail-closed behavior, so it inherits a reviewed precedent rather than introducing a new mechanism.

The two decision points do not share a sha. A squash mints a new commit, so the tag's sha is never the sha the proof loop judged. v0.5.44's tag commit `a4ffe620` is kin#986's squash, while the preflight for that line judged the release-next head. The promote gate bridges that through the merged pull request, and a commit no merged pull request produced resolves to nothing and is refused, which is the intended answer for a hand-pushed tag.

## What the gate checks

`scripts/check-release-proof-artifacts.mjs` takes one exact commit and refuses when the record is absent, when the answer is unreadable, when the verdict is not PASS, when the run judged uncommitted changes, when the record is about a different build, or when the stranger's `archive_sha256` matches no archive any preflight leg judged.

Citability is deliberately not the test. `bin/kin-release-preflight` hardcodes `citable: false` and `lane: "DEV-LOCAL"` on every record it emits, so requiring citability would be the mirror of a check that cannot fail: a check that cannot pass, holding every release forever and looking principled doing it. `allow_dirty` carries the same intent and a real run can clear it. A test asserts the DEV-LOCAL non-citable record is accepted, so the unsatisfiable rule cannot be reintroduced quietly.

## Where the gates sit

`release-train.yml` requires the records for the pull request head before it arms auto-merge. That is the same sha the arming step already pins the merge to with `--match-head-commit`, so nothing can land other than what was gated. A hold drafts the bump, comments once per candidate head rather than once per scheduled cycle, and writes the same `kin.release-hold.v1` marker the alarm job reads, because a rail that stands down invisibly is a failure this workflow's own comments record having had.

`release.yml` requires them again in `finalize_release`, before the `--prerelease=false --latest` flip. The job gains `pull-requests: read` for the bridge. This one protects every path into Latest rather than only the path the train owns.

## Gating performed

29 unit tests pass, registered in the `node --test` list ci.yml already runs. `actionlint` exits 0 on all three modified workflows, and rejects a deliberately broken copy as the control. `scripts/test-release-workflow-authority.py`, `check-rc-build-drift.mjs`, `check-kin-vfs-compat.mjs`, `release-hold-alarm.test.mjs`, the full release `node --test` list, `verify-zero-file-search.py`, `zero_file_search_guard.sh`, `verify-hydration-semantics.py`, `check_runtime_boundaries.sh`, `check_private_repo_coupling.sh` and `test-private-repo-coupling.py` all exit 0. No Rust changed.

## Falsification

Seven mutations, seven named test failures, source byte-identical after each restore. Removing the commit-match check fails `main holds when the records exist but describe a different build`. Removing the `allow_dirty` check fails `judgePreflight refuses a run that judged uncommitted changes`. Removing the stranger archive linkage fails `judgeStranger refuses a real run on bytes no preflight leg judged`. Treating 404 as readable fails `main holds a candidate with no preflight record`. Removing the verdict check fails `judgePreflight refuses a candidate that did not clear preflight`. Dropping the `merge_commit_sha` filter fails `resolveCandidateSha refuses a commit no merged pull request produced`. Dropping the multi-claimant refusal fails `resolveCandidateSha refuses to guess between two claimants`.

Run live against real shas rather than only fixtures. Against main's head it exits 1 with `evidence/bb8bebe3.../preflight.json does not exist on the release-evidence branch`. Against v0.5.44's actual tag commit it prints `Resolved candidate a5a316c8e6320bb25a37f2a3d2c84cb6d276fa63 from landed commit a4ffe620...` and then refuses, which is this gate catching the exact incident that filed the ticket. Against a commit with no merged pull request it refuses at the resolver.

## What this does not do, and the sequencing it needs

This is deliberately `Refs` and not `Closes`. Two pieces of the ticket's acceptance are not in this change.

The publisher does not exist yet. Nothing writes to the `release-evidence` branch, because the proof-loop scripts are umbrella-tracked and that is a kin-ecosystem change. Until it lands, this gate holds every release. That is the current intended state, since the interim hold is already live and #991 is drafted, but it is a decision to take deliberately rather than discover. Landing this first is still right: it satisfies the ticket's own acceptance that an unproofed bump cannot merge or promote, and it fails closed rather than open.

The session holder monitor should not be retired until the publisher is live, or releases will be held by a gate nothing can satisfy with no operator path to satisfy it.

## Do not land this before v0.5.45 cuts

This PR is deliberately held until v0.5.45 has shipped, and the captain submits it only after that. A tag run resolves its workflows from the tag, so if this landed first, kin#991's arming and v0.5.45's promotion would both meet a fail-closed gate with no publisher live and no records posted. The gate would block the very release whose proof loop is running correctly right now with its evidence sitting in local files. It lands with or just before the umbrella publisher, and the first gated release either back-fills its evidence or posts it natively.

Refs FIR-2525

## State at submit, 2026-08-21 00:50Z

The sequencing above is satisfied. v0.5.45 tagged at 00:24:22Z on 657a4e1d, whose tag run carries no copy of this gate. The publisher landed as kin-ecosystem#92 at 23:59Z, and the first real publish ran at 00:41Z: evidence/bb8bebe34057d9a6397db0b2ca49a7d096996800/preflight.json and stranger.env now exist on the release-evidence branch, which that publish created as an orphan branch. So the paragraph that says the publisher does not exist yet described the tree at 20:30Z and no longer describes it. The session holder monitor has been stood down; from this commit forward, holding is done by this gate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -999,6 +999,7 @@ jobs:
             scripts/check-glibc-floor.test.mjs \
             scripts/check-kin-vfs-compat.test.mjs \
             scripts/check-rc-build-drift.test.mjs \
+            scripts/check-release-proof-artifacts.test.mjs \
             scripts/check-release-version.test.mjs \
             scripts/extract-release-notes.test.mjs \
             scripts/prepare-release.test.mjs \
--- a/.github/workflows/release-train.yml
+++ b/.github/workflows/release-train.yml
@@ -58,7 +58,7 @@ jobs:
       # reads exactly like one that never came, which is the reading that keeps
       # the alarm quiet. The artifact stays, because later cycles need this run
       # to be readable from outside it.
-      marker: ${{ steps.plan.outputs.marker }}
+      marker: ${{ steps.proof.outputs.marker || steps.plan.outputs.marker }}
     steps:
       - name: Mint repository-scoped release branch token
         id: app-token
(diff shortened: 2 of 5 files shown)


troyjr4103:
For the machinery owner reviewing this: three decisions are yours, not mine, and I have deliberately not taken them.

The evidence surface is new. Nothing writes to the `release-evidence` branch yet, because both proof-loop scripts are umbrella-tracked and publishing is a kin-ecosystem change. Until that lands, this gate holds every release. That matches the interim hold already in force, but it is a decision to take rather than discover, and it is the reason this PR says Refs rather than Closes.

The layout is a contract between two repos. This side reads `evidence/<candidate-sha>/preflight.json` verbatim as `bin/kin-release-preflight` emits it, and `evidence/<candidate-sha>/stranger.env` verbatim as `bin/kin-stranger`'s `run.env`. Both are the producers' own formats with no transformation, so the publisher is a copy rather than a serializer. If you would rather key or name them differently, say so now and I will change this side, since nothing depends on it yet.

The candidate sha is the pull request head, not main and not the tag. Preflight judges the release-next head; the squash mints a different commit. The promote gate bridges that through the merged pull request rather than requiring a second record, so the publisher only ever writes one record per candidate, under the head sha it actually judged.

Worth knowing before you review the diff: `bin/kin-release-preflight` hardcodes `citable: false` and `lane: "DEV-LOCAL"` on every record it emits, so a gate requiring citability could never pass. This gate uses `allow_dirty` instead, and carries a test that the DEV-LOCAL non-citable record is accepted, so that trap cannot be reintroduced quietly.

troyjr4103:
Machinery-owner review: APPROVE (as a comment, since the shared account cannot formally approve its own PR). Land post-tag exactly as sequenced. Read in full: the shared judge, both workflow integration points, the resolver, and the falsification table.

What makes this right, verified rather than assumed: one judge imported by both decision points so they cannot disagree; fail-closed at every branch with absence and unreadability distinguished on purpose; the per-leg commit match plus the stranger-to-preflight archive chain making the evidence a linked proof rather than two existence checks; the resolver keyed on merge_commit_sha equality with multi-claimant refusal, so a hand-pushed tag resolves to nothing by construction; the workflow reading the gate from origin/main so a pull request cannot smuggle a gate change into its own judgment; and the DEV-LOCAL acceptance test that keeps the unsatisfiable-citability rule from returning quietly. The live refusal of v0.5.44's actual tag commit is the incident closing its own loop.

One property to state out loud so it is a decision and not an accident: judgeStranger requires finished_at and the byte linkage, not arm exit codes or a verdict. That matches doctrine, the stranger report must EXIST and findings get filed whatever they say, so the gate enforces evidence-exists rather than evidence-is-green. Correct as built; anyone tightening it later should read the doctrine first.

One accepted edge: the resolver reads the first page of commit-PR associations. A squash realistically associates one merged PR, and the failure direction if that assumption ever breaks is refusal, not passage, which is the right way to be wrong.

The three decisions, taken as machinery owner:
1. Fail-closed-until-publisher: ACCEPTED. The interim session hold covers the gap, floorbumps' publisher closes it, and the session holder retires only when both gates and publisher are live, per FIR-2525's clause.
2. Evidence layout: ACCEPTED as proposed. evidence/<candidate-head-sha>/preflight.json verbatim as kin-release-preflight emits it, evidence/<candidate-head-sha>/stranger.env verbatim as the archive-mode stranger's run.env. As the producers' owner I commit to that contract, and the publisher will carry a write-time belt mirroring the read-time check: it publishes only records whose legs judged that exact head sha.
3. Candidate sha is the pull request head, promote bridges through the merged PR: ACCEPTED. It matches what preflight actually judges, and one record per candidate is the right economy.

Refs FIR-2525.

PR #1007, opened 2026-08-21
Compare the relation census to the one this store last recorded

troyjr4103:
## The mechanism

`kin graph status` composes the relation-kind histogram at `crates/kin-cli/src/commands/graph.rs:547` and prints its verdict, `✓ No issues detected.`, at `graph.rs:712`. Nothing between those two lines read a previous state, because no previous state existed: the census was computed per request and discarded. So a store that lost an entire relation kind printed the numbers proving the loss, then printed the all-clear over them.

On the rc0545c stranger run, arm brown, a psf/requests store went from 1985 entity-to-entity relations at 23:59:21Z to 1807 at 00:35:16Z, and `UsesType` went from 94 edges to none, under `✓ No issues detected.` The capture is `.kin-coord/handoffs/stranger-rc0545c/brown/out/graph-status-requests.txt`. The loss itself was operator-induced, and Kin warned about the cause unprompted at commit time with `correctness-relevant override active: KIN_DAEMON_DISABLE_LSP set to non-default value "1"`. What Kin could not do was compare its own census to the one before it.

## The fix

A new `crates/kin-core/src/relation_census.rs` holds the record, the comparison rule, and the cause lookup. It follows `crates/kin-core/src/last_admission.rs` closely and on purpose. The record has a schema token in it. Publication is a staged temp file renamed into place after an fsync of the file and then of the containing directory. Reads are three-way, so `Absent` and `Unreadable` stay separate answers that may never render as "unchanged".

Writers, all through `record_relation_census` in `crates/kin-daemon/src/background_work.rs`:

- the end of a completed enrichment sweep in `crates/kin-daemon/src/daemon.rs`, taken just before the sweep publishes its snapshot so the recorded census describes the graph that publication is about to make durable;
- the end of a native commit in `crates/kin-daemon/src/api.rs`, after `graph.create_change` installs the change in the live graph;
- the end of an MCP transaction commit in `crates/kin-daemon/src/mcp_commit.rs`, beside `record_commit_provenance`, because an agent that only ever writes through MCP would otherwise leave the census frozen.

No read path writes. A census stamped on every status reading would compare each reading against the one before it. A store that lost a kind would then announce the loss once and report itself clean ever after, which is the same defect one level quieter.

Reader: `command_graph` in `api.rs` builds a `CensusContext` from the store's record and the daemon's own environment. `build_graph_status_response` compares its own census against that record. It adds one row directly under the histogram, then extends the warnings vector with one sentence per losing kind. The comparison also rides on `GraphCommandResponse` as `relation_census`, which is how `kin doctor` reads the verdict structurally instead of parsing prose, exactly as the reference-edge coverage beside it already does.

Doctor gets one new row, `Relation census`, in `check_relation_census`, with the verdict split into the pure `relation_census_health`. That mirrors `check_reference_edge_coverage` and `reference_edge_coverage_health`.

## The rule

A kind that went from more than zero to zero is a loss, reported unconditionally. A kind that fell by at least `SHARP_DROP_FRACTION` (0.25, named in the rendered line) without reaching zero is a loss. Slippage under that threshold, growth, a new kind and an identical count all appear in the comparison, and none of them withholds the all-clear. An aggregate fall is named even when no single kind crossed the threshold, because on the evidence run `References` fell 412 to 340, comfortably inside tolerance, while the total moved 1985 to 1807.

Losses are pushed as warnings rather than criticals. `graph.rs` already states why for the degraded-reconcile case directly above: criticals set the response error and would turn `kin graph status` nonzero for every caller scripting it. Killing the false all-clear is the requirement here. Changing an exit code is a separate decision.

The cause comes from `env_registry::audit_env`, so a row that names a knob names it in the words the startup warning already used.

## Why a store marker rather than an audit event

The record lives at `.kin/kindb/relation-census`, beside `last-admission`, `head-generation` and `embedding-coverage-complete`, which are the three existing per-store records of exactly this class. This was a deliberate choice over the graph's own provenance surface, and the reasoning is recorded here so it is not reopened.

`kin_model::GraphStore` comes from pinned external crates (kin-model `=0.7.10`, kin-db `=0.7.38`), carries no key/value or metadata surface, and its `GraphSnapshot` is `deny_unknown_fields` and MessagePack-positional, so a new field needs a kin-db release. The one free-form slot is `ProvenanceStore::record_audit_event`, whose `details` field already carries structured JSON on the MCP commit path. It was rejected for two reasons. `query_audit_events` filters only by actor and limit while commit attribution writes one event per changed entity, so on a busy store the census scrolls out of any bounded read and the reader reports "no previous census" while one exists. That is this same silent false negative reintroduced one level down. Separately, `kin-review` resolves attribution and its unreviewed-agent-change signal entirely from audit events and the actor they name, so census events written under a synthetic daemon actor would contaminate review output.

The marker is bookkeeping rather than an answer authority. It names no entity and carries no repository content, and a reader that lied in either direction would change only the wording of a health row. Its zero-file-search allowlist entry is function-scoped for that reason, so any newly added filesystem primitive in the module still fails the guard.

## Falsification

Clean baseline first, because a sabotage run proves nothing without one: `kin-core` 17 passed 0 failed on `relation_census`, `kin-cli` 7 passed 0 failed on the census and doctor filters.

Two passes, each removing one property, each predicted before it ran, each with the edit confirmed applied by `git diff --numstat` reading `1 1 crates/kin-core/src/relation_census.rs` first.

**Pass A removes the comparison.** `compare()` walks only `current.keys()` instead of the union of both key sets, which is the defect this ticket describes. A comparison that iterates only the current census cannot see a kind that is no longer there to iterate.

- `kin-core` 14 passed, 3 failed: `a_kind_that_vanished_is_reported_as_vanished`, `the_comparison_walks_kinds_the_current_census_no_longer_holds`, `a_comparison_against_a_recorded_census_names_the_loss_and_its_cause`.
- `kin-cli` 5 passed, 2 failed: `a_relation_kind_that_vanished_since_the_recorded_census_refuses_no_issues`, `doctor_reports_a_relation_kind_that_vanished_since_the_recorded_census`.
- Still green, which is what makes the pass discriminate: the 40 percent fall, the slip inside tolerance, growth, new kind, identical census, both persistence round-trips, both aggregate-total cases, both cause cases, `doctor_stays_green_when_no_relation_kind_lost_ground`, `doctor_separates_a_store_with_no_recorded_census_from_a_healthy_one`, `a_store_with_no_recorded_census_says_so_and_still_reaches_the_all_clear`.

**Pass B removes the persistence.** `write()` publishes to `kindb_relation_census_path().with_extension("not-published")`, so a census is written and nothing lands where `read()` looks. Chosen over stubbing the body because it leaves no dead code and no warning to muddy the result.

- `kin-core` 15 passed, 2 failed: `a_written_census_round_trips` and `a_census_survives_a_reader_that_did_not_write_it`, exactly the durability pair.
- `kin-cli` 6 passed, 1 failed: only `a_relation_kind_that_vanished_since_the_recorded_census_refuses_no_issues`. Both doctor rows pass, because they build their comparison in memory and never touch the record.

The passes fail disjoint sets except the lifecycle test, which fails under both. That is correct: it is the only case exercising record and compare together. The tree restored byte-identical after each pass (`git diff HEAD --stat` empty).

**A test of mine that could not fail, caught mid-run.** The lifecycle test first asserted that the loss arm did not print `No issues detected` and the control arm did. Neither assertion could do work: `graph_validation_fixture` always raises `all entities are Source` and `N embeddings are still pending`, so the all-clear is unreachable in both arms. The control failed on a clean tree, which exposed it. Both arms now assert on the `⚠ relation kind UsesType lost every edge it held` marker, which exists only when the warnings vector is non-empty, while the all-clear prints only when that vector is empty.

**The guard refused the module first.** `python3 scripts/verify-zero-file-search.py` failed with seven violations across the new read and write bodies before the allowlist entry existed. That refusal is the entry's own falsification: the guard does see this module and does reject it unexcused. It passes now at 178 files scanned, 9 with function-scoped exemptions.

## Current rebase and repair, 2026-08-21

This branch is rebased onto `c6487105590c7ff689e7b5c535c08d412f348a70`, current main after #1042 and #1026. The pushed head is `60a83df4a589d5e0ad18ae20e2b1460529e3bc1e`. The force update used `--force-with-lease` only after `git ls-remote` confirmed the existing remote head was exactly `33be6b3854f668982d9f2c3340f681d883bf7b43`.

The four original pull request commits were replayed in order. The clean local-only repair `2fe584bb43df5edf420d549244362a4f27203566` was reused for the two #1030 call sites rather than recreated. Main's #1040 changed doctor to fetch graph status once per run, so the census row now consumes that same `RunGraphStatus`; it does not restore the old second round trip. The shared fixture carries a healthy census and checks readable and unreadable outcomes for both graph consumers.

Source-level fallback evidence: semantic MCP was daemon-unreachable. The pull request's call-site parser reads this branch as 29 calls, 28 multiline and one single-line, all taking five arguments. Run unchanged against `origin/main`, it reads 26 calls and all take four. A search for `args=4` in the branch parser output exits 1, while the same control against main counts 26 and exits 0.

Verification at base `c6487105` and head `60a83df4`, with exit statuses read directly:

- `cargo fmt --all -- --check` exits 0.
- CI-equivalent `cargo clippy --all-targets --all-features` exits 0 under `RUSTFLAGS=-D warnings` and the repository's 45-lint allow list. Its only warning is the existing future-incompatibility note for third-party `block v0.1.6`.
- `python3 scripts/verify-zero-file-search.py` passes at 179 source files, 9 with function-scoped exemptions and 4 exempt whole-file. `bash scripts/zero_file_search_guard.sh`, `bash scripts/check_runtime_boundaries.sh`, and `bash scripts/check_private_repo_coupling.sh` all pass.
- `cargo test -p kin-cli graph --locked` exits 0. The kin-cli library target reports 121 passed and 0 failed. `a_refused_vector_checkpoint_is_named_even_when_nothing_is_pending`, `a_relation_kind_that_vanished_since_the_recorded_census_refuses_no_issues`, `a_store_with_no_recorded_census_says_so_and_still_reaches_the_all_clear`, `one_doctor_run_fetches_graph_status_once_however_many_rows_read_it`, and `an_unreadable_graph_reaches_the_row_rather_than_being_reported_once` all pass by name. The `graph_status_after_admission` integration target reports 6 passed and 0 failed.
- `git status --short` is empty. `git ls-tree -r HEAD --name-only` prints only `.cargo/config.toml` below `.cargo/`, and `git diff origin/main --stat -- Cargo.lock .cargo/` is empty.

## Not claiming

The new head's hosted checks were triggered by this push and are not claimed green here. The full `bin/kin-dev local-test kin` gate was not run; the targeted graph gate is the compile and behavior check for the E0061 failure class, and the lane preamble says targeted tests, formatting, and Clippy do not take a build slot. This verification is source-level fallback evidence, not semantic MCP proof. The pull request has not been submitted to the hosted merge queue and auto-merge has not been armed.

Closes FIR-2544

### DIFF
--- a/crates/kin-cli/src/commands/graph.rs
+++ b/crates/kin-cli/src/commands/graph.rs
@@ -1780,10 +1780,10 @@ mod tests {
         );
         assert!(
             unchanged
-                .lines
-                .iter()
-                .any(|line| line.contains("No issues detected")),
-            "the control reaches the all-clear, so its absence below means something: {}",
+                .relation_census
+                .as_ref()
+                .is_some_and(|census| !census.reports_loss()),
+            "the control withholds nothing, so the second arm's warning means something: {}",
             unchanged.lines.join("\n")
         );
 
@@ -1809,9 +1809,17 @@ mod tests {
         )
         .unwrap();
         let rendered = lost.lines.join("\n");
+        // The marker prefix is the assertion that matters. A `⚠` line exists
+        // only when the warnings vector is non-empty, and the all-clear prints
+        // only when that vector is empty, so this proves the loss reached the
+        // branch that suppresses "No issues detected". Asserting the absence of
+        // the all-clear string directly cannot prove it here: this fixture
+        // raises unrelated warnings of its own (pending embeddings, uniform
+        // roles), so that line is unreachable in both arms and the assertion
+        // would pass whether or not the census did anything.
         assert!(
-            !rendered.contains("No issues detected"),
-            "a store that lost a whole relation kind is not issue-free: {rendered}"
+            rendered.contains("⚠ relation kind UsesType lost every edge it held"),
+            "the loss is raised as a warning, which is what withholds the all-clear: {rendered}"
         );
         assert!(
             rendered.contains("UsesType went 1 to 0"),
@@ -1856,8 +1864,15 @@ mod tests {
             "the row states what it cannot do: {rendered}"
         );
         assert!(
-            rendered.contains("No issues detected"),
-            "an unrecorded census is not an issue: {rendered}"
+            !rendered.contains("⚠ relation kind"),
+            "an unrecorded census raises no warning of its own: {rendered}"
+        );
+        assert!(
+            response
+                .relation_census
+                .as_ref()
+                .is_some_and(|census| !census.reports_loss()),
+            "and reports no loss to doctor either"
         );
     }
 
--- a/crates/kin-cli/src/commands/graph.rs
+++ b/crates/kin-cli/src/commands/graph.rs
@@ -2761,6 +2761,7 @@ mod tests {
             &drained_graph,
             &Default::default(),
             &crate::commands::resources::EmbedRuntimeState::default(),
+            &Default::default(),
         )
         .unwrap();
         assert!(
@@ -2780,6 +2781,7 @@ mod tests {
                 deferred_vector_checkpoint: Some(REFUSAL.to_string()),
                 ..Default::default()
             },
+            &Default::default(),
         )
         .unwrap();
         let drained_line = drained
@@ -2821,6 +2823,7 @@ mod tests {
                 deferred_vector_checkpoint: Some(REFUSAL.to_string()),
                 ..Default::default()
             },
+            &Default::default(),
         )
         .unwrap();
         let filling_line = filling
@@ -2848,6 +2851,7 @@ mod tests {
             &filling_graph,
             &Default::default(),
             &crate::commands::resources::EmbedRuntimeState::default(),
+            &Default::default(),
         )
         .unwrap();
         let filling_quiet_line = filling_quiet
--- a/crates/kin-cli/src/commands/graph.rs
+++ b/crates/kin-cli/src/commands/graph.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2026 Firelock, LLC
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, HashMap, HashSet};
 
 use anyhow::{Context, Result};
 use kin_mcp::handlers::common::presentation_span_lines;
@@ -38,6 +38,15 @@ pub struct GraphCommandResponse {
     /// an older daemon sends none at all.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub reference_edge_coverage: Option<kin_core::reference_coverage::ReferenceEdgeCoverage>,
+    /// The relation-kind census set beside the one this store last recorded.
+    ///
+    /// Carried structurally as well as in `lines` so `kin doctor` reads the
+    /// comparison rather than parsing prose out of a terminal rendering, which
+    /// is how the reference-edge coverage beside it reaches the same surface.
+    /// Optional because a subcommand that compares nothing has none to report
+    /// and an older daemon sends none at all.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub relation_census: Option<kin_core::relation_census::RelationCensusComparison>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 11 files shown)


troyjr4103:
Ejected from the hosted merge queue at 10:46:13Z (reason failed_checks). The merge-group CI run 32472232659 failed Clippy on ubuntu and macOS shard 1 and Build on macOS shard 2 with four E0061 errors in `kin-cli` (lib test): this PR adds a fifth `census: &kin_core::relation_census::CensusContext` parameter to `build_graph_status_response` (crates/kin-cli/src/commands/graph.rs:380), and call sites that reached main ahead of it in the same queue batch (graph.rs:2636 `filling` and :2666 `filling_quiet`) still pass four arguments. Textually the merge was clean; the conflict is semantic. Repair: rebase onto main (1003, 1004, 1006 landed 10:47Z), pass the census at every call site, re-run the gates, re-submit. Lane censusdiff2 owns the repair.

troyjr4103:
Ejected a second time at 12:48:17Z (reason failed_checks): merge-group CI run 32481838443 failed with two E0061 errors at crates/kin-cli/src/commands/graph.rs:2426 and :2463, four-argument call sites of build_graph_status_response that landed with kin#1030 after this branch's rebase base. Same semantic-conflict class as the first ejection. Repair: rebase onto current main, pass the census at the new sites, re-submit at the head of an empty stretch; kin#1025 (four more call sites) is held behind it.

PR #1026, opened 2026-08-21
Decide the embedding class from the coverage counters, not from a conjunction

troyjr4103:
A `semantic_locate` response carried `semantic_coverage: {indexed: 2112, total: 2112, pending: 0}` and `completeness.classes.embeddings: "absent"` in the same envelope. An agent reading the counters proceeds and an agent reading the class backs off, both from one object. This fixes that, which is the part of FIR-2543 with a single mechanism behind it, and gives the cross-surface half the read time it needs.

## Mechanism

The literal envelope is at `.kin-coord/handoffs/stranger-rc0545c/brown/out/phase2-work.jsonl` line 361. Its `graph_body_gap_paths` reads 0, so the body gap was not the cause, and its coverage note names the real one, "role filter: 15 test-role source path(s) were withheld from ranking and from the counters beside this note".

`SemanticCoverage.complete` is a conjunction. A detached vector index clears it. So does an unfinished index. So does a graph body gap, and so does the test-role filter. `SemanticCoverage::with_graph_bodies` at `crates/kin-cli/src/commands/locate.rs:1851` sets `self.complete = false` whenever `bodies.withholds_tests()` holds, which is a true statement about the population a query ranked over and no statement at all about embeddings.

`embedding_class_states` in `crates/kin-mcp/src/envelope.rs` derived the embedding class from that one flag, mapping `complete == false` to `absent`. The absence gate at `crates/kin-mcp/src/envelope.rs:1363` read the same flag and answered `coverage_partial: the semantic index is incomplete` over a whole index. One flag produced two false readings of one store inside one response, with the store's own counters sitting beside them saying otherwise. Only two of those four causes are about embeddings, and only those two share the `kin embed` remediation every consumer was implying.

## The fix

One embedding verdict, computed where the counters are taken and read by every surface.

`EmbeddingState::observe` at `crates/kin-cli/src/commands/locate.rs:184` decides `present`, `partial`, `absent` or `unknown` from `indexed`, `total`, `pending`, whether this build carries a vector backend, and whether an index was attached to take the reading. Both "nobody could say" checks run before any counter is read, because a counter taken with no index behind it is not a small number. `SemanticCoverage` gained `embedding_state` carrying that verdict, `limited_by` naming every reason `complete` is false, and `read_at` recording when the counters were sampled. `with_graph_bodies` records `graph_role_filter` and `graph_body_gap` in `limited_by` and deliberately leaves `embedding_state` alone.

`SemanticCoverage::embedding_state()` at `crates/kin-mcp/src/envelope.rs:194` is the one accessor. It prefers the verdict the producing surface published, because that is the only reader that knew whether an index was attached, and falls back to the counters for older payloads. That fallback answers `unknown` for every reading the counters cannot decide alone. `embedding_status` reports zero indexed for every retrievable object when no index is attached, so a bare `indexed: 0` is both a fully embedded store whose index did not load and a store nobody embedded. An unrecognized wire word also reads `unknown`, so a future producer cannot certify an answer by naming a state this build has not agreed is certifiable.

`scope_limits()` at `crates/kin-mcp/src/envelope.rs:223` carries the role filter and the body gap into `completeness.limits` as disclosure rather than as a verdict about a substrate neither one measured. `embedding_class_states` and the absence gate both consume the accessor, and the gate now names the cause it actually observed. That includes a new `coverage_role_filter_withheld` reason. With embeddings whole, an absence over a narrowed population is still not an absence over the repository, so the gate keeps refusing, for the true reason instead of a false one.

One consequence is worth stating rather than leaving for a reviewer to find. With the class now
`present` on a role-filtered store, `completeness.status` reads `complete` for that response where
it used to read `partial`. The safety that reading used to provide has not gone anywhere, and it
comes from two places that are both more accurate than the old flag. The absence gate refuses with
`coverage_role_filter_withheld`, because an absence over a narrowed population is not an absence
over the repository. And the role filter always emits a `graph_role_filter:tests_withheld`
degradation, which makes `_kin.verdict` inconclusive, and `Verdict::project_onto_completeness` caps
`bound` back to `at_least` and rewrites the note. The recorded envelope shows both, carrying that
degradation and `verdict.state: inconclusive`. The unit fixtures omit the degradations array on
purpose, so they isolate the class rule from the projection that sits on top of it.

`crates/kin-mcp/src/negative.rs` publishes `embedding_state` and `read_at` beside the counters, and `coverage_clause` says "semantic coverage unknown" when the state is unknown instead of reciting "semantic coverage 0.0%", which was the structural zero in prose.

`crates/kin-cli/src/commands/search.rs` had no index-attached gate at all, so `kin search` published a measured shortfall on a graph whose counters nothing measured. It now takes the same observation locate does.

This PR carried a second commit, an honesty fix for `kin embed` from the stranger run on the released v0.5.45 bytes. That commit is gone as of the 2026-08-21 rebase below, because kin#1030 landed the same fix first. See the rebase note at the end; nothing was lost.

## Falsification

Three passes, each removing one property, each predicted before it ran. Every sabotage was confirmed present in the file before its run and the file was restored to a matching sha256 afterwards. Every exit status was read raw and every failure message in full, never through a pipe.

Pass 1 put the class back on `coverage.complete`. Result `26 passed; 4 failed`, and the headline failure regenerates the defect in the released bytes exactly:

```
case n_of_n_with_a_role_filter_withholding_paths: the class must follow the counters beside it:
  "classes":{"embeddings":"absent"} ... "semantic_coverage":{"indexed":2112,"pending":0,"total":2112}
  left: String("absent")  right: String("present")
```

I predicted five failures and got four. `a_role_filter_declared_only_by_the_body_coverage_object_is_still_read` passed, because it guards the disclosure property in `scope_limits()`, which pass 1 did not remove. The prediction was wrong about which property that test guards. `a_ranked_answer_reports_the_embedding_substrate` passed as predicted, since 40 of 100 is partial under both rules.

Pass 2 kept the accessor and made the legacy fallback read `indexed == 0` as `absent`. It predicted exactly one failure and produced exactly one, `29 passed; 1 failed`:

```
a count nobody could read is unknown, never zero: "classes":{"embeddings":"absent"}
  left: String("absent")  right: String("unknown")
```

Every other new test passed under pass 2, which is what shows each guards a distinct property rather than all restating one.

Pass 3 belonged to the `kin embed` commit that the 2026-08-21 rebase dropped, so it no longer describes anything this PR carries. It is left recorded here rather than deleted, because the falsification it reports did happen; it simply now evidences code that landed under kin#1030 instead. Passes 1 and 2 above are this PR's own.

The main table test drives cases with real counters, one per row and each named on its own: `n_of_n_with_a_role_filter_withholding_paths`, `n_of_n_with_nothing_pending`, `zero_of_n_against_an_attached_index`, `some_indexed_with_work_still_pending`, and `counters_taken_with_no_index_attached_to_read_them`. There is no empty-input arm, because a store with no eligible entity makes the agreement trivially true and could not fail if the rule were removed.

## Gates

The full local gate `bin/kin-dev local-test kin -- --no-fail-fast` returned 0 with
`Summary [ 491.811s] 6632 tests run: 6632 passed (7 slow, 3 leaky), 12 skipped`, over the tree its
own line names, `.kin-coord/worktrees/embedcount-kin @ bd195030 (chore/lane-embedcount-kin)`. It
was re-run after the rebase onto 674af63a for fmt and the four guard scripts; the test total above
is from bd195030, which carries the same six files.

An earlier gate run on the same tree went red on one test,
`kin-cli::live_graph_durability_disclosure mcp_status_and_locate_disclose_live_only_entities_and_then_stop_once_a_commit_records_them`,
and that is worth reporting rather than burying. It fails when the daemon's `try_lock` on
`state.embedding_work` at `crates/kin-daemon/src/api.rs:2057` is refused for the fixture's whole
flat 30-second budget, which happens when an embedding batch holds that lock at
`crates/kin-daemon/src/daemon.rs:232`. Across eight runs the result tracked box load rather than
the diff: `origin/main` passed at load 17, 91 and 38, this branch failed at load 98 and 96 and
passed at load 40, and two partial-revert variants failed at load 112 and 22. The code rules
this change out on mechanism twice over. Its only kin-daemon edit sits inside `#[cfg(test)] mod tests`, which
starts at line 12414, so it is not in the daemon binary the fixture spawns. And the one hunk adding
a graph call, `vector_index_attached` in `search.rs`, is off that test's path entirely: the fixture
calls `kin_graph_status` and `semantic_locate` and nothing else, and the daemon serves
`semantic_locate` from `local_semantic_coverage` at `api.rs:7528`. A bisect had pointed at that
hunk and the call graph disproved it. Filed on FIR-2563 with the run table.

`cargo fmt --all -- --check` exits 0. Clippy exactly as `.github/workflows/ci.yml` runs it, `RUSTFLAGS=-D warnings` with the `.github/clippy-allowed-lints.txt` allow list over `--all-targets --all-features`, exits 0 with zero error lines. `python3 scripts/verify-zero-file-search.py` exits 0 over 177 source files, `bash scripts/zero_file_search_guard.sh` exits 0, `bash scripts/check_runtime_boundaries.sh` exits 0 and `bash scripts/check_private_repo_coupling.sh` exits 0. `git ls-tree -r HEAD --name-only | grep -F '.cargo/'` prints only `.cargo/config.toml`, and `git diff origin/main --stat -- Cargo.lock .cargo/` is empty.

## Left undone

This does not close the ticket. Ask 1, that `kin graph status` and `kin status` derive from one authoritative store counter rather than computing their own, is a larger change than one lane should make during a release freeze. Those two surfaces still sample their own graphs. They do now have `read_at` on the coverage payload. That answers the half of the disagreement that is about time rather than truth. The four-state verdict is there for them to adopt whenever someone unifies them.

Nothing here was reproduced against a live daemon. Every result is from unit tests over payload fixtures plus the envelope the stranger run recorded.

One seam is worth naming. kin-mcp does not depend on kin-cli, so its mirror spells the two scope labels as literals (`SCOPE_LIMIT_ROLE_FILTER`, `SCOPE_LIMIT_BODY_GAP`) with doc comments naming the kin-cli constants they mirror. Nothing in the build enforces that those stay spelled the same, and a rename on either side would silently stop the role filter being disclosed.

## Rebase note, 2026-08-21 12:5xZ

This PR was ejected from the merge queue at 12:13:48Z with reason `merge_conflict` against kin#1030, which was queued behind it and landed first at 12:13:12Z as squash `3c21f868`. Rebased onto `origin/main` by the salvage lane, which holds both sides of the conflict.

What changed, and why nothing was lost. The branch's second commit, `f3674888` "Separate what a kin embed run embedded from what the store holds", touched `crates/kin-cli/src/commands/embed.rs` and nothing else, and its whole content was superseded by what kin#1030 landed: both fix the same stranger finding on the same line. Resolving the conflict to the landed side made that commit empty and the rebase dropped it. `crates/kin-cli/src/commands/embed.rs` on this branch is now byte-identical to `origin/main`, and the branch carries one commit against five files, all of it the FIR-2543 embedding-class work this PR is titled for.

One behaviour did not survive, deliberately. `full_coverage_line`'s zero arm said the daemon's background worker had already covered the store. The landed version does not, because a store already whole when a pass starts may equally have been filled by `kin init` or by an earlier explicit `kin embed`, and nothing at the completion line distinguishes them. The landed line says the work predates the pass, which is what the process can establish. The two tests that asserted the attribution went with it; every other property they guarded is covered on main by `a_pass_that_inherited_a_covered_store_says_the_work_predated_it` and `a_pass_that_embedded_reports_its_own_work_and_the_span_it_moved`.

One difference is named rather than smuggled in. `full_coverage_line` reported the store as entity and artifact totals separately; the landed line reports the coverage span over retrieval keys (`3/53` to `53/53 indexed`). Both are useful and neither is wrong. Adding the entity and artifact split to the landed sentence would need its own test and falsification, so it is a follow-up rather than something a rebase quietly introduces.

Gates re-run on the rebased branch through `bin/kin-lane run embedcount kin`, every exit status read raw and every total taken from its run's own `test result:` summary line rather than an rc file:

- `cargo test -p kin-mcp --lib`: exit 0, `593 passed; 0 failed`
- `cargo test -p kin-cli --lib commands::`: exit 0, `1461 passed; 0 failed`
- `cargo fmt --all -- --check`: exit 0
- clippy exactly as ci.yml runs it, `--all-targets --all-features` under `RUSTFLAGS=-D warnings`: exit 0, no unused or never-used warnings
- `python3 scripts/verify-zero-file-search.py`: exit 0, 178 source files
- `bash scripts/zero_file_search_guard.sh`, `check_runtime_boundaries.sh`, `check_private_repo_coupling.sh`: all exit 0

`git ls-tree -r HEAD --name-only | grep -F '.cargo/'` prints only `.cargo/config.toml` and `git diff origin/main --stat -- Cargo.lock .cargo/` is empty. Pushed with `--force-with-lease` pinned to the pre-rebase head `f3674888`, and the PR head was confirmed equal to the branch head afterwards rather than inferred from a green check.

Refs FIR-2543

### DIFF
--- a/crates/kin-cli/src/commands/locate.rs
+++ b/crates/kin-cli/src/commands/locate.rs
@@ -126,6 +126,95 @@ pub struct RetrievalDegradation {
     pub remediation: String,
 }
 
+/// What the embedding substrate itself was observed to be for one query.
+///
+/// This is the one embedding verdict in the system. It is decided from the
+/// counters beside it and from whether an index was attached to take them, and
+/// from nothing else, so no surface has to re-derive a verdict from
+/// [`SemanticCoverage::complete`]. That flag is a CONJUNCTION over the substrate
+/// AND the population a query ranked over, and deriving an embedding verdict
+/// from it is how a `semantic_locate` envelope came to carry `indexed 2112,
+/// pending 0` beside `classes.embeddings: absent` in one response (FIR-2543):
+/// fifteen test-role paths had been withheld from ranking, which clears
+/// `complete` and says nothing at all about embeddings.
+///
+/// `Unknown` is a state, not a shade of `Absent`. `embedding_status` answers
+/// `indexed = 0` for every retrievable object when no vector index is attached,
+/// which is exactly the reading a never-embedded repository produces, so a
+/// verdict taken from the counters alone cannot tell a fully embedded store
+/// whose index was not loaded from one that was never embedded. The two have
+/// different remediations and only one of them is a real gap.
+#[derive(Serialize, Deserialize, Clone, Copy, Debug, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum EmbeddingState {
+    /// Every eligible entity is indexed and nothing is queued, measured against
+    /// an index that was actually attached.
+    Present,
+    /// Some eligible entities are indexed and some are not.
+    Partial,
+    /// An attached index holds no embedding for any eligible entity.
+    Absent,
+    /// Nobody could say: no vector backend in this build, or no index attached
+    /// to take a reading from.
+    Unknown,
+}
+
+impl EmbeddingState {
+    /// The wire word, so a caller matching on the payload and a caller matching
+    /// on the type read the same vocabulary.
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::Present => "present",
+            Self::Partial => "partial",
+            Self::Absent => "absent",
+            Self::Unknown => "unknown",
+        }
+    }
+
+    /// Decide the state from an observation, at the one place that holds every
+    /// input it needs.
+    ///
+    /// Order matters and is deliberate. Both "nobody could say" checks run
+    /// before any counter is read, because a counter taken with no index behind
+    /// it is not a small number, it is not a number. That includes the
+    /// zero-eligible case: a graph with nothing retrievable and no index
+    /// attached is the shape that let a store report complete semantic coverage
+    /// over an empty graph (FIR-2215), and it is unknowable here rather than
+    /// whole.
+    pub fn observe(
+        supported: bool,
+        index_attached: bool,
+        indexed: usize,
+        total: usize,
+        pending: usize,
+    ) -> Self {
+        if !supported || !index_attached {
+            return Self::Unknown;
+        }
+        if total == 0 || (indexed >= total && pending == 0) {
+            return Self::Present;
+        }
+        if indexed == 0 {
+            return Self::Absent;
+        }
+        Self::Partial
+    }
+}
+
+/// Machine-stable reason [`SemanticCoverage::complete`] is false because this
+/// build ships no vector backend.
+pub const COVERAGE_LIMIT_VECTOR_SUPPORT_DISABLED: &str = "vector_support_disabled";
+/// Machine-stable reason `complete` is false because no index was attached.
+pub const COVERAGE_LIMIT_VECTOR_INDEX_ABSENT: &str = "vector_index_absent";
+/// Machine-stable reason `complete` is false because the index is unfinished.
+pub const COVERAGE_LIMIT_EMBEDDINGS_INCOMPLETE: &str = "embeddings_incomplete";
+/// Machine-stable reason `complete` is false because the role filter narrowed
+/// the population the counters were taken over.
+pub const COVERAGE_LIMIT_GRAPH_ROLE_FILTER: &str = "graph_role_filter";
+/// Machine-stable reason `complete` is false because graph-owned source bodies
+/// are missing for some paths.
+pub const COVERAGE_LIMIT_GRAPH_BODY_GAP: &str = "graph_body_gap";
+
 /// Honest, in-band report of how complete the embedding (semantic) signal was
 /// for a locate query. This is the trust-contract "per-signal degradation"
 /// property: a partial semantic index is surfaced, not hidden behind an opaque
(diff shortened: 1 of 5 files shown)


troyjr4103:
Ejected from the hosted merge queue at 11:42:10Z (reason failed_checks). The failing check was "Feature permutation tests (macos-latest)" in merge-group CI run 32476311918, one test: `daemon_client::probe_process::tests::bounded_probe_runtime_timeout_starts_after_parseable_pid_readiness` panicked at probe_process.rs:1458 with "delayed-ready daemon probe fixture timed out after 400ms" (1634 passed, 1 failed). That is FIR-2491, the probe-process readiness budget losing to a loaded runner, which lane flakytests is fixing; nothing in this PR touches it. Re-submitted unchanged.

troyjr4103:
Ejected from the hosted merge queue a second time at 12:13:48Z, reason merge_conflict: the speculative merge on top of kin#1030 (salvage, which edits the same `kin embed` completion function this PR adds `full_coverage_line` to) conflicts textually. Repair: rebase onto main once #1030 lands, keeping both changes (the daemon's own coverage readings from #1030 and the two tests here), then re-submit. Lane salvage owns the rebase since it holds both sides.

PR #1037, opened 2026-08-21
Write the Kin PATH line where a non-interactive zsh reads it

troyjr4103:
`kin` is now added to PATH in the file a non-interactive zsh actually reads, so a
script, a Makefile, a launchd or systemd job, an editor shelling out, or an agent
running `zsh -c` can find it on a fresh install.

## What was wrong

`kin setup` and `scripts/install.sh` both wrote the PATH line to `.zshrc` and
`.bashrc` and to nothing else. zsh reads `.zshrc` only when the shell is
interactive and reads `.zshenv` on every launch, so the one file that would have
made `kin` findable everywhere was the one file Kin never wrote.

Measured with this host's own zsh, one fixture binary named `kin` on a sandbox
`PATH`, one variable changed:

```
export in .zshrc   $ zsh -c 'kin --version'   zsh:1: command not found: kin   rc=127
export in .zshenv  $ zsh -c 'kin --version'   kin 0.0.0-fixture               rc=0
```

The same pair through a nested `sh`, which is how most tooling reaches a binary:
`zsh -c 'sh -c "kin --version"'` gave `sh: kin: command not found` at rc=127 with
the export in `.zshrc`, and rc=0 with it in `.zshenv`. An interactive zsh reads
`.zshenv` too, so nothing is lost by moving it: `zsh -i -c 'kin --version'`
returns rc=0 on the fixed layout.

This is the fourth symptom of FIR-2236. The other three were fixed already: the
zsh and bash wrappers are inlined at `crates/kin-cli/src/commands/setup.rs:237`
and `:476`, fish wraps the same eight binaries, and a test replays only the public
functions into a fresh shell. The shim-side half of the ticket, which makes the
exclusion hold regardless of how a process was launched, is kin-vfs#123.

## The fix

`shell_path_rc` at `crates/kin-cli/src/commands/setup.rs:1410` says where a
shell's PATH line belongs, which is `.zshenv` for zsh and the same file as the
hook for everything else. The hook does not move with it. Sourcing the projection
hook from `.zshenv` would inject the shim into every non-interactive shell, which
is the opposite of what the hook's own exclusion wrappers exist for, so
`shell_rc` keeps the hook in the interactive file and this decides the PATH line
separately.

`RcBlocks` says which of the two blocks a file carries, and `rc_write_plan`
turns a shell name into the files to write and what each one takes. One entry for
a shell that reads one file, two for zsh. `plan_rc_update` gates each half on it,
so neither file can take the other's block.

The install ledger records the PATH line against `shell_path_rc`, so uninstall
excises the real block rather than a block in a file the line no longer reaches.
A fix that writes to a new home and leaves its readers pointed at the old one is
worse than no fix, so every reader moved with it. `legacy_shell_path_targets` gains `.zshenv`, so
a full uninstall does not leave an export behind pointing at a directory the same
run removed. And `check_shell_path` in `crates/kin-cli/src/commands/health.rs`
reads both files, because reading only the hook's file would report a correctly
installed host as missing its PATH, which is a check that can never pass.

`scripts/install.sh` moves the same line for the same reason, and keeps the
existing guard that declines to write anything on a host with no zsh evidence.

## Falsification

The installer half, rehearsed against a sandbox `HOME` with the block extracted
from `scripts/install.sh` itself. On the fix: a fresh home with `SHELL=/bin/zsh`
got the export in `.zshenv` and nothing in `.zshrc`; a home carrying `.zshrc` got
`.zshenv`; a Linux home carrying `.bashrc` and `.zshrc` got `.bashrc` and
`.zshenv`; a second run added nothing, leaving exactly one export line. The
control is a macOS home with no rc file at all and `SHELL=/bin/bash`, where
nothing was written.

Reverting `install.sh` to `.zshrc` and rerunning the same rehearsal put the
export in `.zshrc`, and `zsh -c 'command -v kin'` against that home returned
rc=1. Restoring the file to a matching sha256 and rerunning put it in `.zshenv`,
and the same shell ran the fixture and returned rc=0.

Four passes on the Rust half, each removing one property, with the sabotage
asserted applied by sha256 before any result was believed and the file restored
to a matching sha256 afterwards. The driver refuses outright when its target
matches anything other than exactly once, so a sabotage cannot land somewhere
else and produce a result about code nobody edited.

Undoing the split, so `shell_path_rc` returns the hook's file for zsh, failed
`zsh_writes_its_path_line_where_a_non_interactive_shell_reads_it` on
`left: 1, right: 2` against `[("/Users/troyfortinjr/.zshrc", HookAndPath)]`, and
failed `install_shell_hook_adds_path_and_hook_blocks_idempotently` with `setup
wrote no .../home/.zshenv, so a non-interactive zsh has no PATH line to read: No
such file or directory (os error 2)`. Result: 156 passed, 2 failed.

Making every file carry the PATH block failed
`each_half_of_a_split_plan_writes_only_its_own_block` with `the interactive file
took the PATH line as well`, and failed the idempotence test with `the PATH line
stayed in the interactive-only file, where a script or an agent cannot see it`.
Result: 156 passed, 2 failed.

Reverting doctor to read only the hook's file failed
`shell_path_reads_the_file_a_non_interactive_zsh_actually_reads` with `doctor
read only the interactive rc, so an install that put the PATH line where a script
can see it reads as broken; got Misconfigured: zsh: .../home/.zshrc does not add
.../kin-home/bin to PATH`. Result: 2 passed, 1 failed.

Dropping `.zshenv` from the uninstall sweep failed
`uninstall_sweeps_the_file_zsh_reads_on_every_launch`, whose message printed the
remaining target list with `.zshrc` present and `.zshenv` gone. Result: 0 passed,
1 failed.

An earlier run of the last two passes is not being reported as evidence. Its test
filters were missing the `tests::` segment, so both selected zero tests and both
reported `test result: ok. 0 passed`, which is a check that cannot fail. The
filters were corrected and every pass above was rerun from scratch.

## Gates

`cargo fmt -- --check` rc=0.

Clippy exactly as `.github/workflows/ci.yml` runs it, under
`RUSTFLAGS=-Dwarnings` with the 45-lint debt allow-list read out of
`.github/clippy-allowed-lints.txt`: `cargo clippy --all-targets --all-features --
$allows` rc=0 in 2m 15s.

The four guard scripts that job runs: `python3 scripts/verify-zero-file-search.py`
rc=0 over 177 source files, `bash scripts/zero_file_search_guard.sh` rc=0,
`bash scripts/check_runtime_boundaries.sh` rc=0, and
`bash scripts/check_private_repo_coupling.sh` rc=0.

`cargo test -p kin-cli --lib commands::setup::` returned 158 passed and 0 failed.
`cargo test -p kin-cli --lib commands::health::` returned 101 passed and 0 failed.
Every run went through `bin/kin-lane run shimshell kin`, whose printed line names
worktree
`/Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/shimshell-kin` and
target
`/Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/cargo-target/shimshell-kin`.

The full local gate was not run under tonight's slot saturation, where a compute
quiet was held for a host at its swap ceiling. Hosted CI is the gate of record
for this PR, and the per-check lines go into
`.kin-coord/reports/shimshell-20260821.md` once every check has a conclusion.

`git ls-tree -r HEAD --name-only | grep -F '.cargo/'` prints only
`.cargo/config.toml`, and `git diff origin/main --stat -- Cargo.lock .cargo/` is
empty: this change touches no dependency and needs no version bump, since `kin`
is not a registry-published crate.

## The other half

kin-vfs#123 carries the shim-side half of this ticket, which makes the exclusion
of Kin's own binaries hold however a process was launched rather than only where
these shell functions reach. That PR says `Refs`; this one closes the ticket.

## Rebased past the 12:48Z queue ejection, 2026-08-21

This pull request entered the hosted queue at 12:23:28Z and was removed from it
at 12:48:18Z, both read from its own timeline, with the ejection reason reported
as `merge_conflict`. Three pull requests landed in that same minute: #1033 at
12:48:00Z, #1035 at 12:48:17Z and #1036 at 12:48:18Z. #1035 is the one this
branch collided with. When this lane picked the pull request up, its
`mergeStateStatus` read DIRTY.

The branch is now rebased onto `origin/main` at `199edead`, which is #1036, and
both commits survived: `git rev-list --count origin/main..HEAD` returned 2 and
the rebase output carries no "patch contents already upstream" line. The rebase
ran with rerere disabled on both invocations.

`git merge-tree --write-tree --name-only origin/main HEAD` named the conflict set
before the rebase started, and it was one file,
`scripts/zero-file-search-allowlist.json`. Both Rust files auto-merged:
`crates/kin-cli/src/commands/health.rs`, which #1035 also changed, and
`crates/kin-cli/src/commands/setup.rs`, which #1029 also changed.

### How the one conflict was resolved

The collision is the `health.rs` entry in the allowlist, whose `reason` is a
single line that both sides rewrote. #1035 appended an eighth justification class
for its home-directory probe and added two pinned expressions,
`let entries = match std::fs::read_dir(home) {` and
`match std::fs::symlink_metadata(&path) {`. This branch inserted a clause into
class (2) saying zsh has two rc files, and moved the
`.and_then(|rc| std::fs::read_to_string(rc).ok())` pin from a bare string to an
object with `count: 2`, because doctor now reads both of them. The resolution
keeps every one of those.

It was resolved by computation rather than by eye, since the merged line is 6443
characters long. The three merge stages were read out of the index and decoded:
#1035's edit is a pure append after the base's 5015 characters, and this branch's
is a pure insertion of 250 characters at offset 586, so the two edits are
disjoint and the merge is unambiguous. The resolved file was then parsed and
asserted deep-equal to that computed merge, which also confirmed the document
still carries 37 allowlist entries and that everything outside this one entry is
byte-equal to `main`. The count object was rewritten to the indentation every
other count object in the file uses.

### The resolution is falsified in both directions

The zero-file-search guard passing is what proves both sides survived, because it
pins expressions by exact count against the merged `health.rs`. Two sabotage
passes, each asserted applied before its result was believed and each restored to
a matching sha256 of `3feb91f8b39ddf78add732988556fad3ae4809430e1f32fac95eb9c038215a18`.

Dropping #1035's two new pins made the guard fail rc=1, naming
`let entries = match std::fs::read_dir(home) {` at `health.rs` line 1235 and
`match std::fs::symlink_metadata(&path) {` at line 1267.

Reverting this branch's pin to a bare string made the guard fail rc=1, naming
`.and_then(|rc| std::fs::read_to_string(rc).ok())` at lines 1604 and 1616, with
the allowlist validation reporting the count mismatch as "occurs 2 times in
scanned code (want 1)".

So the merged file carries #1035's probe at 1235 and 1267 and this branch's two
reads at 1604 and 1616, measured by the guard rather than read off a diff.

### Gates on the rebased tree

`cargo fmt --all -- --check` rc=0.

Clippy exactly as `.github/workflows/ci.yml` runs it, `cargo clippy --all-targets
--all-features` under `RUSTFLAGS=-D warnings` with the 45-lint debt list read out
of `.github/clippy-allowed-lints.txt`, rc=0 in 3m 06s. It printed one warning,
the future-incompatibility notice for the third-party crate `block v0.1.6`, which
predates this change.

All six guard scripts `ci.yml` names returned rc=0:
`python3 scripts/verify-zero-file-search.py` over 178 source files,
`bash scripts/zero_file_search_guard.sh`,
`python3 scripts/verify-hydration-semantics.py` at
HYDRATION_SEMANTICS_VERSION=9, `bash scripts/check_runtime_boundaries.sh`,
`bash scripts/check_private_repo_coupling.sh`, and
`python3 scripts/test-private-repo-coupling.py` with its four tests.

Targeted tests, totals read raw rather than through a pipe:
`cargo test -p kin-cli --lib commands::setup::` returned 160 passed and 0 failed,
`cargo test -p kin-cli --lib commands::health::` returned 106 passed and 0
failed, and `cargo test -p kin-core --lib layout::` returned 19 passed and 0
failed. That last one is #1035's own suite, run here to show the rebase cost it
nothing. Every run went through `bin/kin-lane run shimshell kin`, whose printed
line names worktree
`/Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/shimshell-kin`,
and the tools echoed the tree under test as `84f11f1f chore/lane-shimshell-kin`.

The full local gate is still not run, under the same slot saturation the section
above records, and hosted CI stays the gate of record for this pull request.

`git ls-tree -r HEAD --name-only` prints only `.cargo/config.toml` under
`.cargo/`, `Cargo.lock` carries zero `source = "path` entries against a positive
control of 8 `sparse+` entries, and the diff against the rebase base lists only
the four files this branch owns. `main` moved two commits under this branch while
the gates ran, `dc1b05e4` and `fd277687`, and they touch none of those four; a
merge of this head into the current `main` returns no conflict, so the branch
reads BEHIND rather than DIRTY.

Closes FIR-2236


troyjr4103:
Ejected from the hosted merge queue at 12:48:18Z (reason merge_conflict) after kin#1033, #1035 and #1036 merged in the same minute; #1035 (layout: discover refuses the managed root, doctor probes the shim) is the likely overlap. Repair: rebase onto main, resolve keeping both behaviours, re-submit.

PR #1091, opened 2026-08-22
Grade which budget an elision names, not merely that it names one

troyjr4103:
kin#1084 landed a reason on every elision so a caller knows which budget cut its
answer and which lever recovers it. The graders it landed beside asked only
whether a reason was present, so the distinction they exist to protect was never
actually checked.

The mutant that found it is the one the wave asks for: make the branch fire
always. Stamping every elision `token_budget` passed the whole acceptance suite.
Checks 0 to 4 drive `trace_data_flow`, `semantic_locate` and `impact_analysis`,
none of which has a pack token budget at all, so that reason cannot be true of
them; a caller who believes it raises a lever the tool does not have, gets the
same answer back, and has been told a cut was recoverable when it was not. The
unit tests caught the mutant. The suite that runs against real binaries did not,
and that is the one a stranger's CI reports on.

`grade_reason` now grades the cause, and both graders call it. A cut on a tool
the pack budget cannot reach must name `response_budget` and nothing else.

The same one-sidedness sat in the CLI. `kin context` prints "over budget: every
section keeps a row" when the never-empty floor pushes a pack past its own
budget, and nothing asserted the note was absent on a pack that fits, so making
that branch fire always left all 26 context tests green. The whole-pack test now
asserts the note is absent and asserts the fixture actually fits, because an
absence assertion over a fixture that was never inside its budget proves nothing.

## Falsified

Exit statuses read raw. Neutering `grade_reason` turns `--self-test` red on three
cases, exit 1, and the intact suite exits 0. Making the CLI note fire always
fails `a_whole_pack_carries_no_budget_note_at_all` with "a pack inside its budget
must not report itself over one"; before this change the same mutant left 26
passed, 0 failed.

Two mutants from the same sweep were already dead and stay recorded here rather
than fixed: stamping every elision `token_budget` fails three kin-mcp unit tests,
and marking every row whether or not the budget took anything from it fails
`a_row_that_never_had_a_body_is_neither_marked_nor_counted`.

## Gates

`cargo test -p kin-cli --lib commands::context` 26 passed, `cargo fmt --all --
--check` clean, clippy with the 45-lint allow list under `RUSTFLAGS=-D warnings`
exit 0, and the guard steps enumerated from a real CI job's own step list rather
than grepped out of ci.yml, because a failed guard step skips every step after it
in that job: zero file-search and its answer-module arm, hydration replay,
runtime boundaries, private repo coupling and its guard test, installer checksum
fixtures and the release gate assertions. All exit 0.

Relates to FIR-2482

### DIFF
--- a/crates/kin-cli/src/commands/context.rs
+++ b/crates/kin-cli/src/commands/context.rs
@@ -970,5 +970,16 @@ mod tests {
             !rendered.contains("Raise --budget"),
             "no lever is offered for a cut that did not happen: {rendered}"
         );
+        // The other side of the over-budget note. Without this the note could
+        // fire on every pack and every test here would still pass, which is a
+        // fixture sitting entirely on one side of the branch.
+        assert!(
+            !rendered.contains("over budget"),
+            "a pack inside its budget must not report itself over one: {rendered}"
+        );
+        assert!(
+            whole.pack.as_ref().unwrap().actual_tokens <= 32_000,
+            "the fixture must actually fit, or the assertion above proves nothing"
+        );
     }
 }
--- a/scripts/acceptance/response_budget_elisions.py
+++ b/scripts/acceptance/response_budget_elisions.py
@@ -92,6 +92,27 @@ IMPACT_LISTS = [
 # The reason code a response still over its ceiling publishes.
 OVER_BUDGET_REASON = "response_over_budget"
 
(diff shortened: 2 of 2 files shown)


troyjr4103:
Reopened as a PR from chore/lane-fir2482-reason-kin, the branch kin-lane's own naming lets this lane land. Same commit, same content.

PR #1110, opened 2026-08-23
Keep every caller and retire every stale site across a comment-only commit

troyjr4103:
A `kin commit` of a module docstring lost a real caller and kept a call site
that had moved, and both query answers reported themselves clean. This keeps the
caller, retires the site, and makes the graph say when it is short.

Reproduced on a 19-entity Python fixture in about 40 seconds, the same shape the
rc0550 brown stranger hit on `psf/requests`. Before the commit,
`find_references(Session.send, [calls])` returns `Session.request` and
`SessionRedirectMixin.resolve_redirects` composed through the override, with
`total_upstream: 2`. After 26 lines of docstring and nothing else, it returns one
row, `find_references(HTTPAdapter.send)` reports the call site at both its old
line and its new one stamped `reference_sites_complete: true`, and the relation
census is the only surface that knows: Calls 15 to 13, Overrides 3 to 2, entity
count held at 19. Both responses carry an empty degraded list.

Three mechanisms, measured.

The live reconcile path dropped every same-file edge the incremental linker
resolved, on the ground that the pipeline's own per-file resolution already
carried it. It does not carry `Overrides`: `kin_index::linker` is that kind's
only producer, and its base-class walk resolves a same-file base first. The edge
therefore reached the retire loop as a parser-derived edge with both endpoints in
the file that this pass had not produced, which is the `parser_authoritative`
condition exactly, so a docstring deleted it and `find_references` lost the
caller it composes through that override. `CrossFilePass` now carries a
`same_file` bucket and the reconciler folds it in beside the cross-file one.

A re-derived parser edge also took the identity of whichever relation sorted
first under its key, and a language-server edge derives its own id from the same
triple in its own namespace. When that one sorted first, the parse was written
onto it and the real parser edge went unmatched, surviving with its pre-edit
span, which is how one call site reported two lines. Matching now keeps a parser
identity for a parser edge and adds rather than overwrites when the bucket holds
none, and a surplus parser copy of a key this pass re-derived from this file is
retired whether or not its destination is in the file.

Everything the pass preserves rather than re-derives, language-server enrichment
above all, now has its evidence placed through the declaration that contains it.
Placement is proven: the innermost declaration whose pre-edit lines contain the
span, and only when that declaration merely moved. A span that cannot be placed
is cleared, because an absent line and a wrong line are different answers and
only one of them is honest.

And the census pass publishes what it already knew. `record()` has always refused
a losing census and named the loss in a log nobody tails; it now leaves a durable
hold beside the baseline, retired the moment a census advances, and the MCP
envelope reads it, so `find_references`, `get_context_pack` and every other tool
answer with `relation_census_loss` instead of presenting as clean.

On the fixture after the fix: `total_upstream` is 2 on both sides, every reported
reference line carries the call, and relations hold at 86 with every kind
unchanged and no census warning.

## Tests

Nine unit tests on the two new decision functions, four integration tests driving
the real parser, linker and reconciler beside the FIR-2598 one, two on the hold's
publish-and-retire rule, one on the envelope flag and its absence, and magic
acceptance checks 18, 19 and 20 on a new `mixinadapter` fixture, wired into the
suite `acceptance.yml` already runs and `gate.py` already reads.

Each acceptance check carries its own control. Check 18 refuses to grade a
fixture that never composed a caller through an override, check 19 reports
unreadable rather than passing when no reference line reaches it, and check 20
asserts the healthy store carries no census-loss signal before it introduces one.

Pre-fix, with `crates/` reverted to the base commit and rebuilt:

```
CHECK 18 FAIL a comment-only commit lost caller(s) SessionRedirectMixin.resolve_redirects
CHECK 19 FAIL 2 of 4 reported reference line(s) no longer carry the call:
  Session.request -> Session.send line 27 reads 'import time'
CHECK 20 FAIL find_references answered over a graph short of its own census with
  degraded_signals [], claiming nothing was wrong
0 pass, 3 fail, 0 unreadable
```

Post-fix, same suite, same fixture: 3 pass, 0 fail, 0 unreadable.

Every new test was falsified by a mutation that breaks the behavior it guards,
each with its own recompile proven in the build log and the tree restored clean
afterwards. Restoring the same-file `continue` kills the override test. Returning
the lowest id from `parser_identity_to_keep` kills three unit tests and the
enrichment-identity test. Switching the re-anchor loop off kills the span test
and check 19 while checks 18 and 20 still pass, which is what pins check 19 to
the re-anchor rather than to the override fix. Narrowing
`duplicate_parser_relation` back to `both_in_file` kills the surplus-copy test.
Making `publish_hold` a no-op kills both census tests, and removing the envelope
assignment kills the envelope test.

The zero-file-search guard caught the new record on its first run, which is the
guard working: the `relation_census.rs` pin carried a bare `read` and a second
body by that name turned it red. It is fixed with a counted pin plus
`publish_hold` and a justification rather than a widened exemption, and falsified
by dropping the `publish_hold` pin, which returns the guard to red.

`bin/kin-parity` against this branch's own release binaries: FINAL
PASS-WITH-ALLOWANCES, naming only the three standing allowances (firstrun/9
FIR-2580, firstrun/3 FIR-2501, brownfield/4 FIR-2464).

Full report: `.kin-coord/reports/relay-fir2644-20260823.md`

Closes FIR-2644
Refs FIR-2537


troyjr4103:
Correction on this PR's provenance, recorded here because the squash message on
`main` cannot be changed and must not be rewritten.

The squash commit `efcb535e` carries **another pull request's body**. Its diff is
correct and is this branch unchanged, ten files, and its subject line is this
PR's title. Its body is kinlab#290's FIR-2012 text about a hosted publish smoke,
which has nothing to do with this change.

The cause was a shared scratch file, not a queue fault. This PR was opened with
`gh pr create --body-file <session scratchpad>/pr-body.md`. That scratchpad
directory is shared by every agent lane in the session, `pr-body.md` is an
unnamespaced basename that at least three lanes used, and another lane wrote its
own body over that path between when this one was written and when `gh` read it.
The file at that path has since been overwritten a second time by a third lane.
So `gh` faithfully sent the bytes it was given, and the hosted queue faithfully
minted the squash from them.

Two consequences worth naming. The `Closes FIR-2644` line never reached the
merge, so the ticket was closed by hand with the evidence instead. And anyone
reading `main`'s history at `efcb535e` will find a description of a different
repository's change; the PR body above is the accurate one, and the full lane
report is at `.kin-coord/reports/relay-fir2644-20260823.md`.

The trap is going into `docs/traps.md` in the umbrella: agents sharing one
scratchpad must namespace temp files per lane, a `--body-file` path under the
lane's own worktree is the safe idiom, and the tell is a landed body that
describes another repository's change.

PR #1116, opened 2026-08-24
Count a page thirteen children share once, not thirteen times

troyjr4103:
Kin's daemon budget summed every descendant's resident set, and a resident set counts each shared page once per process that maps it. A daemon and its thirteen enrichment children map the same binary, the same libraries and the same copy-on-write heap, so the total tracked how many descendants there were rather than any memory anyone held. The v0.5.51 stranger read the result: `graph status` claiming 25.3 GiB inside a container hard-capped at 12 GiB whose true peak was 9.99 GiB and whose `memory.events` said the cap was never reached. Not merely wrong, impossible. Background embedding refused on all three candidate stores, so every vector answer that release could give came from an empty index on a box that had room.

Counting the children is not the defect and stays. A language server holding 1.93 GiB in a process of its own is Kin's memory and is charged to the same container. What changes is the per-process figure.

## The reading

Linux takes PSS from `/proc/<pid>/smaps_rollup`, where a page shared thirteen ways is counted once across the thirteen, and falls back to the private set out of `smaps` and then to the resident-minus-shared pair out of `statm`, both of which drop a shared page rather than multiplying it. macOS takes `ri_phys_footprint` through `proc_pid_rusage`, this platform's own footprint number, which leaves out the clean file-backed pages every copy of a binary shares; the choice and what it does not prove are written into the code. Windows already had such a figure and `sysinfo` carries it under a name that describes the API rather than the number: `virtual_memory()` is `PROCESS_MEMORY_COUNTERS_EX.PrivateUsage`, private commit. A daemon that cannot measure itself publishes nothing, exactly as an unreadable process table already did.

The fold is then held to what the kernel says the cgroup is charged, so the figure printed beside a container's cap can never exceed it. The clamp is expected never to bite now that the readings are proportional; when it does it says so on the surface a reader is already looking at, rather than presenting the kernel's number as its own.

## What it is graded by

Against shipped 0.5.51 bytes, inside `docker run --memory=2g`, the two new acceptance checks both FAIL, quoting the defect: a daemon and twenty-three children published as holding 1,450,377,216 bytes while the kernel charged that container 283,222,016, and a published rung of `critical` against a derived budget of 1 GiB while the tree held 62 MiB of private memory. That is the ticket's finding reproduced end to end on the surface it was found on.

A runner has no memory cap, so the workflow runs those two checks a second time inside a container, which is where that arm can fire at all. Every Rust test was falsified in an isolated copy with its own target directory, one mutant at a time, asserting the crate recompiled each round: reading `Rss` where the parser reads `Pss` fails four tests, disabling the clamp fails two, dropping the disclosure clause fails one, and returning `Process::memory()` from the row builder fails the live test. That last mutant is why the live test measures a row from one refresh: the first version compared a finished walk against readings taken around it, and building the process table allocates megabytes into the process being measured, so the mutant went undetected.

Both acceptance graders are falsified in `--self-test`, including a case for a build whose clamp read the cap instead of the kernel charge, which is inside the cap and still impossible.

Two things this does not claim. No vector was computed: the suite runs with `KIN_DAEMON_AUTO_EMBED=0` and loads no model, so what is graded is the admission decision and the rung the embed gate reads. And on macOS there is no second kernel figure to grade the reader against without root, so its coverage is unit tests and an explicit note rather than an end-to-end arm.

Closes FIR-2653


troyjr4103:
Ejected from the merge queue by a registry outage, not by this diff.

The `merge_group` SAST run for `gh-readonly-queue/main/pr-1116-0e7e2d18` failed in `cargo-deny`, and auto-merge came off this PR with nothing marked on it, which is what an ejection looks like from here: OPEN, CLEAN, and not merging.

The cause is in the run's own log. `cargo-deny` could not download from the private registry:

```
error: failed to download from `https://kinlab.ai/registry/cargo/dl/kin-vfs-core/0.4.14`
Caused by:
  failed to get successful HTTP response from `.../kin-vfs-core/0.4.14` (34.107.250.203), got 502
```

https://github.com/firelock-ai/kin/actions/runs/32686017253

The registry is still down as I write this. Three endpoints, all 502: the `kin-vfs-core` download, the `kin-db` download, and `registry/cargo/index/config.json`. So it is the registry rather than one crate.

Worth noting for whoever reads the run: the job's own error line says "cargo-deny failed but reported no vulnerability or unsound advisory, so the failure is in bans, licenses, or sources and belongs to this change." It does not belong to this change. A download that 502s reaches the same classifier a real sources finding does.

Re-arming once the registry serves again.

PR #1141, opened 2026-08-26
Classify colocated framework tests, and promote in-file tests on both pipeline paths

troyjr4103:
Two gaps put the wrong role on real entities. Colocated framework tests read as production source,
and a test declared inside a product file changed role depending on which pipeline path last touched
it.

**Gap 1: the test-filename rule listed eight literal extensions.** `Foo.test.tsx`, `FooTest.java`,
`foo_spec.rb`, `FooTests.cs`, `FooSpec.kt` and every other colocated framework test classified as
production source. A React repository with colocated tests read 100% Source, and the two linker gates
that filter on `role != Test` then removed nothing, because nothing was Test. Those gates exist to
stop a production call fanning out to a test double, so the failure is silent and it is on the
answer, not on a report.

The rule is now delimited rather than a bare suffix, and the delimiter is what makes widening safe.
`latest.java` ends with the letters `test.java`, so a lowercased `ends_with` would call it a test. A
marker counts only when an underscore, a dot, or a capital separates it from the rest of the stem.
The capital case reads the **original** spelling, because the lowercased form cannot tell `FooTest`
from `latest`.

**Gap 2: the incremental path skipped the in-file test promotion.** The full-parse path promotes an
entity the parser recognised as a test, by name, inside a product file. The incremental path assigned
the path role and stopped. So a `#[test] fn` in `src/lib.rs` was Test after a full parse and Source
after an incremental reconcile: the same code, two roles, decided by which path ran last. That is a
defect on its own, independent of whether any repository was ever observed reading all Source.

## Falsification

Five mutation and check pairings, each applied to the committed tree, each restored, with the tree
verified clean and no mutation marker left behind. The driver restores on an `EXIT`, `INT` and `TERM`
trap rather than at the end, because a killed run leaves a mutant that `git status` cannot tell from
a half-finished edit.

| Mutation | Check that went red |
|---|---|
| the marker is matched undelimited and lowercased, which is the naive widening | `classify_colocated_framework_test_files` |
| the rule goes back to the eight literals | `classify_colocated_framework_test_files` and `role_reaches_the_graph_for_every_path_convention` |
| the incremental path assigns the path role and nothing else | `an_in_file_test_keeps_its_role_on_both_pipeline_paths` |
| the promotion is made unconditional, so any file holding a test is all test | `an_in_file_test_keeps_its_role_on_both_pipeline_paths` |

The first is the one worth reading. It is not a broken version of the fix, it is the version a
reasonable person writes when asked to widen a suffix list, and the controls are what catch it:
`latest.java`, `Latest.java`, `latest/release.rs`, `contest.py`, `manifest.rs`, and pytest's own
`src/_pytest/python.py`, each of which ends with a marker's letters and is not a test.

The fourth is the control on the promotion itself. Without it, "promote in-file tests" is satisfied
by calling every entity in the file a test, and the product function beside the `#[test] fn` would go
with it.

**Every role assertion is read back after `apply_to_graph`, not off the in-memory `IndexedFile`,**
so a role that is correct at parse time and lost on the way to the store is caught too. The
incremental arm drives `index_file_with_hint` against a real file and a real blob store, because that
is the path the gap lived on, and a batch-only fix passes the first assertion and leaves the second.

## What I ran

`cargo test -p kin-index` (19 suites green, 361 lib tests plus the new integration file),
`cargo fmt --all -- --check`, and `bin/kin-precheck kin` against this tree: **16 gates ran, 16
passed, 0 failed**. It reports this tree 1 commit behind origin/main, which is kin#1134 landing
while this branch was under test; the queue builds its own merge, and the tool's own threshold is 20,
so this is a warning it prints rather than a state it refuses. Named here because the gate list this
run read is the one on this tree.

`bin/kin-parity` on `61e143ed2`, release profile, clean tree: **FINAL PASS-WITH-ALLOWANCES** in
769.2 s. magic 22 pass / 0 fail / 0 unreadable. brownfield 10 pass / 0 fail / 1 unreadable. firstrun
9 pass / 2 fail. All three non-passes are pre-existing and named in the allowance list, and **none is
from this branch**: brownfield 4 (FIR-2464, the truncated reference walk), firstrun 3 (FIR-2501, the
projection row under Kin's own shell hook) and firstrun 9 (FIR-2580, `kin daemon stop --all`).

Hosted CI is the gate of record and the Windows legs run only there.

## Tickets

Part of FIR-1940

The ticket carries two independent findings. Item 1 is the missing role field, and it is now whole:
the parity half landed in kin#1129 and this is the classifier half that makes the field mean
something, since a `role` that is always `Source` is a field with no information in it.

**Item 2 is untouched and this PR does not go near it.** The six MCP client registrations still
disagree on repo selection: Codex CLI is written with a `--repo` pin to the absolute path of whatever
repository `kin setup` ran in, while Claude Code, Cursor and Gemini CLI resolve from the agent's
working directory, and `kin setup doctor` reports all six equally `ok`. That is why this is `Part of`
rather than `Closes`.

Split from the parity half on purpose. That was a serializer change proved by a serializer test; this
is an ingest change proved by an ingest test that writes to a graph and reads back. Sharing a PR
would let a red run mean either.

## Claims not being made

That any repository was measured reading 100% Source. The ticket's own comment records that as
unverified at whole-graph scale; what is verified is that both gaps produce it on the right
repository shape, and both are now closed with checks that redden when restored. That the marker
list is complete: it covers the conventions the adapters in this registry actually serve, and a
language whose convention is neither delimiter nor capital would need its own arm. That role
classification is now correct everywhere; these are the two gaps the validation pass found in code,
not a proof that no third exists.

### DIFF
--- a/crates/kin-index/src/pipeline.rs
+++ b/crates/kin-index/src/pipeline.rs
@@ -300,13 +300,26 @@ impl IndexPipeline {
             parse_state,
         } = output;
 
-        // Convert extracted entities to model entities
+        // Convert extracted entities to model entities.
+        //
+        // The in-file test promotion below is the same one the full-parse path
+        // applies, and it is here because its absence made role path-dependent:
+        // a `#[test] fn` in `src/lib.rs` was Test after a full parse and Source
+        // after an incremental reconcile, so the same code changed role
+        // depending on which path last touched the file (FIR-1940). A promotion
+        // on one path only is the shape kin#1123 caught in a different consumer.
         let role = classify_file_role(&file_id.0);
+        let test_names: std::collections::HashSet<&str> =
+            tests.iter().map(|test| test.name.as_str()).collect();
         let mut entities: Vec<Entity> = extracted_entities
             .into_iter()
             .map(|e| {
                 let mut ent = e.into_entity_with_source(language, file_id, Some(&source));
-                ent.role = role;
+                ent.role = if role == EntityRole::Source && test_names.contains(ent.name.as_str()) {
+                    EntityRole::Test
+                } else {
+                    role
+                };
                 ent
             })
             .collect();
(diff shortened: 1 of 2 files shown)


troyjr4103:
Recording a queue ejection, because nothing else does.

This PR was armed and taken into the merge queue, and its entry is now gone with `mergedAt` still
null and `state: OPEN`. Read across three consecutive polls to rule out the race where the queue
clears `mergeQueueEntry` one poll before `mergedAt` appears, so this is an ejection rather than a
success in progress. The PR's own checks read 34 pass, 5 skipping, 0 failed throughout, which is why
the ejection is invisible from the PR surface.

The cause, from the `merge_group` runs rather than the PR checks. Run `32930778998` on
`gh-readonly-queue/main/pr-1141-ca01b7ed` failed in `Check & Test macOS shard (1)`:

```
the_admission_ladder_accounts_for_the_whole_of_init
git ["add", "-A"] failed: error: unable to create temporary file: Invalid argument
error: src/mod_4.rs: failed to insert into database
error: unable to index file 'src/mod_4.rs'
fatal: updating files failed
```

That is the runner's git failing to write a temporary file, not an assertion about product
behaviour. Three things say it is not this change. The failure is an OS-level refusal rather than a
failed expectation. This PR touches `crates/kin-index` only, while the failing test is in
`kin-core` and drives a real git fixture. And the same code passed the same suite on this PR's own
head.

Re-arming rather than changing anything. If it ejects again on the same test, that is a different
claim and this note should not be read as having established a flake: one failure with an
infrastructure signature is grounds for a retry, not a diagnosis.

PR #1142, opened 2026-08-26
Let a hosted daemon read repository authority from the backend it has

troyjr4103:
Four routes were unservable on any daemon backed by storage: `/repos/{id}/refs`, `/files`, `/history` and `/archive/tar`. Both branches of `repository_authority_snapshot` refused, for different reasons, and neither refusal was about the data.

The advertised repository was routed by identity into the local branch, which opens authority through the startup workspace binding. `open_with_backend` never sets one, so the daemon answered `local daemon is missing its startup workspace binding` for the repository it exists to serve.

Every other repository took the branch that read the envelope off `InMemoryGraph::to_snapshot()`, which sets `repository_authority` to `None` unconditionally, because authority belongs to the immutable publication manager and never to the mutable graph. Asserting on a field the value's own type erases is a check that cannot pass, so that branch refused every sibling whatever storage held.

## The fix

Select by capability, which is what `repository_transfer_authority` in the same file already does: a backend answers for every repository it serves, and the local binding answers only when there is no backend. Authority comes from the manager either way, since that is the only thing that owns one. The two selections now agree instead of contradicting each other eight hundred lines apart.

Addressing and absence stay decided before authority is opened, so an unserved id and a served id storage holds nothing for both keep their 404s. Opening the manager first would answer both by minting an empty authority, which is a 200 carrying no refs where a refusal belongs.

The manager's own open already refuses a snapshot with no envelope and one whose envelope belongs to another repository, so neither check is repeated here.

## Falsification

Four tests, on the hosted fixture rather than a local one, each proven to go red under a named mutation on the committed tree. Baseline is all green.

| mutation | advertised | sibling | envelope-less | transfer loop |
| -- | -- | -- | -- | -- |
| `MUTANT-IDENTITY-FIRST`, put the identity branch back in front | FAILED | ok | FAILED | FAILED |
| `MUTANT-DERIVED-GRAPH`, read the envelope off `to_snapshot()` again | FAILED | FAILED | FAILED | FAILED |
| `MUTANT-FALLBACK-TO-DERIVED`, fall back to the derived snapshot when the manager refuses | ok | ok | FAILED | ok |

The third row is why the inverse test exists. A fix that quietly falls back instead of refusing passes both positive tests and is caught only by the one that requires the refusal to stand. Without it the suite could not tell a real fix from a weakened assertion.

The script restores on an EXIT, INT and TERM trap, asserts each mutant marker is in the tree before its build starts, and aborts the arm rather than running it unmutated. The tree was confirmed clean and marker free afterwards.

## What I ran

`bin/kin-precheck kin` through `bin/kin-lane run` on this worktree: 16 gates enumerated from ci.yml's own check job, 16 passed, 0 failed, on `245ca1789`. It caught a formatting slip in the second commit that would otherwise have cost a hosted round, which is the whole point of it. It also named the three check-job steps it deliberately does not run, and why, which is the part that makes the tally mean something. `cargo test -p kin-daemon --lib` on the rebased tree: 1042 passed, 0 failed. Rebased onto `origin/main` `5ee045f2a`.

## A cost this adds, named rather than hidden

The manager is opened per request, and on a hosted backend that is not free. kin's own comment at `state.rs:2148` says opening authority there re-downloads the snapshot `open_with_backend` already loaded and may replay the whole history to validate it. `/refs` is a hot route, the prod `kin` object is 69 MB, and `kin-db`'s GCS backend has no snapshot byte cache.

It is not fixed here on purpose. `ProjectionAuthorityCache` solves this for the local path, but it holds an `ActiveApiRepositoryAuthority` and labels it with a `LocalPublicationIdentity`, so it cannot take a hosted manager as is. An authority cache that serves one publication past its replacement is worse than a slow route, and getting the invalidation right needs its own falsification rather than a rider on a correctness fix. FIR-2709 carries it, including a cheaper option: the load path already decodes a snapshot with the envelope and throws it away, so capturing it there would cost a clone instead of a download.

## The other half, and why it needs no new code

This does not make the hosted smoke green on its own. The objects in the graph bucket carry no envelope, because they were exported from the derived graph rather than published through a repository transaction, so `RepositoryAuthorityManager::open` refuses them with `snapshot has no v13 authority envelope`.

The route that fixes that already exists and is the product's own. `kin push --url <daemon-base-url> --ref <ref>` publishes exact repository-v6 history to a native Kin remote, driving `POST /repos/{id}/transfer/receive`. That route resolves authority through `repository_transfer_authority`, which has always preferred `state.storage_backend`, so it was hosted-capable before this change. The receive commits a repository transaction, and `prepare_successor` stamps the envelope. A store with no object mints genesis authority on open, so the first push has something to advance.

The second commit proves it rather than asserting it: `a_hosted_store_admits_a_native_transfer_and_then_serves_the_head_it_admitted` admits a native transfer into an empty hosted-shaped store and reads the head back through `repository_authority_snapshot`. Both halves had to work and only one of them did, so a test covering either alone would have been green while the pair stayed useless.

It also pins the precondition the bucket violates. A store holding an envelope-free snapshot refuses at `RepositoryAuthorityManager::open`, which happens before the transfer is applied, so an exported object cannot be repaired in place and has to be moved aside first. Order: land this, deploy it, move the objects aside, then push. Nothing in that needs a kin-db change.

Closes FIR-2708
Part of FIR-2707

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -9682,32 +9682,59 @@ async fn repository_authority_snapshot(
     state: &DaemonState,
     repo_id: &str,
 ) -> Result<kin_db::GraphSnapshot, (StatusCode, String)> {
+    // Capability decides which authority answers, never identity. A daemon with
+    // a storage backend reaches every repository it serves through that
+    // backend, the one it advertises included; a daemon without one has only
+    // the local binding it opened at startup. This is the selection
+    // [`repository_transfer_authority`] already makes, and the two now agree.
+    //
+    // Deciding by identity instead is what made these routes unservable on a
+    // hosted daemon. `DaemonState::open_with_backend` leaves
+    // `cached_workspace_id` and `local_repository_backend` unset, so routing
+    // the advertised repository into the local branch reached a binding that
+    // cannot exist there and answered "local daemon is missing its startup
+    // workspace binding" for the repository the daemon was built to serve.
+    //
+    // The other branch could not answer either. It read the envelope off
+    // `InMemoryGraph::to_snapshot()`, which sets `repository_authority` to
+    // `None` unconditionally because repository authority belongs to the
+    // immutable publication manager and never to the mutable graph. Asserting
+    // on a field the value's own type erases is a check that cannot pass, so
+    // that branch refused every sibling whatever storage held. Authority is
+    // read from the manager here for the same reason the local branch always
+    // did: it is the only thing that owns one.
+    if let Some(backend) = &state.storage_backend {
+        // Addressability and absence stay decided where they already were, and
+        // before authority is opened. An unserved id is a 404 naming the served
+        // identity, and a served id storage holds nothing for is a 404 too.
+        // Opening the manager first would answer both by minting a fresh empty
+        // authority, turning two distinct refusals into one 200 carrying no
+        // refs.
+        repo_scoped_graph(state, repo_id).await?;
+        let repository_id = RepositoryId::new(repo_id.to_string()).map_err(|error| {
+            (
+                StatusCode::UNPROCESSABLE_ENTITY,
+                format!("invalid repository identity: {error}"),
+            )
+        })?;
+        // The manager's own open refuses a snapshot with no envelope and one
+        // whose envelope belongs to another repository, so neither check is
+        // repeated here; repeating them would assert over bytes that already
+        // passed the same validator.
+        let authority = RepositoryAuthorityManager::open(repository_id, Arc::clone(backend))
+            .map_err(repository_authority_error)?;
+        return Ok(authority.read_authority().snapshot().clone());
+    }
+
     if repo_id == state.cached_repo_id {
         let authority = ActiveApiRepositoryAuthority::open(state)?;
         return Ok(authority.manager.read_authority().snapshot().clone());
     }
 
-    let graph = repo_scoped_graph(state, repo_id).await?;
-    let snapshot = graph.to_snapshot();
-    let metadata = snapshot.repository_authority.as_ref().ok_or_else(|| {
-        (
-            StatusCode::FAILED_DEPENDENCY,
-            format!("repository {repo_id} has no repository-v6 authority envelope"),
-        )
-    })?;
-    if metadata.repository_id.as_str() != repo_id {
-        return Err((
-            StatusCode::FAILED_DEPENDENCY,
-            format!(
-                "requested repository {repo_id} but durable authority belongs to {}",
-                metadata.repository_id
-            ),
-        ));
-    }
-    metadata
-        .validate_against_snapshot(&snapshot)
-        .map_err(repository_authority_error)?;
-    Ok(snapshot)
+    Err((
+        StatusCode::NOT_FOUND,
+        format!("repository {repo_id} is unavailable in local daemon mode"),
+    ))
 }
 
 fn repository_metadata(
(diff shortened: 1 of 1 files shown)


troyjr4103:
Recording a silent merge-queue ejection, since the queue does not mark the PR it ejects.

Added to the queue 05:03:51Z, removed 05:17:10Z by `github-merge-queue[bot]` with an empty reason on the timeline event. Position was 1, state `AWAITING_CHECKS` on every poll until the entry went null.

No check failed. The entry's own `merge_group` runs on `gh-readonly-queue/main/pr-1142-ca01b7ed063f4da9394fcdf3c923778a341fe620` were `SAST` success, `PR Text Hygiene` success, and `CI` plus `Acceptance` still `in_progress` at the moment of removal, last updated 05:14:27Z and 05:04:13Z. They are now orphaned and will never conclude, because the ref they ran against is gone.

Nothing on this side changed. The branch head has been `245ca17891c7` since 04:26:02Z and was not pushed to. The PR-side rollup was 32 SUCCESS, 5 SKIPPED and one in progress.

For contrast, the other entry in the queue tonight, #1141, was ejected at 04:55:06Z for a legible reason: its `merge_group` `CI` had concluded `failure`. This one had no such conclusion, which makes it a different case rather than the same one.

Queue configuration at the time: `grouping_strategy ALLGREEN`, `max_entries_to_build 5`, `check_response_timeout_minutes 90`. The entry was roughly thirteen minutes old, so the 90 minute timeout was not reached.

Re-enqueuing. If it ejects again without a concluded check, the pattern is worth a ticket rather than another retry.

troyjr4103:
Correction to the comment above. It was not a silent ejection, and "No check failed" was wrong.

The failure was there the whole time, one level below where I looked. Inside `merge_group` CI run 32932604105, the job **`Feature permutation tests (macos-latest)`** completed with `failure` at the step **`Test featureless kin-cli`**, and `Feature permutation tests (ubuntu-latest)` was cancelled behind it. The run itself still reads `in_progress`, because its other jobs were still going when the entry was pulled.

That is the mechanism, and it is worth stating plainly for whoever hits this next: **the queue ejects on a required job's check-run the moment that job fails, before the workflow run concludes.** So the run-level status stays `in_progress` and the timeline event carries no reason. Reading runs gives you "nothing failed"; reading jobs gives you the failure. When a queue entry vanishes, read jobs, never runs.

The command behind that step is `cargo test -p kin-cli --no-default-features --lib --tests --locked`. This branch's local gates did not cover it: `cargo check -p kin-daemon --all-targets`, `cargo test -p kin-daemon --lib` and `bin/kin-precheck kin`'s clippy all run with default or all features, so a featureless break in a dependent package was outside every one of them. Reproducing it locally now.

What stands from the comment above: the timeline event, the times, the untouched branch head, and the queue configuration. What does not stand is the claim that nothing failed, and the framing of this as the catalogue's unmarked-ejection case. It was marked; I read the wrong level.

troyjr4103:
Cause, and it is not this change.

The failing test is `commands::setup::tests::a_shell_that_reads_one_file_still_gets_one_plan` in kin-cli's featureless permutation. It counted two fish rc targets because the `macos-latest` runner it landed on carries a real `/Users/runner/.config/fish/config.fish` beside the test's temp path. The test does not isolate `HOME`, while its sibling at `setup.rs:17966` does, with an `EnvVarGuard` on a tempdir. So it passes or fails on which runner it gets, and nothing in `api.rs` reaches it. A fix is landing separately as lane fishrc.

Three readings of the same code, which is what settles it:

| where | runner | `Feature permutation tests (macos-latest)` |
| -- | -- | -- |
| this PR's own head `245ca1789`, 04:29:05Z | one without the file | success |
| `merge_group` run 32932604105, 05:04Z | one with it | failure |
| the same command locally, `cargo test -p kin-cli --no-default-features --lib --tests --locked` | this laptop | the named test passes |

Same commit, opposite outcomes, so the variable is the runner.

Worth writing down, because I got this wrong on the way here. I reached for a negative control and picked the wrong one: main's push run at the same base `ca01b7ed0` passed the same job, and I read that as isolating my change. It does not. That comparison varies the code **and** the runner at once, and attributes the whole difference to the code. The control that actually holds the code fixed was this PR's own earlier run, sitting one API call away the whole time. On hosted CI, a same-code re-run is a control; a different-code same-base run is not.

PR #1146, opened 2026-08-26
Take kin-db 0.7.62 for write-time relation endpoint admission

troyjr4103:
kin-db now refuses, at the write, a relation naming an artifact, test, contract, work item, verification run or external reference the store does not carry, instead of admitting it and failing some later unrelated transaction over the strand. That is the asymmetry FIR-2613 is about: the transaction gate validated endpoints and a direct upsert did not, so a stranded edge got in first and the refusal fired later, on whichever commit happened to drop the endpoint. A deleted file could take an unrelated commit down that way.

Entity endpoints stay permissive on purpose. kin's own linker mints a canonical external-import placeholder whose destination entity is intentionally absent, and this repository's `graph validate` reports a graph carrying one as passing every integrity check. Judging that shape needs the predicate that recognizes it, which lives here in kin-index where kin-db cannot read it, so kin-db does not get to make that call yet. FIR-2715 carries the move that makes it possible.

The same release stops `remove_entities_for_file` keeping a cross-file edge with a dangling destination, makes the strand count readable, and fixes a `has_incoming_of_kinds` fallthrough that let an edge from nowhere keep an entity alive.

## Fifteen fixtures now admit what they always meant to have

The gate found fifteen fixtures relating an artifact or a test the store never admitted. That builds a repository that could not exist: every persist gate already refuses those edges, so these graphs were ones no snapshot would have accepted, independent of this pin. Admitting what a fixture always meant to have makes it more correct, not less; these tests now assert about a repository that could really hold what they claim.

Artifacts are admitted through a transaction carrying a `TreeDelta::Added`, the idiom `repository_rename_planner` already used for its own tree blobs. Tests are admitted through `create_test_case` with no scopes, so the identity exists and the fixture's own edge stays the only relation it asserts about.

| File | Tests | Refused on |
|---|---|---|
| `python_impact_walk_agreement` | 4 | source artifact |
| `python_value_reference_dead_code` | 4 | source artifact |
| `python_bare_builtin_call_resolution` | 3 | source artifact |
| `reference_call_site_lines` | 1 | source artifact |
| `repository_rename_planner` | 1 | source artifact |
| `commands::graph` unit tests | 2 | source test |

`repository_rename_planner` needed one more step. Its extraction-incomplete certificate named two freshly minted artifact ids, describing a dependency between two files the repository does not contain, so `add_file` now returns the identity it admitted and the test names those.

The three python harnesses each carry their own admission helper rather than sharing one. `tests/common/mod.rs` is bounded-subprocess support with a module doc that says so, and pulling kin-db graph types into a module every CLI test binary compiles would be a worse trade than three local copies of ten lines.

## The lock diff, and a stowaway

Two lines, the version and the checksum, keeping the `sparse+` registry source.

`cargo update -p kin-db --precise` also tried to downgrade `winapi-util`'s Windows-only dependency from `windows-sys 0.61.2` to `0.48.0`, which has nothing to do with this pin and is invisible on a host that cannot build that target. It is restored to main's resolution both times it appeared, on the 0.7.61 attempt and again here, so it is reproducible rather than a one-off. Both `windows-sys` versions already exist as packages in the lock, so the restore leaves it self-consistent, and `cargo metadata --locked` re-resolved kin-db 0.7.62 from the sparse registry afterwards without rewriting the lock.

## Gates

`cargo nextest run --workspace --locked --no-fail-fast`, run under the daemon lock because the featureless kin-cli suite spawns kin-daemons in temp directories: **7,349 tests run, 7,348 passed, 1 failed**. `cargo test --doc --locked` exit 0. `cargo fmt --all -- --check` clean.

`--no-fail-fast` is not decoration here. The first run of this gate stopped at its first failure, reporting 1 failed out of 316 with 7,018 never run, which is a symptom rather than a count and would have hidden the other thirty-six.

The one remaining failure is `kin-core::init_phase_ladder_contiguous the_admission_ladder_accounts_for_the_whole_of_init`, and it is not this change. It asserts that no wall-clock stretch of a real init runs with no admission span open, so its subject is scheduler behaviour, and it has now been read three times on this host at 7.1%, 8.4% and 3.1% against a 2.0% allowance while total init time stayed within 4% of itself. The gap tracks host contention and nothing else; this change adds one map lookup per relation write. It passes on CI, where the most recent successful run on `main` is green on every Check and Test job. Filed as FIR-2718 with all three readings and the fix direction.

Closes FIR-2613.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3285,9 +3285,9 @@ dependencies = [
 
 [[package]]
 name = "kin-db"
-version = "0.7.60"
+version = "0.7.62"
 source = "sparse+https://kinlab.ai/registry/cargo/"
-checksum = "d968316efe650186062f10d97bc18de1a59169ad9e9d0aed739df9c12873834e"
+checksum = "5adaa50328a871716579ac3827bc04d4bf7cf3ce670d4c1d6f9c5e8c1dd11a77"
 dependencies = [
  "bincode",
  "bstr",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -155,7 +155,7 @@ kin-lsp = { version = "=0.7.0", registry = "kin" }
 # backend on every target (incl. Linux, where it cannot compile). Metal is re-enabled
 # additively on macOS via a `[target.'cfg(target_os = "macos")'.dependencies]` block in the
 # binary/runtime crates (kin-cli, kin-daemon). Feature unification makes that additive.
-kin-db = { version = "=0.7.60", registry = "kin", default-features = false }
+kin-db = { version = "=0.7.62", registry = "kin", default-features = false }
 kin-vfs-core = { version = "=0.4.14", registry = "kin" }
 
 [profile.release]
--- a/crates/kin-cli/src/commands/graph.rs
+++ b/crates/kin-cli/src/commands/graph.rs
@@ -1651,9 +1651,52 @@ mod tests {
     use super::*;
     use kin_model::{
         ArtifactId, Entity, EntityMetadata, FilePathId, FingerprintAlgorithm, GraphNodeId, Hash256,
-        LanguageId, Relation, RelationId, RelationOrigin, RepoPath, ResolvedArtifact, ResolvedTree,
-        SemanticFingerprint, SourceSpan, TreeEntry, Visibility,
+        LanguageId, LocatedEntry, Relation, RelationId, RelationOrigin, RepoPath, ResolvedArtifact,
+        ResolvedTree, SemanticFingerprint, SourceSpan, TestCase, TestId, TestKind, TestRunner,
+        TransactionDelta, TreeDelta, TreeEntry, VerificationStore, Visibility,
     };
+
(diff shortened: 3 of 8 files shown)


troyjr4103:
Ejected from the merge queue, recorded here because nothing else marks it.

Timeline: `added_to_merge_queue` 07:30:58Z, then `removed_from_merge_queue` by `github-merge-queue[bot]` at 07:48:12Z, which is the same second PR #1145 merged and became main's tip `48fea9ba`. This entry's speculative merge had been built on that base, and it went out when the queue re-formed. Its own merge-ref checks were still running at the time: Secret Scan, SAST and PR Text Hygiene had passed on `gh-readonly-queue/main/pr-1146-48fea9ba`, with CI and Acceptance in flight.

Nothing on this PR says any of that. It reads `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`, all 41 checks green, head unchanged at `8cb34909`. A green and mergeable PR that simply never merges is what an ejection looks like from the PR surface, which is why this note exists rather than a silent re-land.

No code change. Re-arming the queue on the same head.

PR #1157, opened 2026-08-26
Carry a payload into the transfer harness, and pin which bound refuses it

troyjr4103:
**Rebased onto kin#1156 and flipped.** This branch was written before the Git-authority
bootstrap landed and pinned the 409 both arms then got. It was built to go red the day that
bound moved, keyed on the STATUS rather than on a sentence, and that is exactly how it failed:
`left: 422, right: 409` on the over-cap arm with the message naming the byte budget. What
follows describes the flipped shape; the falsification table below covers both.

Every transfer test in `api.rs` proves the LOOP. The one the hosted republish runbook cites,
`a_hosted_store_admits_a_native_transfer_and_then_serves_the_head_it_admitted`, builds its
pack from `seed_replica_change`, whose changes carry empty `tree_deltas` and empty
`entity_deltas`. Every hosted repository is content, so the loop passing was never evidence
the republish can work.

This is the harness that carries one. It exists before the route that would let a payload
through, so its job today is to pin the refusals precisely enough that the day something
moves, the right assertion goes red and the wrong one does not.

Tests only, one file, no product code, no kin-db, no version bump, no lock.

## The fixture pair, and why it is a pair

One builder parameterised by blob count, so the two arms differ in exactly one variable. Each
stages a Git worktree, writes N distinct megabyte blobs, commits once, and runs
`kin_core::init_from_git_adopting` against a hosted identity, which is how the republish
runbook stages its five stores.

24 MiB against 4 MiB, mirroring the shapes FIR-2746 measured as not fitting and fitting.

The arithmetic is asserted before any push depends on it, and so is the ISOLATION, which is
the half that matters. Each blob stays an order under `MAX_TRANSFER_BODY_BYTES` (8 MiB) and
the blob count three orders under `MAX_TRANSFER_BODIES` (4096), so a refusal is attributable
to the decoded-closure cap rather than to a neighbour. An arm that crossed a different cap
would refuse for a reason this harness is not about, and its refusal would read exactly like
the one it is about.

Content is deterministic and distinct per blob. Distinct because `validate_pack` deduplicates
bodies by hash, so a fixture of identical blobs would collapse to one body and measure
deduplication while still looking like 24 MiB on disk. Deterministic because the reassembly
comparator hashes against the bytes the fixture wrote, and a fixture that differs run to run
cannot be compared to anything.

## The pin, and what it pins

`a_payload_bearing_push_bootstraps_under_the_cap_and_is_refused_over_it` drives both arms over
a real socket at an empty hosted-shaped peer, and pins the DIVERGENCE kin#1156 created. Both
arms were measured before anything was pinned, by reporting rather than asserting:

```
MEASURE over:  status=422  invalid repository transfer: the next publishable step of this
               transfer is one exact change and it carries a 16777250 byte source body
               closure, over negotiated limit 16777216; a single change cannot be split
               across continuation packs
MEASURE under: status=200  {..."git_authority_delta":{"old":null,"new":{...}},
               "roots_before":{"generation":0},"roots_after":{"generation":1}...}
```

The under-cap arm PUBLISHES, and the receipt is read rather than the status.
`git_authority_delta.old == null` with `.new` present is authority being established on a
replica that had none; roots moving from generation 0 to 1 is the destination really having
been empty. A 200 alone says only that the route answered.

The over-cap arm is REFUSED BY NAME, 422, naming the negotiated byte limit and the indivisible
step, plus an assertion that the Git-authority bound does not come back.

Both halves are needed. Asserting only the publication would pass on a build that had also
stopped refusing the over arm; asserting only the refusal would pass on a build that refused
everything, which is what the 409 did the day before.

`a_payload_bearing_push_under_a_minted_identity_is_refused_on_identity_instead` is what makes
that evidence rather than a restatement of "pushes fail". Same payload, same route, one
property changed, and it requires a DIFFERENT answer: the identity refusal naming the minted
id, and explicitly not the authority bound. If the two ever agreed, the pin would be measuring
the route's willingness to refuse anything.

### The forward-looking assertion in the first draft could not fail

It asserted the over-cap arm must not report `decoded body closure exceeds`, and its doc
comment said the arm moving to that string is how the pin goes red once Git-authority
bootstrap lands. It would not have.

That wording is `validate_pack`'s and is unreachable from a source-built pack. The source's
own byte check returns `Assembled::OverNegotiatedBound` (`repository_transfer.rs:949`), which
is SEGMENTATION rather than refusal: the segment builder halves the change budget and
rebuilds. It refuses only when the step is indivisible, and a one-commit fixture is, so it
refuses with a different sentence. The assertion was keyed on a string that can never appear,
so it would have stayed green through exactly the change it was written to catch.

The tell is the STATUS. `RepositoryTransferError::Invalid` maps to 422 and `Conflict` to 409
(`repository_transfer_error`, `api.rs:10046`), so 409 today and 422 once the bound moves
separates them with nothing to spell. The two sentences the size bound can actually produce
are asserted beside it as the weaker reading, so whoever hits the failure is told which
wording to look for.

### The number in the refusal is a lower bound, not the payload

A 24 MiB fixture refuses at **16777491 bytes**, 275 over the cap. `assemble_segment_pack`
returns the moment the running total crosses, so it names the total at the crossing and stops
counting. An assertion that it equalled the payload would fail against correct code.

### Two public builders refuse the same condition with different sentences

`build_repository_transfer_pack` maps `OverNegotiatedBound` straight to `Err(invalid(reason))`
with the short reason. `build_repository_transfer_segment`, which is what `push_to_remote`
calls, adds the indivisible-step wording. The first draft of the size probe called the sibling
and got `carries a 16777491 byte source body closure, over negotiated limit 16777216` with no
mention of the indivisible step. Probing the wrong builder pins wording no push can produce,
which is the same class as the assertion above.

Through the push path the refusal is, captured verbatim by a temporary assertion flip whose
restore is proved by `git diff --quiet` rather than by the word restored:

```
invalid repository transfer: the next publishable step of this transfer is one exact change
and it carries a 16777491 byte source body closure, over negotiated limit 16777216;
a single change cannot be split across continuation packs
```

## The order, measured rather than inferred

Nothing had established which bound a payload-bearing push meets first. The over-cap arm
observes the AUTHORITY refusal while being, on its own arithmetic, half again too large to
ever fit. So the authority comparison runs first and the size bound is unobservable end to
end until it moves.

That is why `the_size_bound_answers_for_real_once_the_authority_bound_is_out_of_the_way`
exists. It builds the expectation from the source store's own transfer status, so the
authority hash agrees with itself, sets the destination unborn the way an empty hosted store
reads, and runs everything downstream for real against a real payload. Without it the size
bound's behaviour would have been an assertion nobody had ever run, first observed on the day
it mattered.

## The destination check, which this harness was built to be ready for

Until kin#1156 no route put a payload on a hosted replica, so the comparator ran against the
source and proved only that it could answer. It now reads the hosted replica's OWN authority
snapshot and requires it to name every payload body by content address with the same declared
length, while the source side independently verifies what those addresses' bytes are with
SHA-256, a different hash function from the address, so the two cannot agree by construction.

It cannot use the workspace-walking comparator. A bootstrapped hosted replica commits with
`workspace_mutation: null`, so `ActiveApiRepositoryAuthority::open` refuses it by name with
"local daemon is missing its startup workspace binding". That refusal is the replica behaving
correctly, and reading it as a failure was the easy mistake here.

### Three findings from falsifying it, all against my own work

**It passed when pointed at the source.** A Git-admitted store names its own bodies from
creation, so an after-only reading is satisfied by either replica. Fixed with a before-reading
requiring the hosted replica to name zero external objects first.

**Then it passed again with only the after-reading retargeted.** Before from an empty replica,
after from a full one, no single replica described by either. Two correct readings that
jointly guarantee nothing, because the property lives in their agreement and neither endpoint
can see it. Fixed by naming the observed replica once, so a change of target moves both halves.

**Then the arm could not compile,** because the binding sat above the source state, and a
build failure grades nothing. The driver refused to tally rather than counting it. Moving the
declaration below the source made the mutation compiling, and the arm lands red on the before
assertion.

A fourth arm came out as malformed rather than passing: it mutated the receipt ASSERTION to a
weaker one and the test then passed, which is what weakening an assertion always does. An arm
has to remove a property of the thing under test, not of the test.

## The reassembly comparator

Byte-verified by content hash against the bytes the fixture wrote, never by count, because a
replica holding the right number of artifacts with the wrong bytes passes any count. The
expected side is a SHA-256 taken when the bytes were written; the actual side is a SHA-256 of
what `load_source_blob` returns. Hashing both with Kin's own digest would compare the store to
itself.

It is route-agnostic on purpose: it takes an authority and a digest map, so the day a route
lands a payload on a hosted replica it is pointed at that authority with the same map. Until
then it runs against the source replica, because a comparator first run on the day a route
lands is one whose first result nobody can tell from a bug in the comparator.

The design lane's seam-surface note arrived after this branch was gated. Processing it
repoints the comparator and refines the empty-replica precondition; it changes where the
comparator points and not what it asserts, so it lands as a follow-up rather than holding this.

## Falsification: eighteen arms, eighteen where declared

Re-run against the final head rather than left pointing at the sha it first passed on. Driver
archived at `.kin-coord/reports/citier1-fir2729-falsify-20260826.py`.

It restores on EXIT, INT and TERM; asserts each mutation's marker is in the tree before the
build starts; asserts the test actually RAN rather than reading an exit code, because
`--exact` with a bare name selects nothing and cargo exits 0; and grades any exit of 128 or
more as NOT RUN with no tally printed. `git grep KIN_FALSIFY_P` over `crates/` returns zero
afterwards.

| arm | check | declared | result |
|---|---|---|---|
| baseline x5 | all five | GREEN | GREEN |
| the over-cap arm no longer exceeds the byte cap | fixtures straddle | RED | RED |
| the two arms are edited to agree | fixtures straddle | BUILD FAILS | BUILD FAILS |
| a single blob crosses the PER-BODY cap | fixtures straddle | RED | RED |
| every blob carries identical bytes | push pin | RED | RED |
| the fixture stops adopting the hosted identity | push pin | RED | RED |
| the arms swap sides of the cap | push pin | RED | RED |
| the identity control adopts the peer's own id | identity control | RED | RED |
| the destination check reads the SOURCE's snapshot | payload pin | RED | RED |
| the comparator stops comparing bytes | comparator | RED | RED |
| the comparator skips a path it cannot find | comparator | GREEN | GREEN |
| both producers of the missing-path refusal removed | comparator | RED | RED |
| the size probe leaves the destination born | size probe | RED | RED |
| the over arm carries the under arm's payload | size probe | RED | RED |

### The three arms that are the useful part

**One mutation did not do what I believed.** The identity control's first mutation swapped a
non-matching hosted identity for a DIFFERENT non-matching one, so the property survived and
the arm read GREEN. A green from a mutation that removed nothing is indistinguishable from a
passing falsification. It now adopts the peer's own id.

**One guard was unreachable, so it could not be falsified at all.** The comparator refuses a
path the replica does not project, and no arm ever asked for one, so a mutation deleting that
guard changed nothing. Enlarging the mutation could not fix that; the test needed an arm that
reaches the guard, and now has one.

**One arm is declared GREEN on purpose.** A missing path has TWO producers of the refusal:
the lookup panics, and the count check catches the shortfall. Removing either alone leaves
the property held, so GREEN is correct for each, and a separate arm removes both and is RED.
That is what makes the GREEN a claim rather than an excuse.

### A third grade the driver had to learn

The first sweep against the final head returned NO TALLY, correctly. Taking clippy's finding
on the blob-count comparison moved it into a `const` block, so the mutation setting the two
arms equal stopped producing a red test and started producing a compile error, which the
driver grades NOT RUN because a build that failed for an unnamed reason grades nothing.

That arm is now two. One sets the over arm to 8 blobs, still greater than the under arm's 4,
so the const block is satisfied and the RUNTIME assertion answers. The other sets it to 4 and
is declared `BUILD FAILS`, a grade awarded only when the compiler output names the const
assertion, so a build broken for any other reason still reads NOT RUN. A mutation that trips
two guards at once proves neither.

## What I ran

`bin/kin-precheck kin --repo-dir kin=<worktree>`:

```
kin-precheck: repo=kin tree=.kin-coord/worktrees/citier1-kin sha=b920617ebf484f8e5976aa26d910e27d084bf5be branch=chore/lane-citier1-fir2729harness-kin
kin-precheck: 16 gates ran, 16 passed, 0 failed
```

It failed first, on clippy, on two findings both mine: a runtime assertion over two constants,
and a needless borrow of an owned `Vec`. Caught before this pull request, which is what the
gate is for. The first is a const block now rather than an allow, which is stronger than the
assertion it replaces.

`cargo test -p kin-daemon --no-fail-fast`, exit 0. Read the way a total has to be read: SIZE
from `-- --list`, 1097 tests in the inventory; VERDICT from what interleaving cannot corrupt,
zero `... FAILED`, zero `test result: FAILED`, zero `panicked at`, zero `error[E`, against a
positive control of 7 `test result: ok` lines that must be and is nonzero and a never-emitted
string that must be and is zero. All five new checks confirmed present by name in the `--list`
inventory with a control name that must and does return zero, because a total cannot tell a
test that ran from one the filter never reached.

`bin/kin-parity` is not run and not claimed. This diff is tests in one crate and builds no
binary the acceptance suites exercise differently. Hosted CI is the gate of record and the
Windows legs run only there.

## Tickets

Part of FIR-2729
Part of FIR-2746

Closing neither, deliberately. This records the bounds precisely; it does not move them.

## Claims not being made

That any of this unblocks the hosted republish. It does not.

That the size bound is the last one. FIR-2729's native-route bound is untouched here, and the
fixture is Git-admitted, so the Native admission rule is out of its scope by construction.

That the fixtures mirror kin's tree in shape. They mirror it in the one property the cap
reads, the decoded total, and deliberately not in blob count, because matching 842 blobs would
have brought the count caps into a test that is about bytes.

That the destination can SERVE the payload bytes on demand. What is proved is that its
committed authority names every one of them by content address with the right length. A body
read against a hosted backend is not reached here.

That anything ran against prod. Nothing did. No bucket object, no hosted daemon, no gpu,
daemon or quiet lock.

---

## Rebased and the fixture sizes derived, 2026-08-26 (wedge lane)

This PR was ejected from the merge queue four times with `failed_checks`, and the
failing test named its own mechanism:
`the_payload_fixtures_straddle_the_transfer_cap_and_nothing_else`.

**The harness was right and the harness caught it.** The straddle self-check
already reads the cap from the product constant rather than a literal, and #1160
moved `MAX_TRANSFER_DECODED_BODY_BYTES` from 16 MiB to 64 MiB. So the over arm's
24 MiB stopped being over the cap and the check failed on every speculative merge
tree, exactly as it was built to. What went stale was the FIXTURE SIZES, not the
cap. A group ejection leaves its reason only on the `merge_group` timeline, which
is why nobody watching the PR surface saw it.

Both arms now derive from the ceiling:

```rust
const CAP_BLOBS: usize = (MAX_TRANSFER_DECODED_BODY_BYTES / PAYLOAD_BLOB_BYTES as u64) as usize;
const PAYLOAD_BLOBS_OVER_CAP: usize = CAP_BLOBS + 1;
const PAYLOAD_BLOBS_UNDER_CAP: usize = if CAP_BLOBS / 2 < 4 { CAP_BLOBS / 2 } else { 4 };
```

so the straddle holds across any future cap change by construction rather than by
a literal somebody remembers to edit. The over arm is the minimum overshoot, one
blob past the ceiling.

**The comments changed because a claim in them stopped being true.** The arms were
originally sized to mirror real repositories: the over arm carried kin's own tip
tree at 27.77 MiB, which did not fit at 16 MiB, and the under arm carried kin-db
at 4.30 MiB, which did. kin's tip now fits comfortably. The arms cannot mirror a
repository any more, because the raised ceiling sits above every hosted
repository's tip closure, which is exactly what raising it bought. They mirror the
BOUND instead, and the comments say so rather than repeating the old claim.

## A third isolation assertion

The test had two isolation checks, that no blob crosses the per-body cap and that
the over arm stays clear of the body-count cap, both so a refusal here is
attributable to the byte cap alone. There is a third bound and nothing asserted
it: bodies travel base64, so an over-cap arm can be large enough that the
TRANSPORT refuses it with a 413 before the decoded-closure check runs, and that
refusal reads like the one this harness is about.

It fits today with 6.7 MiB of margin, and it is one legal edit from not fitting.
The assertion computes the wire size the way `http_body_limit_for` computes the
limit it is compared against, so the two cannot drift apart on what base64 costs.

## Cost, measured rather than assumed

Exceeding a decoded-byte cap means shipping that many bytes, so the over arm goes
from 24 MiB to 65 MiB and no construction avoids it.

**`a_payload_bearing_push_bootstraps_under_the_cap_and_is_refused_over_it` runs in
55.83 s**, and both payload tests together in 56.34 s. That is inside the bar, so
this ships plain rather than reaching for `#[ignore]`, because an ignored test on
hosted CI is a guard that never runs.

Shrinking the ceiling for the test was considered and rejected on evidence:
`configured_transfer_limits()` reads `std::env::var` at call time, so setting
`KIN_TRANSFER_MAX_DECODED_BODY_BYTES` in a test mutates process-global state
shared with every other test in the binary. That is a known race class and saving
test minutes is not worth introducing it.

## Falsification

Four mutations against the committed tree, each red at its own named assertion,
tree restored after every arm with zero markers and `git diff HEAD --quiet` clean.

| mutation | verdict | which assertion answered |
|---|---|---|
| the over arm reverts to the stale literal 24 | RED | `the over-cap arm must exceed the decoded closure cap: 25165824 vs 67108864` |
| the under arm is pushed past the over arm | **COMPILE-RED** | `the two arms must differ in blob count and nothing else` |
| the under arm lands exactly on the ceiling | RED | `the under-cap arm must fit under the decoded closure cap: 67108864 vs 67108864` |
| the blob grows to one byte under the per-body cap | RED | `refused by the decoded-closure cap, not by the transport: 100663284 base64 bytes against a 97867096 byte limit` |

Two of those are worth reading. The second does not fail at runtime at all: it
fails to COMPILE, because that pair is guarded by a `const` block the original
author added deliberately with a comment saying it makes the two arms agreeing a
compile error. That is a stronger guard than a runtime assertion, and the third
mutation exists because of it, landing the under arm exactly ON the ceiling so the
const block stays happy and the runtime assertion answers instead. Both guards are
now shown to work.

The fourth proves the new assertion is load-bearing rather than decorative.
Growing the blob to one byte under the per-body cap is legal against every other
check in the test, and only the new one catches it.

## Gates

`bin/kin-precheck kin` through `bin/kin-lane run` against the lane worktree:
**16 gates ran, 16 passed, 0 failed**. The first run was 15 of 16 with clippy
denying `manual_div_ceil` on the new assertion; fixed by using `div_ceil`, which
is what the product's own `http_body_limit_for` already uses.

`cargo test -p kin-daemon --lib api::tests::a_payload_bearing_push`: 2 passed,
0 failed, zero `test result: FAILED`, zero `... FAILED`, zero `panicked at`.

One note on running these locally: the leaf test name alone matches nothing under
`--exact`, and a filter that selects nothing exits 0 and prints `test result: ok`,
which reads exactly like a pass. The full path is
`api::tests::the_payload_fixtures_straddle_the_transfer_cap_and_nothing_else`.

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -17488,6 +17488,162 @@ mod tests {
         std::fs::remove_dir_all(&fixture.working).ok();
     }
 
+    /// What the transfer will say about these fixtures once the authority bound
+    /// moves, MEASURED rather than reasoned about.
+    ///
+    /// The push pin beside this cannot reach the size bound: the authority
+    /// comparison runs first and answers 409 for both arms. So the size bound's
+    /// behaviour would be an assertion nobody had ever run, written against a
+    /// route that does not exist, and the first time anyone saw its result
+    /// would be the day it mattered. That is the shape of an assertion that
+    /// turns out to have been wrong all along.
+    ///
+    /// This runs it today by removing the one thing that stops it. The
+    /// expectation is built from the SOURCE store's own transfer status, so its
+    /// `git_authority_hash` is the source's own and `TransferSourceContext::read`
+    /// agrees with itself; the destination is set unborn, which is what an empty
+    /// hosted store looks like. Everything downstream of the authority
+    /// comparison then runs for real against a real payload.
+    ///
+    /// What it establishes, and none of it was measured before:
+    ///
+    /// The under-cap arm assembles a pack. Its body closure covers every blob
+    /// the fixture wrote, verified by content hash against the source digests
+    /// rather than by counting bodies.
+    ///
+    /// The over-cap arm is refused, and refused by the INDIVISIBLE-STEP path
+    /// rather than by `validate_pack`. A one-commit history is one change, and
+    /// `Assembled::OverNegotiatedBound` is segmentation, so the builder has
+    /// nothing smaller to try. The sentence names `over negotiated limit`, which
+    /// is what the push pin's negative assertions are keyed on, so this is also
+    /// what proves those strings are the right ones to watch for.
+    #[cfg(unix)]
+    #[test]
+    fn the_size_bound_answers_for_real_once_the_authority_bound_is_out_of_the_way() {
+        let main = kin_model::RefName::branch(b"main").unwrap();
+
+        for (label, blobs, fits) in [
+            ("under", PAYLOAD_BLOBS_UNDER_CAP, true),
+            ("over", PAYLOAD_BLOBS_OVER_CAP, false),
+        ] {
+            let hosted_id = hosted_repository_id();
+            let hosted_repository = RepositoryId::new(hosted_id).unwrap();
+            let fixture = payload_fixture(label, blobs, &hosted_repository);
+            let authority =
+                ActiveApiRepositoryAuthority::open_layout_for_test(&fixture.layout).unwrap();
+
+            let status = kin_remote::repository_transfer::repository_transfer_status(
+                &authority.manager,
+                &hosted_repository,
+                &main,
+            )
+            .unwrap();
+            let mut expectation =
+                kin_remote::repository_transfer::RepositoryTransferExpectation::try_from(status)
+                    .unwrap();
+            // An empty hosted store's ref is unborn. Both fields move together
+            // because `validate_expectation` requires them both present or both
+            // absent, and a half-set destination would be refused for a reason
+            // that is not the one under test.
+            expectation.destination_target = None;
+            expectation.destination_head = None;
+
+            // The SEGMENT builder, because that is what a push calls
+            // (`push_to_remote` -> `build_repository_transfer_segment`). Its
+            // sibling `build_repository_transfer_pack` refuses the same
+            // condition with a shorter sentence, so probing that one would pin
+            // wording no push can produce. Measured, not read: the first draft
+            // of this test called the sibling and got
+            // `carries a 16777491 byte source body closure, over negotiated
+            // limit 16777216` with no mention of the indivisible step.
+            let built = kin_remote::repository_transfer::build_repository_transfer_segment(
+                &authority.manager,
+                &main,
+                &expectation,
+            );
+
+            if fits {
+                let pack = built
+                    .unwrap_or_else(|error| {
+                        panic!(
+                            "the {label} arm must assemble a segment once authority agrees: {error}"
+                        )
+                    })
+                    .pack;
+                let closure: u64 = pack.bodies.iter().map(|body| body.byte_len).sum();
+                assert!(
+                    closure >= fixture.payload_bytes,
+                    "the {label} arm's body closure must cover the payload it wrote: \
+                     {closure} against {}",
+                    fixture.payload_bytes
+                );
+                // By content hash, never by count. A pack carrying the right
+                // number of bodies with the wrong bytes passes any count.
+                let carried: std::collections::BTreeSet<String> = pack
+                    .bodies
+                    .iter()
+                    .map(|body| {
+                        hex::encode(Sha256::digest(
+                            &body
+                                .decode()
+                                .expect("a pack body decodes to its declared length"),
+                        ))
+                    })
+                    .collect();
+                for (path, want) in &fixture.source_digests {
+                    assert!(
+                        carried.contains(want),
+                        "the {label} arm's pack carries no body whose bytes hash to the \
+                         fixture's {path}"
+                    );
+                }
+            } else {
+                let error = built.err().unwrap_or_else(|| {
+                    panic!(
+                        "the over-cap arm must be refused once authority agrees, or the \
+                         cap FIR-2746 records does not bind"
+                    )
+                });
+                let message = error.to_string();
+                assert!(
+                    message.contains("over negotiated limit"),
+                    "the {label} arm must name the negotiated byte limit: {message}"
+                );
+                assert!(
+                    message.contains("cannot be split across continuation packs"),
+                    "a one-commit history is one change, so the refusal must be the \
+                     indivisible-step one rather than a segmentation retry: {message}"
+                );
+                // The number the refusal reports is a LOWER bound, not the
+                // payload. `assemble_segment_pack` returns the moment the
+                // running total crosses the limit, so it names the total at the
+                // crossing and stops counting. Asserting it equalled the
+                // fixture's 24 MiB would fail against a correct implementation.
+                assert!(
+                    message.contains(
+                        &kin_remote::repository_transfer::MAX_TRANSFER_DECODED_BODY_BYTES
+                            .to_string()
+                    ),
+                    "the refusal must name the limit it enforced, which is what makes \
+                     the number in it readable: {message}"
+                );
+                // The variant decides the HTTP status the push pin watches, so
+                // it is asserted here rather than left to be assumed there.
+                assert!(
+                    matches!(
+                        error,
+                        kin_remote::repository_transfer::RepositoryTransferError::Invalid(_)
+                    ),
+                    "the size bound must be Invalid, which maps to 422 and is what \
+                     separates it from the authority bound's 409: {message}"
+                );
+            }
+
+            drop(authority);
+            std::fs::remove_dir_all(&fixture.working).ok();
+        }
+    }
+
     /// Two replicas of ONE repository: a real local one with a graph-owned
     /// workspace projecting into a working directory, and a peer that forked
     /// from it and has since moved `refs/heads/main` ahead by one exact edit.
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -17185,16 +17185,34 @@ mod tests {
     /// for bound one, because bound two would keep it failing. Pinning WHICH
     /// bound answered is what makes this go red at the right moment, and going
     /// red is what it is for: when Git-authority bootstrap lands, the under-cap
-    /// arm should publish and the over-cap arm should move to
-    /// `decoded body closure exceeds`, and both of those are this test failing.
+    /// arm should publish and the over-cap arm should refuse on size, and both
+    /// of those are this test failing.
     ///
-    /// It also establishes the ORDER, which nothing had measured. `validate_pack`
-    /// enforces the byte cap on the destination's side of a pack that the
-    /// source only builds after `TransferSourceContext::read` has already
-    /// compared the two replicas' Git authority, so the authority bound is
-    /// reached first and the size bound is unobservable end to end until it
-    /// moves. The over-cap arm proves that by observing the authority refusal
-    /// while being, on its own arithmetic, half again too large to ever fit.
+    /// **The tell is the STATUS, not a sentence, and the first draft of this
+    /// test got that wrong in a way that could not fail.** It asserted the
+    /// over-cap arm must not report `decoded body closure exceeds`, which is
+    /// `validate_pack`'s wording. A source-built pack never reaches that line.
+    /// The source's own byte check returns `Assembled::OverNegotiatedBound`
+    /// (`repository_transfer.rs:949`), which is SEGMENTATION rather than
+    /// refusal: the segment builder halves the change budget and rebuilds. It
+    /// only refuses when the step is indivisible, and a one-commit fixture is,
+    /// so it refuses at `:601` with a different sentence naming
+    /// `over negotiated limit`. The original assertion was keyed on a string
+    /// that can never appear, so it would have stayed green through exactly the
+    /// change it was written to catch.
+    ///
+    /// `RepositoryTransferError::Invalid` maps to 422 and `Conflict` to 409
+    /// (`repository_transfer_error`, `api.rs:10046`), so the status separates
+    /// the two bounds with nothing to spell: 409 today, 422 once authority
+    /// bootstrap lands. The message assertions are kept beside it as the
+    /// second, weaker reading.
+    ///
+    /// It also establishes the ORDER, which nothing had measured.
+    /// `build_repository_transfer_segment` reads the source context, and its
+    /// Git-authority comparison, before any byte is counted, so the authority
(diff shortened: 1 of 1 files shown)


troyjr4103:
Ejected from the merge queue, recorded here because the queue does not mark the PR it ejects and this one still reads CLEAN and OPEN with `mergeQueueEntry: null`.

Found from the outside, by lane crossrepo, while checking why PR 1168 was sitting at position 1. The `merge_group` run at 20:50:36 (run 33012426178) built this entry alongside 1168 and failed:

```
test api::tests::a_payload_bearing_push_bootstraps_under_the_cap_and_is_refused_over_it ... FAILED
assertion `left == right` failed: the over arm must sit on the side of the cap its name claims:
25165824 bytes against 67108864
test api::tests::the_payload_fixtures_straddle_the_transfer_cap_and_nothing_else ... FAILED
```

Failed on `Check & Test ubuntu shard (1)`, `Check & Test macOS shard (1)` and `Feature permutation tests (macos-latest)` under `Test gcs feature`, so it is not one platform.

Both tests are added by this PR and exist nowhere else: `gh pr diff 1157 | grep -c` returns 2 for them, and the same grep against PR 1167 returns 0 as a negative control. Neither name exists on `origin/main` at `36d4ed689`, where a `--exact` run filters to zero tests, which is why the failure could not be attributed to main.

The reading is the fixture, not the assertion: the arm named "over" measures 24 MiB against a 64 MiB cap, so it sits under the bound its name claims. A fixture sized in one unit against a cap defined in another is the usual cause.

Worth knowing that this PR's own checks were green. The merge queue builds a speculative merge and runs the suite there, which is where it went red, so nothing on the PR surface shows it. Re-arming without changing the fixture will eject it again.

No action taken on this PR beyond this comment; it belongs to its own lane.

PR #1168, opened 2026-08-26
Weigh cross-repo authority in the verdict, not only when an absence is claimed

troyjr4103:
A find_references answer that returned rows is presented as the reference set. When the configured cross-repo spine went stale or answered incompletely, every cross-repo row is missing from that set and nothing marked it partial. The verdict read six inputs and none of them can see the spine, so the answer came back certified. That is find_references certifying an answer as the whole set while recording that a class is absent, one level up, where the class is every other repository.

The absence gate already weighs cross-repo authority, and weighs it well, but it is scoped to answers that claim an absence, and that scoping is right for what it was written for. This adds cross_repo as its own named verdict input so the fact reaches a populated answer too. The rule is not restated anywhere: negative.rs owns it and now exposes it, so the gate and the verdict can never disagree about what counts as a gap, and because the clause text is identical the composer's label dedupe collapses them to one clause on an answer where both speak.

Three states stay silent, because they are facts about an install rather than limits on an answer: no spine configured, a repository the spine has not registered, and a caller that never bound to a repository. The third is the one worth naming. kin refs resolves its binding from KIN_REPO_ID, which is an override rather than something an install sets, so every CLI reference answer on every machine reports an uncoded unavailable. Weighing that would mark all of them a lower bound, which is exactly the regression the gate's own scoping comment warns about. The gate keeps its stricter reading on the absence path, where an answer asserting that nothing references a symbol while reporting no cross-repo authority is what it exists to refuse.

Three test fixtures gain the anchor a real answer always carries. The producer always writes authority_anchor beside the roots and always writes focal_entity, and the gate has always required the anchor to name the focal the query asked about, so a fixture carrying roots and no anchor described a response the handler cannot emit. One of those fixtures says in its own comment that it means to be complete.

## Falsification

Seven mutations against all five tests in the class, on the committed tree, each proven applied by finding its marker before the build, each restored from an explicit source under an EXIT, INT and TERM trap that prints the sha it restored to and the marker count, which read zero every time. No row is green across.

| mutation | tests it turns red |
|---|---|
| the gap reading goes silent | both lower-bound tests |
| the input is unwired from the readings | both lower-bound tests, the healthy-states control, the certifying control |
| a note refuses instead of staying silent | the healthy-states control, on the first two states |
| the unbound-caller carve-out is removed | the healthy-states control, on the third state alone |
| a note certifies instead of staying silent | the healthy-states control, on its silent-input assertion alone |
| the gap clause takes a different label | the dedupe test |
| the unregistered carve-out is removed | the healthy-states control, on the second state alone |

The healthy-states control asserted in place inside its loop at first, so the first case failing ended the test and the two after it never ran. Two mutations both fired on the first case and stopped there, which left the middle state defended by nothing. It now collects and reports every state, and the last three rows above are what that bought: each names a different state or a different assertion, so no arm is covered only by an earlier arm failing sooner.

The driver's own marker guard was falsified too. A no-op mutation makes it abort that arm and exit nonzero rather than running the arm green. It needed the check: with two files under mutation, grep -c prints a per-file count, so the guard had been erroring and falling through.

## The sixth test, added after the matrix

The dispatch names two tools and the absence gate names the same two, so those strings are written twice in one file. A rename that moved one and not the other would leave the `bulk_check_references` arm matching nothing, and no `find_references` test can see that: the verdict would simply stop weighing cross-repo authority on bulk answers, silently, with every assertion still green. That arm shipped with no coverage until this.

It now has its own case in both directions. Falsified by renaming the dispatch arm, which turns it red while the `find_references` lower-bound test stays green under the same mutation, so the two arms are independently covered rather than one riding on the other.

## What was run

kin-precheck against the lane worktree at 1440dbdd5, which enumerated 16 gates from ci.yml's own check job and reported 16 passed, 0 failed, 0 commits behind origin/main. cargo test -p kin-mcp: 685 passed, 0 failed. cargo test -p kin-daemon as the consumer. Hosted CI is the gate of record.

## What this does not do

It does not report cross-repo coverage on kin doctor or graph status, which is the other half of the ticket's acceptance. It does not touch the identity defect that leaves the spine empty in the first place, which is FIR-2763 and is held pending a decision on which side owns repository identity.

Part of FIR-2764

### DIFF
--- a/crates/kin-mcp/src/envelope.rs
+++ b/crates/kin-mcp/src/envelope.rs
@@ -2844,11 +2844,16 @@ mod tests {
                 // cross-repo authority is not the fully-resolved answer these
                 // cases are about: it would read as a floor for a reason the
                 // test never meant to exercise.
+                "focal_entity": { "id": "0195f2a1-0000-7000-8000-00000000f0ca" },
                 "cross_repo": {
                     "status": "available",
                     "authority_complete": true,
                     "authority_revision": "sha256:complete",
                     "authority_roots": { "local": "local-root" },
+                    "authority_anchor": {
+                        "repo_id": "local",
+                        "entity_id": "0195f2a1-0000-7000-8000-00000000f0ca",
+                    },
                 },
                 // Every class but the one the case varies is present, and the
                 // host can produce reference edges. Since FIR-2672 every
@@ -2902,11 +2907,16 @@ mod tests {
                 "references": vec![json!({"name": "extract_tags"}); 5],
                 "relation_kinds": ["calls", "imports", "references"],
                 "degradations": [],
+                "focal_entity": { "id": "0195f2a1-0000-7000-8000-00000000f0ca" },
                 "cross_repo": {
                     "status": "available",
                     "authority_complete": true,
                     "authority_revision": "sha256:complete",
                     "authority_roots": { "local": "local-root" },
+                    "authority_anchor": {
+                        "repo_id": "local",
+                        "entity_id": "0195f2a1-0000-7000-8000-00000000f0ca",
+                    },
                 },
                 "edge_coverage": {
                     "scope": "language",
--- a/crates/kin-mcp/src/negative.rs
+++ b/crates/kin-mcp/src/negative.rs
@@ -1856,7 +1856,7 @@ fn focal_is_method(payload: &Value) -> bool {
 /// cannot act on, describing a state the response already reports in full under
 /// `cross_repo`, teaches the reader to discount every limit including the real
 /// ones.
-enum CrossRepoQualifier {
+pub(crate) enum CrossRepoQualifier {
     /// The spine answered completely. Nothing to report.
     Complete,
     /// A configured spine failed, went stale, or answered incompletely. The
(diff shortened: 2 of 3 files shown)


troyjr4103:
Ejected from the merge queue at about 21:18Z, recorded here because the queue does not mark the PR it ejects and this one still reads OPEN with `mergeQueueEntry: null`. Confirmed across three polls reading `mergedAt` first, so it is an ejection rather than the one-poll window where the entry clears just before `mergedAt` appears.

Cause, from the `merge_group` run's own log rather than from this PR's checks, which are all green:

```
FAIL kin-core::init_phase_ladder_contiguous the_admission_ladder_accounts_for_the_whole_of_init
panicked at crates/kin-core/tests/init_phase_ladder_contiguous.rs:235:36:
admit the fixture repository: "this Git repository has 1 admission blocker(s):
  - Git lock /tmp/.tmpnlwnkg/source/.git/objects/pack/multi-pack-index.lock indicates
    concurrent repository mutation"
```

That is a git lock left in a temp fixture by a concurrent operation on a loaded runner, not a code defect. This PR changes three files, all in kin-mcp: `verdict.rs`, `negative.rs`, `envelope.rs`. It touches no kin-core file, nothing on the init or admission path, and nothing git-related, so it cannot reach that fixture.

The earlier attempt at 20:50Z failed for a different and also unrelated reason: PR 1157's two payload-transfer-cap tests, which are recorded on 1157.

Re-arming. Nothing on this PR changed, and nothing about the failure suggests it should.

PR #1170, opened 2026-08-26
Pin a sibling under its manifest identity instead of refusing it for having one

troyjr4103:
**Do not land this until the founder answers which side owns repository identity.** Opened as a draft so CI grades it and the answer is a one-click unblock. The reproduction, the fix, and both halves of the answer are here; what is not here is a decision.

## What is broken

The daemon refuses every sibling repository whose registry id differs from its manifest's `repo_id`. Those two strings are written by different producers in different alphabets. The manifest mints a UUID in `KinManifest::new`. The registry records the directory name in `kin_migrate::update_registry`, which is what `kin init` calls. A UUID is never a directory name, so the comparison cannot pass for any repository on any host, and cross-repo spine authority has been empty since the check landed on 2026-07-27.

The cost is larger than the absent siblings. One refused sibling sets the incomplete flag, and that invalidates the cross-repo edges of every registered repo including the primary, so a single mismatch marks the whole edge set dirty. With no sibling pinned the resolver's candidate set is empty too, which is the direct reason `cross_repo_edges` reads zero.

## Why five weeks of green tests never saw it

The existing spine test takes its sibling's registry id from `kin_core::init`'s minted manifest identity and writes that into the registry itself. One fixture writes both sides of the comparison, so they agree by construction, and the test passes on a tree where the two real producers disagree.

The first commit here changes exactly one thing about that fixture: it registers the sibling by calling the same `update_registry` that `kin init` calls, and reads the id back out of the registry rather than restating it. No repository identity is written twice, so it can only pass when the two producers agree. On unmodified main it fails with `repo_count` 1 where 2 is required.

## The two halves, separable on purpose

The daemon commit makes every existing install work by reading the manifest at startup, so there is no migration and no user has to run anything. It also says nothing about which identity the spine keys on: it asserts the consequence instead, that a sibling registered the way `kin init` registers one is pinned and its edges materialize. That is what lets the open question be answered either way without moving the check.

The registry commit stops new rows being written in the wrong alphabet, so the two sides agree at the source rather than only after the daemon corrects for it. It adds one function in kin-core for the question both producers were answering separately, because a second copy of that rule is how they came apart.

## One reading that bears on the decision

Keying the spine on the registry label also works, once the refusal is gone. A falsification arm that does exactly that leaves both spine checks green, which is the check proving it does not smuggle in an answer. That same arm turns the duplicate-identity test red, and that is the finding: the manifest identity is what makes duplicate detection possible at all. Two registry rows pointing at one repository are invisible when the key is the label, because two labels never collide.

## Falsification

Two matrices, both on the committed tree, markers proven present before each build, restore from an explicit source under an EXIT, INT and TERM trap.

Daemon side, five arms against four tests. Restoring the identity refusal turns three red. Pinning no siblings turns all four red. Removing the duplicate refusal turns only the duplicate test red. Tightening it into refusing every sibling after the first turns only its healthy control red. Keying on the registry label is the control described above.

Registry side, three arms against three tests. Recording the directory name again, dropping the no-manifest fallback, and having the shared authority answer with the directory name each turn exactly one test red, and a different one each time.

An earlier duplicate-guard arm came back green, and the reason is on the record in its own commit: it refused from the second sibling onward and the fixture has one sibling, so the mutation applied, changed the code, and left the property it targeted standing. The guard it was written for had no test at all. It has two now, the refusal and its healthy control.

The registry driver's own marker guard earned itself in the same run. Two arms whose patterns missed after rustfmt were aborted rather than run green, and the driver exited nonzero.

## What was run

`kin-precheck` against the lane worktree at be49c5bbc, 16 gates enumerated from ci.yml's own check job, 16 passed, 0 failed, 0 commits behind origin/main. The kin-daemon, kin-migrate and kin-core suites. Hosted CI is the gate of record.

## What this does not do

It does not reap the 3,014-entry registry whose 94 live paths this host carries, which is FIR-2767. It does not touch the three-attempt spine give-up under a writer, which is FIR-2766. It does not disclose an empty spine on any surface, which is FIR-2764 and is PR 1168.

Part of FIR-2763

---

## The eager-sibling bound, and what measuring it changed (wedge lane, 2026-08-26)

The identity fix above re-arms a serial pass that opens every pinned sibling's
whole workspace graph. This adds the bound the founder required before it lands,
plus the disclosure that makes a bounded answer say so.

**Two states share one inequality, and conflating them is the hazard.** A capture
the bound stopped is BOUNDED: deliberate, and its captures are sound. A sibling
that failed to load is INCOMPLETE: a shape nobody chose. They get separate signals
at the source. `capture_set_complete` keeps meaning nothing failed, and the
bounded fields carry captured-of-registered counts beside it.

**A silent cap would have been the worse bug.** Skipped siblings never register,
so the completeness comparison sees equality and reports a complete capture set
over a subset nobody was told about.

**`/spine/health` gains a third reading, because two could not see this state.**
A capture stopped at its bound is complete by BOTH existing fields: nothing
failed, and every registered sibling pinned. Yet siblings were never consulted.
Without the third reading this endpoint reports a bounded answer as healthy, which
is the bare zero it exists to prevent arriving from a new direction. The healthy
control asserts it false on a daemon that capped nothing.

**Invalidation is scoped where it was blanket**, which fixes a live defect rather
than a hypothetical: one sibling failing used to mark every registered repo's
cross-repo edges dirty including the primary. Edges between captured repos were
recomputed from those captures moments earlier and are sound.

### What the measurement said, and it refuted the premise

Run under the daemon lock against this host's real registry. Debug profile: the
SHAPE is the finding, the seconds are not a product claim.

```
bound | seconds | captured/registered
    0 |    0.00 |                0/10
    1 |  123.00 |                1/10
    2 |  123.16 |                2/10
    4 |  124.47 |                4/10
```

**The cost is not per sibling.** One to four adds 1.47 s against a 123 s first
sibling. This work was sized on "81 siblings times 0.87 GiB, serial, so bound the
count". The count is nearly free.

**And ten siblings pin, not 81.** The registry holds 3016 entries, 81 have a
`.kin` directory, 80 a readable manifest, and the daemon pins ten. The 81 that
sized this work is the directory count.

Per-sibling attribution, and a permutation to separate one-time initialization
from one expensive repository, because a slow FIRST sibling predicts both:

- forward: first-loaded `fastapi` 192.56 s, other nine 0.01 to 1.63 s
- reversed: first-loaded sibling 0.02 s, and `fastapi` still 120.47 s from last

The cost follows the repository, not the position. `fastapi`'s store is 1.89 GiB
against a next-largest of 0.03, sixty-three times the next, at one hundred and
eighteen times the load. **It is size, not a pathology.** Filed as FIR-2789.

### The default stays 16, with its justification rewritten

16 was chosen to bound a per-sibling cost that does not exist. It stays, as a
**safety cap that makes startup bounded rather than a tuned per-sibling budget**,
because the only materially different settings are zero and not-zero, and zero
turns the spine off, which is the defect FIR-2763 exists to end. Saying so beats
letting the old arithmetic be inherited as a measured number.

### Parked, deliberately

Moving spine init off the startup path to pay lazily on first cross-repo query was
approved tonight on the reading that the 123 s was one-time initialization. The
permutation above refuted that premise, the ruling was pulled back before it was
spent, and lazy-loading a large store relocates a three-minute wait onto whoever
asks the first cross-repo question. **The trigger is not built here.** The bound
and the disclosure stand on their own: a silent subset was the real defect and
that fix needs no arithmetic to justify.

### Falsification

Each state falsified by the other's mechanism, since two producers of one
inequality is the trap class this defect came from. Tree restored after every arm,
zero markers, `git diff HEAD --quiet` clean.

| mutation | target | verdict | the other check |
|---|---|---|---|
| the bound stops capping | bounded | RED | ok |
| a failed load reports itself as bounded | incomplete | RED | ok |
| a cap reports itself as incomplete | bounded | RED | ok |

**Each mutation reddened only its target**, which is what proves the two checks
are independent rather than both keying on the inequality.

### Gates

`bin/kin-precheck kin` through `bin/kin-lane run`: **16 gates ran, 16 passed, 0
failed**. The spine and bound tests: 7 passed, 0 failed.

Part of FIR-2763
Part of FIR-2772


troyjr4103:
Founder decision, relayed by captain-bb so it cannot be missed in message crossings: SHIP BOTH COMMITS. The identity question is settled (the manifest UUID is canonical; the registry commit ships too; the path-beside-id listing is the approved follow-up). The ONE gate before this PR lands: the daemon commit must carry the conservative configurable bound on eager sibling loading with the overflow disclosed (spine limited to N of M), per FIR-2763's requirement, because the identity fix re-arms the serial sibling-load term at real-host scale (81 siblings, 70.67 GiB here). Undraft and land once the bound is in. Captain messages of 20:47Z and 20:52Z carry the full reasoning.

PR #1179, opened 2026-08-27
Take kin-db 0.7.67 for the conversion persist cut

troyjr4103:
kin pinned `=0.7.65`. kin-db 0.7.67 is published and carries FIR-2683's cut to
the snapshot persist.

**The persist-side effect is fir2651's measurement, not mine:** that lane
measured `kindb.commit.persist_successor` at 1260.8 MiB grown and called it the
largest single remaining term in a conversion. My own FIR-2782 attribution
prices the plan path and never runs a commit, so it says nothing about persist
either way, and I am not the evidence for this pin.

The case for taking it is conversion-side. The stranger container that OOM-killed
a one-file commit under FIR-2782 had converted first, with `memory.peak` sitting
exactly at its 12 GiB cap and 1861 reclaim events during setup alone, before the
commit that died. That is where this headroom is spent.

What 0.7.67 does **not** fix, so nobody reads more into this than it carries: the
planner's two whole-store clone sites are untouched. On kin-db main the `Carried`
site is still at `repository.rs:5068` with the clones at `:5380` and `:5384`,
substantively identical to 0.7.65's `:5014`, `:5326` and `:5330`, and the only
commit touching that file between the two versions is #240, a gitlink change.
Those clones are FIR-2808.

## The pin move, and the two traps it walked into

**An exact requirement refuses `--precise`.** The catalogue's sequence, run
`cargo update -p kin-db --precise <ver>` while the manifest still carries the old
requirement, is for a loose requirement. Against `=0.7.65` cargo refuses with
"candidate versions found which didn't match: 0.7.67" and leaves the lock
untouched, which I verified rather than assumed. So the manifest moved first, and
the exactness is also what removes the hazard that sequence exists to avoid,
since a one-member range has no maximum to land on by accident.

**A stray hunk rode along and was removed.** The re-resolve opportunistically
moved `winapi-util 0.1.11`'s edge from `windows-sys 0.48.0` to `0.61.2`, a
Windows-only change this host cannot build, inside what should be a two-line pin.
It was reverted, and `cargo metadata --locked` then accepts the lock, which
proves the bump did not require it.

## The lock diff, as its own artifact

```
 [[package]]
 name = "kin-db"
-version = "0.7.65"
+version = "0.7.67"
 source = "sparse+https://kinlab.ai/registry/cargo/"
-checksum = "82aeb8e23d98d55e8d98fe4c3aac9fb76c0776385bdab7df927a1465dade9941"
+checksum = "1b11a925682d8fe6114ae50bb96a1b55eb19babe7520f2efb77402f39cc344ca"
```

`git diff --numstat -- Cargo.lock` is `2 2`. Two versions, two checksums, nothing
else. The entry keeps its `sparse+` source and a checksum.

**Both tracked lockfiles swept.** This workspace carries two, `Cargo.lock` and
`fuzz/Cargo.lock`, and a root re-lock alone would ship fuzz targets on the old
version. Both read zero occurrences of `0.7.65`, with the new version present in
the root lock as the positive control. `fuzz/Cargo.lock` carries no kin-db entry
at all, so it needed nothing.

Every workspace crate takes `kin-db = { workspace = true }`, so the requirement
has exactly one manifest home, `Cargo.toml:158`, which the sweep confirms.

## Gates

Gated with `--locked` on purpose: the DEV-LOCAL gate replaces the crate under
test with a local path and therefore cannot validate a registry pin at all.

Part of FIR-2782. Part of FIR-2651.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3285,9 +3285,9 @@ dependencies = [
 
 [[package]]
 name = "kin-db"
-version = "0.7.65"
+version = "0.7.67"
 source = "sparse+https://kinlab.ai/registry/cargo/"
-checksum = "82aeb8e23d98d55e8d98fe4c3aac9fb76c0776385bdab7df927a1465dade9941"
+checksum = "1b11a925682d8fe6114ae50bb96a1b55eb19babe7520f2efb77402f39cc344ca"
 dependencies = [
  "bincode",
  "bstr",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -155,7 +155,7 @@ kin-lsp = { version = "=0.7.0", registry = "kin" }
 # backend on every target (incl. Linux, where it cannot compile). Metal is re-enabled
 # additively on macOS via a `[target.'cfg(target_os = "macos")'.dependencies]` block in the
 # binary/runtime crates (kin-cli, kin-daemon). Feature unification makes that additive.
-kin-db = { version = "=0.7.65", registry = "kin", default-features = false }
+kin-db = { version = "=0.7.67", registry = "kin", default-features = false }
 kin-vfs-core = { version = "=0.4.14", registry = "kin" }
 
 [profile.release]


troyjr4103:
**Blocked. Do not merge.** kin-db 0.7.67 breaks a kin safety refusal, measured
with a two-arm control that holds everything fixed except the pin.

| arm | result |
|---|---|
| `=0.7.67`, this branch | **FAILED** in 1.28s |
| `=0.7.65`, `origin/main` | ok in 1.47s |

Same test, same runner, same box, one variable. The test is
`kin-remote repository_transfer::tests::a_bootstrap_carrying_native_history_over_a_gitlink_is_refused_by_name`.
kin asserts that native history carried over a gitlink cannot bootstrap. Under
0.7.67 the transfer returns `outcome: Committed` on a tree whose entry is
`mode: Gitlink`, so the refusal never happens. It fails in isolation, so this is
deterministic and not a parallelism artifact.

The likely mechanism, from the changelog rather than from reading the diff:
kin-db#240, "Authenticate a gitlink against the authority the same transaction
installs", landed in 0.7.66, inside the span this pin crosses.

The arm-B swap restored to `=0.7.67` with a clean tree, printed as the restored
value rather than as the word "restored", because a driver that stages anything
cannot restore with a bare `git checkout --`.

### Second failure, deliberately NOT attributed

`kin-core::init_phase_ladder_contiguous the_admission_ladder_accounts_for_the_whole_of_init`:

> init_from_git ran for 3790 ms with no admission phase open, 8.8% of its
> 43072 ms, over the 2.0% this guard allows. The stretch sits immediately before
> `the end of init, with no phase open`.

It reproduced in both full-workspace runs at ~71s and passes alone at 8.7s, so it
is workspace-parallelism-shaped. No two-arm control was run on it, so it may be
pre-existing and this PR does not claim otherwise. Worth its own look: the traps
catalogue already carries an entry about this guard's blindness to its own tail,
and this failure is a tail stretch.

### What is still good here

The pin mechanics are clean and stay useful whenever the gitlink question is
settled: the lock diff is `2 2`, one version and one checksum, `sparse+` source
and checksum retained, both tracked lockfiles swept to zero occurrences of the
old version, and the stray `winapi-util`/`windows-sys` hunk that rode along on
the re-resolve was reverted and proven unnecessary by `cargo metadata --locked`.

Someone has to decide whether kin's guard is now wrong because the refusal moved
layers, or whether 0.7.66 lost a safety check. That is a cross-repo question.

Part of FIR-2782. Part of FIR-2651.

troyjr4103:
Superseded by #1182: this pinned kin-db 0.7.67 without the guard rework FIR-2810's verdict requires, so it can never go green. #1182 carries the same pin plus the corrected landing guard, the new hostile-pack refusal guard, the pull path's admit step, and the capability fixture, with the six-arm falsification in its body.

PR #1225, opened 2026-08-28
FALSIFICATION ONLY: prove the widened kin-vfs guard is red on a pull request

troyjr4103:
Temporary. Proves FIR-2887's third falsification arm: a deliberate kin-vfs pin disagreement planted in rc-build.yml must turn the fast gate lint and policy check red ON A PULL REQUEST, not only on a push to main. Closed as soon as the check reports. Do not land.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -930,6 +930,21 @@ jobs:
       - name: Check the flake quarantine
         run: python3 scripts/check-quarantine.py
 
+      # ADDED here rather than moved out of `check`, deliberately, for two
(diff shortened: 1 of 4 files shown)


troyjr4103:
Falsification arm 3 for FIR-2887 is proven and this branch has served its purpose. The planted rc-build.yml pin disagreement turned 'Fast gate lint and policy' FAILURE on a pull request, and the only failing step was 'Verify Kin/kin-vfs compatibility at the pinned release input'. Before this change that job did not run the guard at all and the check job that did is skipped on pull requests. Closing unmerged; never intended to land.

PR #1234, opened 2026-08-29
Gate #1217's feature-only items, and stop grading an unclassifiable pid as a killed daemon

troyjr4103:
kin `main` has been red since #1217 landed, and a second failure joined it with
#1232. Every push to `main` since has produced a red run, and each one reads like
new news. This fixes both.

Neither leg runs on a pull request, so this PR's own green does not prove either
fix. That is the same push-only-guard shape as FIR-2892, and it is why the
evidence below is local runs of the exact CI commands rather than a check name.

## The boundary, read from the runs

`gh run list --repo firelock-ai/kin --branch main --workflow ci.yml`, sorted by
`createdAt` explicitly, with each run's failing jobs read from its own `jobs`
array rather than from the rollup:

| created (UTC) | run | head | failing jobs |
|---|---|---|---|
| 2026-08-28T21:37:07Z | 33213340084 | `7fc7801d7` | none |
| 2026-08-28T21:43:06Z | 33213757329 | `0b2d92815` (#1217) | Windows authority tests, Windows authority runtime tests, Feature permutation tests (ubuntu-latest) |
| 2026-08-28T23:47:06Z | 33221540930 | `e7122ccb9` (#1229) | the same three |
| 2026-08-29T01:54:36Z | 33227618508 | `5088b0e4c` (#1232) | those three plus Windows installer + vector release build |
| 2026-08-29T04:42:43Z | 33234472343 | `a943eb20a` (#1230) | the same four |

`git log --oneline 7fc7801d7..0b2d92815` and
`git log --oneline e7122ccb9..5088b0e4c` are each exactly one commit, so both
attributions are single-commit ranges rather than a guess over a span.

## 1. Dead-code lints, from #1217 `0b2d92815`

`RUSTFLAGS: -D warnings` is workflow-level (`ci.yml:40`), so `dead_code` is
denied. Three jobs build `kin-daemon` with the feature set that makes #1217's new
items unreachable, and all three fail during the build:

```
error: function `hosted_vector_producer_profile` is never used
error: variant `ConflictReloading` is never constructed
error: could not compile `kin-daemon` (lib) due to 2 previous errors
error: variants `ConflictInstallWinner`, `ConflictWithoutCursor`, `IndeterminateInstalled`, `IndeterminateNotInstalled`, and `CommittedWrongDigest` are never constructed
error: methods `corrupt_vector_loads`, `vector_save_count`, `last_expected_vector_cursor`, `inject_next_save_fault`, `current_vector_artifact`, and `corrupt_current_vector_indexed_count` are never used
error: could not compile `kin-daemon` (lib test) due to 4 previous errors
```

Nothing here is Windows-specific. `cargo check -p kin-daemon
--no-default-features --lib --tests --locked` under `RUSTFLAGS="-D warnings"`
reproduces all six lines on macOS, exit 101.

Each item is gated on the feature that builds its producers, found with
`git grep` over every call site rather than by guessing:

| item | producers | gate |
|---|---|---|
| `hosted_vector_producer_profile` | `flush_embed_progress` (`embeddings`), `persist_vector_sidecar` (`vector`) | `cfg(any(feature = "embeddings", feature = "vector"))` |
| `HostedVectorPersistenceState::ConflictReloading` | `reload_hosted_vector_persistence_after_conflict`, `persist_hosted_vector_artifact`, both `cfg(feature = "vector")` | `cfg(feature = "vector")` on the variant and on its two reader arms |
| `DurableVectorSaveFault` | five `#[cfg(feature = "embeddings")] #[test]` fns | `cfg(all(test, feature = "embeddings"))` |
| the six harness methods | every call site is an `embeddings` test | `cfg(feature = "embeddings")` |

No `allow(dead_code)`, blanket or `cfg_attr`, was added.

Two items are in this change that the compiler had not reported yet. Gating
`vector_save_count` and `last_expected_vector_cursor` leaves the `saves` and
`last_expected_cursor` fields written but never read, which is the same lint one
step later, so those fields and their write sites inside `save_vector_artifact`
are gated with them. That is why `save_vector_artifact` grew five `#[cfg]`
attributes rather than one.

`DurableVectorTestBackend` itself stays ungated on purpose. Four tests drive it
with no feature gate at all, two in `state.rs` and two in `api.rs`, so gating the
whole struct would have been fourteen attributes cheaper and would have deleted
four featureless tests.

## 2. The Windows admission contract, from #1232 `5088b0e4c`

The failing step is `Assert the Windows admission contract` (`ci.yml:749`), and
its errors are:

```
##[error]Windows exact-Git admission failed
##[error]Windows native-unborn bootstrap failed
##[error]Windows admission violated its shipped contract (2 check(s) failed)
```

Both are `require_admitted` in `scripts/assert-windows-init-contract.sh`, which
fails when `kin init` exits non-zero. `.kin` was published in both cases, and the
captured output carries `a daemon serving this store was killed`.

(The `::error::Windows registry authority --fix must fail as unsupported` line
visible in that job's log at 02:17:41Z is the step's own echoed source, not an
emitted error. It carries the command-echo escape, it lives at `ci.yml:698` in
`Verify Windows build provenance and feature set`, and that step passed.)

#1232 changed `kin init` from returning `Ok(())` unconditionally to exiting
`EXIT_ENRICHMENT_UNATTESTED = 7` when the store can prove a daemon serving it was
killed. That decision is right and is unchanged here. What is wrong is what "can
prove" reads on a host whose liveness probe cannot answer:

1. `init.rs:265` calls `daemon_death::recorded_for_store`.
2. `daemon_death.rs:50` calls `kin_daemon_spawn::peek_unwatched_daemon_death`.
3. That opened with `if process_is_alive(serving.pid) { return None; }` and
   graded everything else as a death.
4. `process_is_alive` is `matches!(process_liveness(pid), Liveness::Alive)`.
5. `process_liveness` is `#[cfg(unix)]` only. Its `#[cfg(not(unix))]` arm returns
   `Liveness::Unknown` unconditionally.

So on Windows `process_is_alive` answers `false` for every pid including a
running one, and step 3's negation graded a kill for every store carrying a
serving record. `kin init` publishes a store and starts a daemon for the
enrichment sweep, so that is every `kin init` on Windows. Before #1232 the same
false reading only printed a warning; #1232 wired it to the exit status.

`process_is_alive`'s own doc comment already forbids this use: "`true` is
positive proof of life, and every other outcome is the absence of proof rather
than proof of death." Two call sites read it as the negation.

Both graders now delegate to `*_with(kin_root, process_liveness)` seams that
grade a death on `liveness(pid) == Liveness::Dead` and nothing else. The probe is
a parameter so a test can produce the `Unknown` no host here produces on demand.
This reaches `unix` too, where `EPERM` on a pid recycled under another user maps
to `Unknown` and was being graded as a killed daemon.

**What this does not restore.** Windows still has no daemon-kill detection, and
had none before: `DaemonWatch::record_exit` has a `#[cfg(not(unix))]` arm
returning `None`, and `cgroup_memory` reads Linux files. This removes a false
positive; it does not remove a working detector. Said plainly, because "Windows
exits 0 again" is only good news if nothing true went with the false part.

**Evidence caveat.** I cannot run Windows here, so the Windows half rests on
source reading plus the run-level differential above, not on execution. One
control I ran and am not citing: the passing installer job at `e7122ccb9` has
zero occurrences of "a daemon serving this store was killed", but the script only
`cat`s the captured `kin init` output when a check FAILS, so a passing run cannot
contain that line whatever the binary did. That absence measures the script's
logging, not the behaviour.

## Where the new test runs

`a_pid_this_host_cannot_classify_is_not_a_death` runs wherever
`cargo test -p kin-daemon-spawn --lib` runs, so on both Check & Test legs. It does
NOT run on Windows: `ci.yml:404` and `ci.yml:413` are the only two
kin-daemon-spawn legs on that target and they are filtered to
`managed_spawn_fence` and
`tests::a_detached_spawn_does_not_hold_the_callers_stdout_open`. The test proves
the logic on a host where `Unknown` is injected; the source proves Windows is the
platform that answers `Unknown` for every pid.

No Windows leg asserts the old behaviour either. `kin-cli`'s `daemon_death::tests`
build serving records at `pid: u32::MAX` and would go red on Windows under this
change, but `Windows authority CLI tests` runs only the filters at
`ci.yml:462-491`, none of which selects them.

## Verification

Every command below ran in the lane worktree through `bin/kin-lane run`, which
prints the path, sha and branch it resolved.

Compile, `RUSTFLAGS="-D warnings"`, the five flag sets CI builds:

```
WINFIX SUMMARY featureless=0 default=0 gcs=0 cli=0 spawn=0
```

Tests, the CI steps verbatim:

```
WINFIX2 ARM spawn-lib-tests rc=0              cargo test -p kin-daemon-spawn --lib
WINFIX2 ARM featureless-kin-daemon-tests rc=0 cargo test -p kin-daemon --no-default-features --lib --tests --locked
WINFIX2 ARM featureless-kin-cli-tests rc=0    cargo test -p kin-cli --no-default-features --lib --tests --locked
WINFIX2 ARM gcs-nextest rc=0                  cargo nextest run -p kin-daemon --features gcs --locked
WINFIX2 SUMMARY spawn=0 daemon=0 cli=0 gcs=0
```

Zero `test result: FAILED` and zero `FAIL [` lines across all four. The gcs arm's
own verdict block reads `1217 tests run: 1217 passed (3 slow, 2 leaky), 3 skipped`.
The new test executed rather than being filtered out:
`test tests::a_pid_this_host_cannot_classify_is_not_a_death ... ok`.

Falsification, run on the clean pushed tree with the unmutated baseline graded
first. Each arm removes a property of the product, never weakens an assertion,
and a lint arm is graded only when the LINT fired, since an arm that fails to
compile is its own grade and not a red:

```
CONTROL listing real=1 fake=0 (expect 1 and 0)
ARM M0 baseline featureless=0 livetest=0
M1 cfg-still-on-target=0 (expect 0)
M2 cfg-still-on-target=0 (expect 0)
M3 cfg-still-on-a-ConflictReloading-site=0 (expect 0)
M4 mutation sites applied=2 (expect 2)
VERDICT M0-baseline GREEN
VERDICT M1-corrupt_vector_loads RED-ON-TARGET rc=101 lint=corrupt_vector_loads
VERDICT M2-hosted_vector_producer_profile RED-ON-TARGET rc=101 lint=hosted_vector_producer_profile
VERDICT M3-ConflictReloading RED-ON-TARGET rc=101 lint=ConflictReloading
VERDICT M4-unknown-is-not-a-death RED-ON-TARGET rc=101
WINFIX-FALSIFY distinct-arms=5 (expect 5) fails=0
RESTORED to c737ef6fa; dirty=[]
```

M1 ungates one harness method whose body still compiles featurelessly, M2 the
producer-profile helper, M3 the `ConflictReloading` variant AND both of its
reader arms, which is the pre-fix tree for that item (ungating the variant alone
leaves the match non-exhaustive, a compile error rather than this lint). M4
removes the property the daemon-death fix adds, restoring
`liveness(pid) == Liveness::Alive`, and the test went red on its `Unknown`
assertion by name rather than on a later one.

`bin/kin-precheck kin`:

```
kin-precheck: 16 gates ran, 16 passed, 0 failed, on kin c737ef6faa920e259f0788425fa000b462be00a9
```

Refs FIR-2892

### DIFF
--- a/crates/kin-daemon-spawn/src/lib.rs
+++ b/crates/kin-daemon-spawn/src/lib.rs
@@ -1127,10 +1127,33 @@ pub fn read_serving_daemon(kin_root: &Path) -> Option<ServingDaemon> {
 /// attributes nothing either.
 ///
 /// A daemon whose pid is still alive is not settled: it is serving, and its
-/// record is doing its job.
+/// record is doing its job. Neither is one whose pid this host cannot
+/// classify, which is a third answer and not the second: see
+/// [`settle_unwatched_daemon_death_with`].
 pub fn settle_unwatched_daemon_death(kin_root: &Path) -> Option<DaemonKillRecord> {
+    settle_unwatched_daemon_death_with(kin_root, process_liveness)
+}
+
+/// [`settle_unwatched_daemon_death`] against a named liveness probe.
+///
+/// The probe is a parameter so a test can produce the answer a real host will
+/// not produce on demand, which is the `Unknown` this grading turns on.
+///
+/// A death is graded on `Dead` and on nothing else. `!process_is_alive(pid)`
+/// looks like the same test and is not: it folds `Unknown` in with `Dead`, so
+/// a host whose probe cannot classify a pid reports every serving record as a
+/// killed daemon. [`process_liveness`] is implemented for `unix` only and
+/// answers `Unknown` everywhere else, so on Windows that negation graded a
+/// kill for every store that had ever started a daemon, and `kin init` began
+/// exiting non-zero for it once FIR-2639 wired the record to the exit status.
+/// `unix` reaches it too, through the `EPERM` a reused pid owned by another
+/// user returns.
+fn settle_unwatched_daemon_death_with(
+    kin_root: &Path,
+    liveness: impl Fn(u32) -> Liveness,
+) -> Option<DaemonKillRecord> {
     let serving = read_serving_daemon(kin_root)?;
-    if process_is_alive(serving.pid) {
+    if liveness(serving.pid) != Liveness::Dead {
         return None;
     }
     retire_serving_daemon(kin_root);
@@ -1167,8 +1190,18 @@ pub fn settle_unwatched_daemon_death(kin_root: &Path) -> Option<DaemonKillRecord
 /// what the store WOULD record, not what the store has recorded, and a caller
 /// wanting the tally reads [`read_daemon_kill_record`] instead.
 pub fn peek_unwatched_daemon_death(kin_root: &Path) -> Option<DaemonKillRecord> {
+    peek_unwatched_daemon_death_with(kin_root, process_liveness)
+}
+
+/// [`peek_unwatched_daemon_death`] against a named liveness probe. Grades a
+/// death on `Dead` alone, for the reason
+/// [`settle_unwatched_daemon_death_with`] states.
+fn peek_unwatched_daemon_death_with(
+    kin_root: &Path,
+    liveness: impl Fn(u32) -> Liveness,
+) -> Option<DaemonKillRecord> {
     let serving = read_serving_daemon(kin_root)?;
-    if process_is_alive(serving.pid) {
+    if liveness(serving.pid) != Liveness::Dead {
         return None;
     }
     if read_daemon_death_note(kin_root).is_some_and(|note| note.pid == serving.pid) {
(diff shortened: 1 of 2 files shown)


troyjr4103:
Correction to the squash message, which is immutable on main as df375efa8: it says `save_vector_artifact` grew five `#[cfg]` attributes rather than one. The measured count is seven, flagged pre-landing by the independent review (review-kin1234, P2-1) by walking the function's brace depth from `state.rs:529` to its close at 652: `#[cfg` lines at 565, 579, 581, 605, 618, 637 and 646, six `#[cfg(feature = "embeddings")]` and one `#[cfg(not(feature = "embeddings"))]`. Every other number the review checked in the body is right. This comment exists so a reader of the commit message finds the measured value here.

PR #1255, opened 2026-08-30
Bound Git history at the derivation, and the kin-db contract that refuses the result

troyjr4103:
**This does not work end to end and must not land as a feature.** The derivation bounds correctly and the store cannot be written, because kin-db and kin-model require a repository's admitted change set to be ancestry-closed to the Git roots. That is a contract in two other repos, not a defect in this branch, and the finding is the deliverable here.

## The wall, measured in two directions and read in the third

**With the Git authority installed as it is today,** every Git commit projection must have a semantic alias. `kin-db-0.7.71/src/storage/repository.rs:3234`. Running `cargo test -p kin-core --test init_history_limit`:

```
Graph("invalid operation: Git commit projection 37c90ee2... has no
repository-scoped semantic alias")
```

Projections are built from every commit record in the object set, inside kin-model's `GitExternalAuthority::from_raw_parts` (`git_authority.rs:651`), so a whole capture makes one per Git commit while aliases exist only for admitted ones.

**Omitting the authority for a bounded admission** is the obvious workaround, and the field allows it: `git_external_authority` is `Option<GitExternalAuthority>` and kin-db skips the block when it is absent. Tested, and it refuses from the other side:

```
Graph("invalid operation: transaction alias fb3b0b5c... has no commit
projection in the resulting Git external authority")
```

So the alias set and the projection set must be equal. That cost one build and it was worth it: nothing in the first error predicts the second.

**Making them equal by bounding the capture instead** fails too, read rather than run because the reading is unambiguous. A commit's decoded dependencies are `[tree, ...parents]`, and `kin-model-0.7.22/src/git_authority.rs:1059` requires every dependency whose `requires_closure_object()` is true to be present, which is every kind except `Gitlink` (`:129`). The boundary commit's parents are absent by construction, so a bounded capture refuses with `MissingDependency`; kin-db checks the projected parents at `repository.rs:3271` as well.

Together: every admitted commit's parents must be admitted, so any set holding a non-root commit pulls its parents transitively to a root, and an ancestry-closed set is whole history. No window satisfies the contract, so nothing inside this repo closes it.

## What is on the branch and does work

The derivation bounds. `derive_semantic_git_history` is the single rule shared by plan construction and by every re-derivation that proves a plan, so a bound there cannot drift between them. The window walks HEAD's first-parent chain and strips parent edges leaving the admitted set, which makes the oldest admitted commit a root by the rule genesis already uses. `HistoryLimit::Whole` is the type's `Default` and `init_from_git` keeps its signature, so every existing caller is unbounded by construction; two integration tests cover the default separately and both pass. The flag, the manifest record, the `kin log` sentence, the budget recommendation and a four-check acceptance suite are all written.

## What it would take to finish

A cross-repo change, kin-model then kin-db then kin. kin-model must let a commit's parent dependency be absent when the authority declares that commit an admission boundary, which needs a new field and therefore a `GIT_EXTERNAL_AUTHORITY_SCHEMA_VERSION` bump; tolerating an absent parent unconditionally would accept a corrupt capture everywhere. kin-db must make the alias and projection sets a bijection over the admitted set. That size is not measured here and no estimate is offered.

## What is not claimed

No memory measurement was taken, because no bounded store could be written to measure. Whether a bounded conversion fits a smaller machine is unknown. FIR-2955 owns the daemon memory this was meant to work around and is untouched.

Refs FIR-2041
Refs FIR-2955

Filed as FIR-2967, which carries all three code sites, both measured errors and the change shape.

### DIFF
--- a/crates/kin-cli/src/commands/health.rs
+++ b/crates/kin-cli/src/commands/health.rs
@@ -309,6 +309,7 @@ pub async fn run_health_checks() -> HealthReport {
     checks.push(check_background_work().await);
     checks.push(check_embedding_model().await);
     checks.push(check_memory_floor());
+    checks.push(check_unconverted_history_budget());
     checks.push(check_commit_memory_headroom());
     checks.push(check_daemon_kill_record());
     checks.push(check_interrupted_init());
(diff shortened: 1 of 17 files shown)


troyjr4103:
**Closed by founder ruling on 2026-08-30: keep full history, and make the daemon efficient with all of it.** The bounded-import design is not being pursued. The branch is kept because the contract it measured is worth having written down, and because a bounded import may still be wanted later for its own reasons.

## What this measured, which is the reason to keep the branch

A repository's admitted change set must be ancestry-closed to the Git roots. Three checks enforce it, and no window except whole history satisfies all three.

**Every Git commit projection must have a semantic alias**, `kin-db-0.7.71/src/storage/repository.rs:3234`. Measured on this branch:

```
Graph("invalid operation: Git commit projection 1c863d965115c962428d693e38c547c62901ac8f
has no repository-scoped semantic alias")
```

**Every alias must have a projection**, `repository.rs:3142`. Measured by omitting the Git authority delta, which the field permits since `git_external_authority` is `Option<GitExternalAuthority>` and kin-db skips the block at `:3189` when absent. It refuses from the other side:

```
Graph("invalid operation: transaction alias fb3b0b5cc22e84751c1241b0e182ad3b46e16163
has no commit projection in the resulting Git external authority")
```

So the two sets must be equal. That was worth a build: nothing in the first error predicts the second, and omitting the authority is the workaround anyone tries first.

**A commit's parents must be in the closure**, `kin-model-0.7.22/src/git_authority.rs:1059`, which refuses with `MissingDependency` for any dependency whose `requires_closure_object()` is true. `Gitlink` is the one kind for which it is false (`:129`), because a submodule's history is somebody else's authority.

Together: any set holding a non-root commit pulls its parents transitively to a root, so ancestry-closed is whole history.

## What is on the branch

The derivation bounds correctly. `derive_semantic_git_history` is the single rule shared by plan construction and by every re-derivation that proves a plan, so a bound there cannot drift between them; the window strips parent edges leaving the admitted set, making the oldest admitted commit a root by the rule genesis already uses. `HistoryLimit::Whole` is the type's `Default` and `init_from_git` keeps its signature, so every existing caller is unbounded by construction.

Test grid, which is the blocker stated precisely: `init_budget` 15 passed, `history_bound` 4 passed, and `init_history_limit` 3 passed 2 failed, where every test that does not bound passes and every test that does bound fails on the same kin-db assertion for the same commit.

Also on the branch and independent of the boundary: a `Whole-history conversion` row in `kin doctor` that runs before a store exists, and three budget tests including one that pins the forecast's factoring against its old expression. Neither ships from here. The budget and doctor wording names `--history-limit`, and a remedy naming a command that refuses is FIR-2956 in a new costume.

Design record and citations: FIR-2967. Ruling: FIR-2041.

Refs FIR-2041
Refs FIR-2967

PR #1281, opened 2026-08-30
Give vcs_read_surfaces:content its own edit to admit

troyjr4103:
`vcs_read_surfaces:content` has failed on every main Acceptance run since kin#1258. The defect is in the suite, not the product.

## Mechanism, from the run's own two adjacent lines

CI run 33295456256:

```
CHECK saw_the_edit PASS  the tree moved b797a254 -> 3673bc6d, so the answer was measured
                         against the working copy
CHECK content      FAIL  a pass over an edited tracked file reported nothing changed
```

`saw_the_edit` passing is causally why `content` fails. Since #1258 `kin status` admits before it reads, so `check_saw_the_edit` does not observe the edit it makes, it **takes** it. The tree hash moving between its two status reads is exactly how that check passes. `check_content` then ran `kin admit` with nothing outstanding, and `kin admit` correctly reported a no-op.

The check written to prove read-after-admit works is the one that made the next check unreachable.

## Two hypotheses this refutes, on that run's own evidence

**Not a missing daemon.** `held_merge` PASSED and needs `kin merge` plus `kin conflicts` to run. `basis` read "as admitted 0s ago". `diff_scope` PASSED.

**Not a missing live graph.** `unadmitted` PASSED on the same run, with the no-daemon path naming its gap by name.

## The change

`check_content` writes a third module body of its own before admitting, so an edit reaches `kin admit` that no earlier status read has taken. The ordering comment above `CHECKS` described the pre-#1258 world and is corrected with it, because leaving it would invite the next reader to rebuild the same bug.

Suite-only. No product code moves.

## Why it was missed

The suite was validated end to end against a staged binary at `6438c64d1`, which is #1254's content. Read-after-admit arrived in #1258 and the suite was never re-run against a binary carrying it. Measured, not asserted: `git show 6438c64d1:crates/kin-cli/src/commands/status.rs | grep -c admit_before_reading` returns 0, and the same read against `origin/main` finds the call at line 975.

A green taken against a staged binary expires the moment the product moves past it, and nothing about the green says so.

## The class, third instance in one night

Read-after-admit has now broken three acceptance checks by removing the condition each one graded, after the two in `working_copy_freshness` under FIR-2820. A read that mutates has to be ordered like a mutation: a check that grades an unadmitted state has to create it after the last status call, and a suite sharing a fixture across steps has to be read as a sequence of mutations once one of its reads admits.

Closes FIR-2990

### DIFF
--- a/scripts/acceptance/vcs_read_surfaces_repro.py
+++ b/scripts/acceptance/vcs_read_surfaces_repro.py
@@ -144,6 +144,21 @@ def total_by(entries, key):
 '''
 
 
+# A THIRD body, for `check_content` alone. Same declarations, same count, body
+# only, like MODULE_AFTER; distinct from it so an edit reaches `kin admit` that no
+# earlier status read has already taken.
+MODULE_FOR_ADMIT = '''"""Roll ledger entries up into totals."""
+
+
+def total_by(entries, key):
+    """Sum amounts under one key, rounded to whole units."""
+    buckets = {}
+    for entry in entries:
+        buckets[entry[key]] = round(buckets.get(entry[key], 0) + entry["amount"])
+    return buckets
+'''
+
+
 def run(cmd, cwd=None, env=None, timeout=600):
     process = subprocess.Popen(
         cmd, cwd=cwd, env=env,
@@ -493,6 +508,20 @@ def check_diff_scope(suite):
 
 
 def check_content(suite):
+    """`kin admit` over a content-only edit must not report a no-op.
+
+    Makes its OWN edit, and that is the whole correction. Since kin#1258
+    `kin status` admits before it reads, so `check_saw_the_edit` above does not
+    merely observe the edit, it TAKES it: the tree hash moving is exactly how that
+    check passes. By the time this one ran, there was nothing left to admit and
+    `kin admit` correctly said so, which failed this check on every main
+    Acceptance run after kin#1258 landed.
+
+    The check that proves read-after-admit works was consuming the state the next
+    check grades. A read that mutates has to be treated as a mutation when
+    ordering an experiment around it.
+    """
+    suite.write_tracked_module(MODULE_FOR_ADMIT)
     rc, out, err = suite.kin_run(["admit"])
     if rc != 0:
         return Result("content", UNREADABLE,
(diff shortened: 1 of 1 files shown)


troyjr4103:
## Falsification, run against a binary proven before it graded anything

`kin 0.6.2 (2edc23d524a628e36a8458e2271547847bd047b0 chore/lane-vcsreads-content)`, exact sha match to this branch's head, carrying "not measured against the working copy" twice with a fabricated control at zero. That check exists because the defect this PR fixes came from grading a suite against a binary that predated the behaviour.

Four arms, one binary, so the code varies and the runner does not.

| arm | shape | got | the assertion that fired |
|---|---|---|---|
| M0 | fixed suite, daemon present (the CI shape) | **PASS** | "the pass named the content change: ... content changed, with no artifact or entity added or removed" |
| M1 | fixed suite with the new edit line DELETED | **FAIL** | "a pass over an edited tracked file reported nothing changed" |
| M2 | `origin/main`'s suite verbatim | **FAIL** | same detail, reproducing the CI red exactly |
| M3 | fixed check, no daemon can run | **UNREADABLE** | "kin admit exited 1 ... daemon startup failed" |

M0 ran all seven checks green. M1 and M2 reach the same failing assertion by two different routes, one from deleting my line and one from the shipped bytes, which is what makes the fix load-bearing rather than coincident.

### M3 was unsound the first time, and its wrong answer was a PASS

The first M3 stopped the daemon and assumed the next command would find none. It came back PASS, and I nearly filed that as a product defect. Measured instead:

```
A. after init/commit     daemons-for-this-store=1
B. daemon stop rc=0      daemons-for-this-store=0
C. status, no daemon     names-the-gap=True   daemons=0
D. admit rc=0            daemons-for-this-store=1
VERDICT: kin admit STARTED a daemon
```

**"No daemon" is not a state you reach by stopping one, because every command restarts it. It is a state you reach by making the daemon unstartable.** The corrected arm points `KIN_DAEMON_BIN` at a real executable that refuses rather than at a missing path, so it tests "no daemon can run" and not "the binary is absent". `content` then returns UNREADABLE naming the startup failure, zero daemons exist for the store throughout, and the control confirms status still names the gap.

So the no-daemon shape cannot report nothing changed, which was the bar.

### Related

FIR-2991 files the sweep this class earns: 20 of 22 suites carry the ordered-CHECKS-over-one-fixture shape, 16 of those read `kin status`, and 2 are known broken. Carrying the shape is not carrying the defect, so fourteen are unaudited rather than broken.

PR #1292, opened 2026-08-30
Give both new suite steps the visibility shape, and restore the gate's order

troyjr4103:
**Main's Acceptance has run zero suites since kin#1284 landed.** The job stops at "Falsify and enforce the workflow step visibility rule", so every suite after it reports "no report" and the gate grades nothing. #1284 broke it and #1288 added the second violation. Both are mine.

This is the check-that-cannot-fail shape arriving through the front door of two new checks being added.

## Three causes

**1. Both new suite steps omit the visibility shape**, missing the trailing:

```yaml
echo "rc=$rc" >> "$GITHUB_OUTPUT"
if [ "$rc" -ne 0 ]; then
  echo "::warning title=<suite> returned nonzero::suite process exit $rc; verdict deferred to the Acceptance verdict step"
fi
```

I copied the sibling step's body down to its `GITHUB_STEP_SUMMARY` block and stopped four lines short of where that body ends. The steps read correct beside their siblings and were not.

**2. The gate's `--report` order did not match the step order.** Steps run `blame_attribution`, `diff_content`, `vcs_read_surfaces`; the gate listed them reversed. The guard requires suite outputs and gate inputs to be one-to-one **in order**, which is what makes a report provably reach the verdict that grades it rather than merely existing somewhere.

**3. `EXPECTED_SUITES` was stale**, so the inventory the guard pins was wrong the moment either suite landed.

## Verification

Run with the two commands the step itself runs, in its order, rather than an approximation:

```
--self-test   all 2 controls clean, all 24 adversarial mutants caught
(guard)       workflow authority holds: 24 suite reports reach one always-running verdict in order
```

Both suites' own self-tests still pass, 24 and 21 cases.

**Falsified:** removing one of the two annotations returns the guard to exactly one violation, so it still catches the shape it exists for.

## The rule

Second instance tonight of one shape. The first was a compile error in `kin-daemon`'s test targets that a narrower `cargo test` could not see after the wide check had found its siblings.

**A new entry in a list that something else pins is not finished until the pin is updated and the pinning check is run.** The guard existed, it was correct, and it was never invoked locally against either change.

Closes FIR-3000

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -956,6 +956,10 @@ jobs:
             grep -E '^CHECK ' acceptance/blame_attribution.log || echo '(no CHECK line was printed)'
             echo '```'
           } >> "$GITHUB_STEP_SUMMARY"
+          echo "rc=$rc" >> "$GITHUB_OUTPUT"
+          if [ "$rc" -ne 0 ]; then
+            echo "::warning title=Blame-attribution acceptance suite returned nonzero::suite process exit $rc; verdict deferred to the Acceptance verdict step"
+          fi
 
       - name: Diff-content acceptance suite (verdict comes from the gate step)
         id: diffcontent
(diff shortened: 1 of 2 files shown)


troyjr4103:
Superseded by #1293, which carries the same fix (warning branch and rc output on both suite steps, suite table extended) and landed first as 510be53f9 while this PR's shard 1 was still running. The diagnosis in this body is the one that found the shape; credited in #1293's body and the captain report. Closing so the two do not conflict on the same two files.

PR #1296, opened 2026-08-30
Show a conflict's three sides as source, checked back against the digests the merge recorded

troyjr4103:
rc062j finding (1). A parked conflict tells you a decision is needed and nothing about what the decision is between.

## Before, measured on clean main `7682e5005c4fa7fd`

One two-line Python file edited on both branches. One entity row's three sides:

```
base    value=present   32 bytes of digest
ours    value=present   32 bytes of digest
theirs  value=present   32 bytes of digest

source text 'OURSMARKER'    anywhere in the JSON listing: NO
source text 'THEIRSMARKER'  anywhere in the JSON listing: NO
source text 'def summarize' anywhere in the JSON listing: NO
```

And `kin conflicts` carried no flag that would print one: `--json`, `--profile-out`, `--profile-summary`, `--help`.

## After

```
summarize in pkg/core.py (entity d4912d97-520c-4315-9da0-fb58d4ab9b6e)
  base -> ours:
    @@ -1,2 +1,2 @@
     def summarize(rows):
    -    return len(rows)
    +    return OURSMARKER(rows)
  base -> theirs:
    @@ -1,2 +1,2 @@
     def summarize(rows):
    -    return len(rows)
    +    return THEIRSMARKER(rows)
```

Marker counts in the rendering went from 0/0/0 to OURSMARKER 3, THEIRSMARKER 3, `def summarize` 6, with 0 sides refused.

## Why it needs re-materializing at all

`MergeSideValue::Present` carries `Hash256` of the model value, not a content address, so nothing can be looked up from it. The bytes exist only in the graph at the three changes the merge bound. So `--show` resolves the graph at base, ours and theirs, takes the entity's span inside its artifact's blob, and prints each side's departure from the common ancestor through the diff command's own `render_hunks`. Rendered against base rather than against each other, because the question a settlement answers is which side's departure to keep, and a diff of ours against theirs shows neither departure.

Every side is hashed back to the digest the record holds before it is offered, which is what makes this an account of the merge rather than of the graph as it stands now. A side whose value has moved is named in `unverified` and not printed, and so is one whose bytes cannot be read, because a body that is quietly missing reads exactly like an identity that is absent, and those are opposite facts.

Three refusals rather than approximations: a span off character boundaries or past the end of its blob (a clamped slice would render a body that is not the entity's while looking exactly like one that is); a symlink or gitlink, which carries a target rather than a body; and a relation, which has no source of its own, so rendering its endpoints would print the same bodies twice under another name.

The request field is defaulted, so a client sending `{}` keeps today's behaviour and today's cost. Reading bodies resolves three graphs and reads a blob per side, which a listing does not otherwise do, and `default` is the acceptance arm that holds that line.

## Acceptance

```
CHECK bodies FIR-2960 PASS render PASS both sides render as a diff against the
  common ancestor; verified PASS every rendered side hashed back to its recorded
  digest; default PASS the default listing carries no bodies
```

Seven checks, exit 0. The suite's self-test goes from 35 grader assertions to 44, each new grader given an input that must pass and one that must fail, plus a control that a rendering showing no source at all reads UNREADABLE rather than verified.

## Falsification

```
arm  bodies  removes
B0   PASS    nothing
B1   FAIL    the digest re-check, made unable to pass
B2   PASS    the digest re-check, DELETED
```

B1's detail shows the check gating the output path rather than sitting beside it: `render FAIL the rendering never shows OURSMARKER, THEIRSMARKER; verified UNREADABLE nothing was rendered, so nothing was verified`. The `verified` grader answered UNREADABLE rather than PASS on that silence, which is the control built for exactly this arm.

**B2 is green and that is a finding, not an oversight.** Deleting the verification outright changes nothing a healthy fixture can see, because a merge whose three sides all still hash to their recorded digests cannot distinguish "checked and passed" from "not checked". Catching a genuine divergence needs a record whose digest disagrees with the graph, and I could not construct one through the product's own surfaces without corrupting state by hand. So what is proven is that the check gates the rendering. What is NOT proven is that it fires on a real divergence in the wild.

Three arms, three distinct binaries, tree restored with zero mutation markers.

Refs FIR-2960

### DIFF
--- a/crates/kin-cli/src/commands/conflicts.rs
+++ b/crates/kin-cli/src/commands/conflicts.rs
@@ -16,7 +16,15 @@ pub const CONFLICTS_REPORT_SCHEMA: &str = "kin.conflicts.v1";
 
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
-pub struct ConflictsRequest {}
+pub struct ConflictsRequest {
+    /// Materialize each conflict subject's three sides as source.
+    ///
+    /// Defaulted, so a client that sends `{}` keeps today's behaviour and
+    /// today's cost. Reading bodies resolves the graph at three changes and
+    /// reads a blob per side, which a listing does not otherwise do.
+    #[serde(default)]
+    pub bodies: bool,
+}
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ConflictsReport {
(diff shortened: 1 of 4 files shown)


troyjr4103:
Captain: converted to draft so a green lander cannot land this body as the squash message. Two things unblock it. (1) The Before section is filed under `7682e5005c4fa7fd`, which is not a commit in this repository (`git log -1 7682e500` finds nothing; not an ancestor of origin/main): re-measure the before-arm on the PR's real base (e5bf7bf1d or the current origin/main) and name that sha and the command. (2) B2 green means the digest verification has no arm that can fail: add a fixture arm that overwrites one side's stored bytes (or records a wrong digest) so the intact code names the side in `unverified` and prints nothing for it while the verification-deleted mutant prints the corrupt body, plus a removed-blob arm. Mark ready when both are in the body with the grid.

PR #1315, opened 2026-08-30
Gate MCP daemon revival on readiness rather than liveness

troyjr4103:
`await_revived_daemon` polled `/health` and returned the daemon's URL on a 200. `/health` is liveness: the daemon's own handler refuses a repo-scoped query with "/health is liveness-only and will not lazy-load repo graphs", and it answers as soon as the HTTP server binds, before the store is open. So revival handed back a URL and the caller's next tool call failed against a daemon that was still loading.

## What was measured before anything was written

The archive stranger run `rc063a` (candidate `f2854caf6`, rc-build 33319689639) measured the window on the shipped candidate. `curl http://127.0.0.1:44961/health` answered **200 in 7 ms** while the same daemon's `/mcp/tools/call` refused, `kin daemon status` printed `requests running pid 3955 port 44961 861 entities`, and the retry failed with "daemon was revived at http://127.0.0.1:44961 but the retry still failed". A third retry seven minutes later succeeded with no intervention. Evidence at `/Users/Shared/kin-stranger/rc063a/brown/work/EXPERIENCE.md:308-313`.

The source reading that decided the fix is in `.kin-coord/reports/rc063classify-20260830.md` under "FIR-3030, building lane". `/readiness` already exists on the same router and returns 200 only once `is_initialized` is set, after a snapshot load or a completed reconciliation cycle, and an empty but initialized workspace is ready.

## A second defect this surfaced, and it is a check that could not fire

`still_starting_message` produced `daemon on port N is still starting`, which does not contain `DAEMON_STILL_STARTING` ("daemon is still starting"), the constant `is_still_starting_error` matches. That classifier is the only reader of the constant and it gates the retry branch that tells an agent not to restart `kin mcp start`. The branch could never fire for the health half of a revival while firing correctly for the port half, whose text in `kin-daemon-spawn` carries the constant verbatim. One revival had two halves disagreeing about whether their own outcome was retryable.

The existing test asserted `contains("still starting")`, a substring that held while the classifier returned false. That is why the gap survived, and the new test asserts through `is_still_starting_error` with an exited-daemon control that must not classify.

## The timeout now separates two states that are not the same news

A daemon answering `/readiness` with 503 is serving and the store open is what remains. One answering nothing has published a port and has not begun serving. Both are retryable, and reporting them identically would leave a mutation that swaps the arms invisible.

## Falsification

The endpoint choice was an inline format string, so no test could reach it. It is now a named constant behind a three-state probe, and the test serves one socket answering `/health` 200 and `/readiness` 503, which is the state a real daemon is in mid-open. The test asserts the liveness route really does answer 200 first, so the readiness assertion beside it means something.

Grid run on the committed tree, each arm naming the assertion that answered:

| arm | mutation | result | assertion that fired |
|---|---|---|---|
| M1 | route back to `/health` | RED | `daemon_delegate.rs:2852` the probe reported ready while readiness said 503 (`left: Ready`) |
| M2 | message drops the constant | RED | `daemon_delegate.rs:2761` the report is not classified as still-starting |
| M3 | collapse the two observations into one sentence | RED | `daemon_delegate.rs:2898` the silent report carries the serving clause |

Three distinct lines, so no arm is red for another arm's reason. Restore verified by printing the value restored to (`const REVIVAL_READINESS_ROUTE: &str = "/readiness"`) with zero files differing from HEAD. Baseline M0 green, and a fabricated test name runs 0 tests so the filters can answer zero.

## Gates, on the rebased tree `defedf7a8`

`cargo fmt --all -- --check` rc=0. `cargo test -p kin-mcp` **804 passed, 0 failed**. Run through `heavy-slot.sh` with a per-lane target.

`cargo clippy -p kin-mcp --all-targets -- -D warnings` fails, and it fails inside `kin-git` and `kin-parser` before reaching this crate. Both are byte-identical to `origin/main` on this branch (`git diff --name-only origin/main HEAD -- crates/kin-git` returns 0 files, same for `kin-parser`, against `crates/kin-mcp` returning 1 as the control), so this branch does not touch the code clippy refuses. Reported rather than worked around.

## What this does not close

The ticket's acceptance also asks for a real-store reproduction: a daemon whose open outlasts the probe interval, with a dead-daemon fast-fail control. That needs the daemon lock, which is held for FIR-3010; this lane reserved rather than polled and the runtime arm is still owed. The unit grid proves the endpoint choice and the classifier gap; it does not stand in for the store-scale arm.

FIR-2842's half is left alone deliberately. Its hardcoded 15 s is gone at source, `still_starting_message` derives its number from the patience actually waited, and the candidate's own message names the real 300 s and the real environment variable. A comment recording that is on that ticket.

Closes FIR-3030

### DIFF
--- a/crates/kin-mcp/src/daemon_delegate.rs
+++ b/crates/kin-mcp/src/daemon_delegate.rs
@@ -1621,7 +1621,24 @@ async fn await_revived_daemon(
         .await
         .map_err(|e| format!("MCP revival: {e}"))?;
 
-    // Poll /health until the daemon is ready, under the same patience.
+    // Poll /readiness until the daemon can actually serve, under the same
+    // patience.
+    //
+    // This polled `/health`, and `/health` is liveness. The daemon says so in
+    // its own words: a repo-scoped query to it is refused with "/health is
+    // liveness-only and will not lazy-load repo graphs". It answers 200 as soon
+    // as the HTTP server binds, which is before the store is open, so a revival
+    // that returned on it handed the caller a URL and the caller's very next
+    // tool call failed against a daemon that was still loading. A stranger run
+    // measured exactly that: `/health` answered 200 in 7 ms on the revived
+    // daemon while `/mcp/tools/call` refused, and the retry failed on a daemon
+    // that answered the same call correctly minutes later.
+    //
+    // `/readiness` is the endpoint that answers the question this loop is
+    // asking. It is registered on this same router beside `/health` and returns
+    // 200 only once `is_initialized` is set, which happens after a snapshot load
+    // or a completed reconciliation cycle. An empty but initialized workspace is
+    // ready, so this does not wait on content the repository does not have.
     let new_base = format!("http://127.0.0.1:{port}");
     let probe = reqwest::Client::builder()
         .connect_timeout(Duration::from_millis(300))
@@ -1629,9 +1646,13 @@ async fn await_revived_daemon(
         .build()
         .map_err(|e| format!("MCP revival: build probe client: {e}"))?;
     let deadline = tokio::time::Instant::now() + patience;
+    // Whether the daemon ever answered the readiness route at all, which
+    // separates "bound and openly not ready" from "not answering yet". Both are
+    // retryable and they are not the same news, so the timeout says which.
+    let mut answered_readiness = false;
     loop {
-        if let Ok(resp) = probe.get(format!("{new_base}/health")).send().await {
-            if resp.status().is_success() {
+        match probe.get(format!("{new_base}/readiness")).send().await {
+            Ok(resp) if resp.status().is_success() => {
                 // A daemon the supervisor does not know about is unroutable to
                 // every other client, so registration is part of starting one.
                 if let Err(error) =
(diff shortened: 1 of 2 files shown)


troyjr4103:
Third commit adds the acceptance suite the ticket asks for, `scripts/acceptance/mcp_revival_readiness_repro.py`, covering the half that needs no fleet lock.

It grades which endpoint carries the signal, using a stub answering `/health` 200 and `/readiness` 503 from one socket. Four checks, all falsified before this landed:

| arm | mutation | result |
|---|---|---|
| A1 | ready fixture answers 503 too | FAIL `readiness_route_carries_the_signal`, "carries no more information than /health" |
| A2 | prober asks `/health` instead | FAIL `prober_asked_readiness`, "paths seen: ['/health']" |
| A3 | stub's `/health` answers 503 | UNREADABLE `fixture_control`, so the suite refuses to grade rather than passing on a fixture that proves nothing |
| control | unmutated | 4 PASS, 0 FAIL, 0 UNREADABLE |

The `prober_asked_readiness` check reads the paths the socket actually received rather than the result, because a prober that asked `/health` and guessed right would pass a result-only check.

The closed-port control binds a port, reads the number and releases it rather than hardcoding one, and reports UNREADABLE if anything answers there.

Placed beside `first_query_readiness_repro.py` rather than inside it: that suite grades disclosure on the first query after `kin init`, this one grades the revival after a mid-session daemon death and which route it asks for.

Still owed, unchanged from the body: the real-store arm, a daemon whose open outlasts the probe interval with a dead-daemon fast-fail control. That needs the daemon lock, held for FIR-3010. Left as a comment rather than a body edit, since a body edit re-triggers PR text hygiene and flips mergeStateStatus to BLOCKED while it re-runs.

PR #1321, opened 2026-08-30
Take the CI reading for the git-index fix

troyjr4103:
The acceptance reading for kin#1319, squash `6b3a5a453250d4d72253e907cbadbfb3c9e04ed8`.

That change set `GIT_OPTIONAL_LOCKS=0` on the git chokepoint in `kin-buildinfo`'s build script, so the script stops rewriting the git index it declares as one of its own `rerun-if-changed` inputs. The claim was that this ends a wasted recompile in every fast-gate shard. The claim is testable only on CI, and this is the test.

## What to read

This PR's `Fast gate test shard` log has two ``Finished `test` profile`` lines, one for `cargo nextest list` and one for `cargo nextest run`. Split the `Compiling` lines on them.

| | before, run 33326549732 | this PR |
| -- | -- | -- |
| phase 1, `nextest list` | 24 crates, 2m 43s | |
| phase 2, `nextest run` | **4 crates**, 1m 49s | **must be 0** |
| the four | kin-buildinfo, kin-cli, kin-daemon, kin-integration-tests | |

## Why a comment-only change

A comment in the file under test makes the diff select the same crates a real change would, so the shard does the same work, while altering no behaviour. Zero non-comment lines change, asserted mechanically before the commit.

An empty commit would not do: with no changed file the scope selector narrows and the shards would compile little or nothing, which reads as a pass for the wrong reason.

## The falsification, if this reads zero

Remove the switch on a third commit of the same shape. Phase 2 must return to four. A zero that cannot go back to four is not evidence.

I am not claiming the result in this description. The numbers go in a comment once the run concludes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1241,7 +1241,13 @@ jobs:
               raise SystemExit(1)
           LISTING
 
+      # DIAGNOSTIC, on a branch that never lands. FIR-3045's first fix was
+      # refuted by its own acceptance: phase 2 still recompiled four crates.
+      # This names the input cargo considers dirty instead of guessing a second
+      # time. Remove before this branch could ever merge; it does not.
       - name: Test
+        env:
+          CARGO_LOG: cargo::core::compiler::fingerprint=info
         run: |
           set -euo pipefail
           readarray -t SCOPE < "$RUNNER_TEMP/scope.args"
--- a/crates/kin-buildinfo/build.rs
+++ b/crates/kin-buildinfo/build.rs
@@ -34,15 +34,40 @@ fn main() {
         .or_else(|| find_workspace_root(&manifest_dir))
         .unwrap_or_else(|| manifest_dir.clone());
 
+    // `git rev-parse --git-path` answers RELATIVE to the repository in a normal
+    // checkout and ABSOLUTE in a linked worktree, and cargo resolves a relative
+    // `rerun-if-changed` against the PACKAGE manifest directory rather than the
+    // repository root. So a bare `.git/HEAD` here registers
+    // `crates/kin-buildinfo/.git/HEAD`, which does not exist, and a
+    // rerun-if-changed on a missing file makes cargo treat this unit as dirty on
+    // every invocation.
+    //
+    // Measured on CI before this was joined: cargo said
+    //   stale: missing ".../crates/kin-buildinfo/.git/HEAD"
+    //   dirty: FsStatusOutdated(StaleItem(MissingFile { path: ... }))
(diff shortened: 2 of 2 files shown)


troyjr4103:
Closing unmerged. This PR existed to take measurements, not to change the tree, and it took five.

| run | commit | what it carried | phase 1 | phase 2 |
| -- | -- | -- | -- | -- |
| 33326549732 | main, before anything | no fix | 24 crates, 2m 43s | **4 crates**, 1m 49s |
| 33335063224 | `6b3a5a453` | kin#1319's `GIT_OPTIONAL_LOCKS=0`, on main | 24 crates, 2m 15s | **4 crates**, 1m 29s |
| 33335132217 | `523d1c31e` | a comment-only no-op commit | 49 crates, 1m 58s | **4 crates**, 1m 34s |
| 33335619683 | `c896767a6` | the `CARGO_LOG` fingerprint probe | 24 crates, 2m 15s | **4 crates**, 1m 39s |
| 33336196295 | `8d39541db` | the same probe **plus the absolute-path fix** | 24 crates, 1m 45s | **0 crates**, 0.26 s |

## What the five bought

Runs one to three **refuted** kin#1319. `GIT_OPTIONAL_LOCKS=0` stops the build script rewriting the git index, which was measured and is real, but the index was never the trigger and phase 2 stayed at four.

Run four **named the input**, in cargo's own words:

```
stale: missing "/home/runner/work/kin/kin/crates/kin-buildinfo/.git/HEAD"
dirty: FsStatusOutdated(StaleItem(MissingFile { path: ... }))
```

`git rev-parse --git-path` answers relatively in a normal checkout and absolutely in a linked worktree, and cargo resolves a relative `rerun-if-changed` against the **package** directory rather than the repository root. So the script registered `crates/kin-buildinfo/.git/HEAD`, which does not exist, and a rerun-if-changed on a missing file makes the unit dirty on every invocation. The other three crates followed as `StaleDepFingerprint` on that build script's unit.

It is invisible in a linked worktree, where the path resolves. Every lane checkout in this fleet is one, which is why no local investigation could have found it.

Run five is the **acceptance**: four crates to zero, 1m 39s to 0.26 s, against run four as a single-variable control on the same branch with the same probe.

## Why this closes rather than lands

Its commits are a comment, a diagnostic and a fix. The comment and the diagnostic must not reach main, and the fix lands on its own branch without them. The evidence belongs in these runs and in FIR-3045, not in the tree, and landing four comment lines in the exact file whose `rerun-if-changed` sensitivity is the subject would churn kin-buildinfo's dependents once more for zero content.

PR #1333, opened 2026-08-31
Materialize graph sections before first reopen

troyjr4103:
## Outcome

Kin now materializes the workspace-base graph section at the tail of `kin init`, before the first daemon reopen. Developers also get an explicit offline `kin graph materialize --json` repair verb with a stable, typed receipt.

## What changed

- Pin the published registry dependency `kin-db = 0.7.81`.
- Share the exact per-repository OS runtime authority between daemon startup and offline maintenance without changing process-identity ownership.
- Run graph-section materialization after verified init publication and before enrichment or daemon autostart.
- Keep materialization out of daemon HTTP and ordinary publish paths.
- Return canonical 64-character lowercase semantic-change IDs in the v1 JSON receipt.
- Preserve optional `source` wire compatibility for older daemon responses.
- Add the FIR-2955 acceptance check and focused init, contention, idempotence, JSON, and compatibility coverage.
- Move only the lifecycle filesystem exemption that the shared runtime authority extraction moved.

## Verification

- `bin/kin-parity`: PASS-WITH-ALLOWANCES. Magic 26 of 26 passed, including FIR-2955. The only allowances are the existing FIR-2464, FIR-2501, and FIR-2580 entries.
- `bin/kin-precheck kin`: 16 of 16 passed.
- Shared runtime authority: 2 of 2 passed.
- Daemon lifecycle: 78 of 78 passed.
- Graph materialization mapping and JSON: 3 of 3 passed.
- Init then explicit reopen: 1 of 1 passed.
- Older daemon response compatibility: 1 of 1 passed.
- Publication lease registry-pin regression: 1 of 1 passed.
- Zero file-search verifier and poisoned-tree falsifier: passed.

Falsification removed each guarded behavior independently. The init test failed when init skipped materialization, the contention test failed when authority opened before the shared lock, JSON tests failed when `resolved_at` reverted to a byte array, and the old-daemon response test failed when optional-field omission was removed.

## Boundaries

This change does not add a daemon materialization route, modify ordinary publication, or claim to resolve the separate revision-map divergence observed in kin-db measurement. Fold-produced and separately re-derived revision timelines remain a distinct correctness investigation.

Supersedes #1261.

Closes FIR-2955

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1909,7 +1909,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "19753d40da53d0ec41604750eeb969097a90fb2d7f7992730d904541c04e2c19"
 dependencies = [
  "bstr",
- "hashbrown 0.17.1",
+ "hashbrown 0.16.1",
 ]
 
 [[package]]
@@ -3289,9 +3289,9 @@ dependencies = [
 
 [[package]]
 name = "kin-db"
-version = "0.7.77"
+version = "0.7.81"
 source = "sparse+https://kinlab.ai/registry/cargo/"
-checksum = "bb78d778724f05b4409c43c48c7a47abdb653c048435fe416a974293d3e98e06"
+checksum = "81af647f5d6309ff18d9818e33eebc6e7cfcdfad87793df561a1e6d584086780"
 dependencies = [
  "bincode",
  "bstr",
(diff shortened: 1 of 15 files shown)


troyjr4103:
Superseded by #1334. The replacement has an identical reviewed tree but recreates the dependency commit without the automation branch's reserved marker, as required by the registry admission policy.

PR #1354, opened 2026-09-02
Read C types through extern C blocks and name typedefs by their identifier

troyjr4103:
On the demo's hiredis store an agent named `redisReader` correctly and the verifier could not resolve it, so the demo blamed Kin for a name that is in the source at `read.h:98`. It was not just that type. The C adapter never entered a `linkage_specification`, which is where tree-sitter puts the body of the standard C header idiom, because `#ifdef __cplusplus / extern "C" { / #endif` opens the brace inside one conditional and closes it inside another. That hid the entire public API of eight hiredis files.

## Measured first

Read-only against the demo's store on 2026-09-02, `kin graph export --json --limit 5000`: 730 entities, 1659 relations, of which Function 532, Macro 148, Class 28, Method 12, TypeAlias 8, EnumDef 2. Enumerating every `typedef struct`, named struct, union, enum and typedef alias in the tracked sources gives 78 declarations, and 29 of them were in the graph.

`read.h` contributed 26 entities and every one was a Macro: no types, no prototypes. The three Class nodes attributed to `hiredis.h` were all bodyless forward declarations at `:41`, `:107` and `:108`, not definitions, so the graph claimed hiredis defines `timeval` while the real 40-line `redisContext` at `:258` was absent. Two of the eight TypeAlias nodes carried a raw declarator as their name, such as `(redisPushFn)(void *, void *)`, which no query can reach.

## Change

`extract_c_node` now walks into `linkage_specification`, `declaration_list` and `ERROR`. The last one matters for `adapters/libuv.h`, where a `#if` arm opens a function body and the `#else` arm opens a second signature, so the braces never balance and the whole file becomes one ERROR node whose children are well formed.

Four narrower defects the walk exposes are fixed with it. A `typedef union` had no branch and was filed as a TypeAlias; it is a Class now, mirroring struct and bare-union handling. The typedef name came from the raw text of the declarator field, so the chain is followed to the name at the bottom instead, reading the leaf's text rather than trusting its kind, because the grammar spells `size_t` as a `primitive_type`. Only the first declarator field was read, so `typedef struct X {} A, *Ap;` lost `Ap`; every declarator is read now, and a name introduced through a pointer or array declarator becomes a TypeAlias rather than another name for the record. The struct, union or enum tag is recorded alongside the alias when the two differ, because C reaches the type either way.

A specifier with no body is a forward declaration or a bare type reference, so it no longer counts as a definition on sight. It is held back and admitted only when nothing in the file defines that name. That keeps an opaque handle in the graph, stops a forward declaration from shadowing the definition further down the same header, and stops `struct timeval tv;` from claiming the file defines `timeval`. A declaration that merely names a type no longer ends the read either, so `struct redisReply *redisCommand(...);` records its prototype.

## Proof

Eleven new tests in the adapter's test module, on a hiredis-shaped header carrying a tagged typedef struct, an anonymous typedef struct, a plain named struct, an enum, a union and two function-pointer typedefs. Each was falsified by breaking the production behavior once: ten mutations, each turning exactly the tests written for it red, then restored to 28 of 28 green. The ERROR test asserts the fixture actually produces a single top-level ERROR node, because the first fixture I wrote passed without the ERROR arm and proved nothing.

A fresh `git clone` of hiredis at the demo's own commit `29ea279`, indexed with `kin init --no-enrich` from this branch's build into a scratch KIN_HOME:

| Kind | before | after |
|---|---|---|
| Function | 532 | 814 |
| Macro | 148 | 278 |
| Class | 28 | 59 |
| TypeAlias | 8 | 24 |
| EnumDef | 2 | 10 |
| Constant | 0 | 9 |
| Method | 12 | 12 |
| **total** | **730** | **1206** |

All 78 declarations now resolve. `read.h` goes 26 to 36, `hiredis.h` 27 to 86, `async.h` 1 to 31, `sds.h` 5 to 66, `ffc.h` 1 to 277, `adapters/libuv.h` 0 to 13. `dict.h` and `net.h` are unchanged at 29 and 13, which is the control: files with no linkage specification do not move.

`kin locate redisReader` now returns `read.h` first with an `entity_resolve` signal and a named match, and the entity's span is `read.h` lines 98 to 118, the definition itself. Before, `read.h` ranked seventh on an embedding signal with no entity resolution.

`bin/kin-precheck kin`: 16 gates enumerated from ci.yml's check job, 16 ran, 16 passed, 0 failed, on tree `4cca04073` with this change applied. `cargo test -p kin-parser` is green across all 13 targets, and the round-trip fuzz suite in `kin-integration-tests` is green.

No UsesType edges move: `ENRICHABLE_LANGUAGES` in `crates/kin-core/src/reference_coverage.rs:53` covers Rust, Python, TypeScript and JavaScript only, and `lsp_adapter_for` in `crates/kin-daemon/src/daemon.rs:415` returns None for C, so no C entity has ever carried one.

Related: FIR-3085

### DIFF
--- a/crates/kin-parser/src/languages/c_lang.rs
+++ b/crates/kin-parser/src/languages/c_lang.rs
@@ -40,13 +40,34 @@ impl LanguageAdapter for CAdapter {
         };
 
         let mut entities = Vec::new();
+        let mut forward_decls = Vec::new();
         let mut relations = Vec::new();
         let mut imports = Vec::new();
         let root = tree.root_node();
         let mut cursor = root.walk();
 
         for child in root.children(&mut cursor) {
-            extract_c_node(&child, source, file_id, &mut entities, &mut relations);
+            extract_c_node(
+                &child,
+                source,
+                file_id,
+                &mut entities,
+                &mut forward_decls,
+                &mut relations,
+            );
+        }
+
+        // A forward declaration names a type whose members are spelled out somewhere
+        // else. Keep it only when this file never spells them out, so an opaque handle
+        // still reaches the graph while `struct redisContext;` does not shadow the
+        // definition that follows it further down the same header.
+        for candidate in forward_decls {
+            let defined_here = entities
+                .iter()
+                .any(|entity| entity.kind == candidate.kind && entity.name == candidate.name);
+            if !defined_here {
+                entities.push(candidate);
+            }
         }
 
         // Recursively extract all includes and ALL_CAPS macro usages across the whole file
@@ -90,6 +111,7 @@ fn extract_c_node(
     source: &[u8],
     file_id: &FilePathId,
     entities: &mut Vec<ExtractedEntity>,
+    forward_decls: &mut Vec<ExtractedEntity>,
     relations: &mut Vec<ExtractedRelation>,
 ) {
     match node.kind() {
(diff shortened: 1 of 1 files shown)


troyjr4103:
One tradeoff a reviewer should see before this lands, measured on the same fresh clone.

Making header bodies visible also makes their prototypes visible, and the name-resolution linker treats a prototype as a call target equal to the definition. A prototype has no body, so every `Calls` edge touching one is an incoming call that landed on a declaration.

| | demo store | this branch |
|---|---|---|
| names declared in a `.h` and defined in a `.c` | 38 | 152 |
| incoming call edges landing on a prototype | 36 | 435 |
| names where no caller reaches the `.c` definition | 6 | 40 |

The defect pre-dates this change: `net.h` has no `extern "C"` block, so its prototypes were always visible and already captured six definitions, `redisKeepAlive` and `redisContextConnectUnix` among them. This change scales it to 40, `redisCommand` and `redisReaderGetReply` included. Ask the new graph who calls the `read.c` implementation of `redisReaderGetReply` and the answer is nobody; both callers land on the `read.h` prototype.

Filed as FIR-3087 with the full numbers and a fix shape: emit a declaration-to-definition edge from the adapter, which can already tell the two apart, and teach the linker to prefer a definition when a name resolves to both. That is a kin-index change with its own resolution-tier semantics, so it is not a rider on this parser PR.

Worth knowing while reading it: this change is a large net gain in coverage, 730 entities to 1206 with every type declaration resolving, and a bounded regression in call-edge precision on 40 names. If the demo would rather not see the precision cost at all, FIR-3087 should land first.

While measuring this I also confirmed that cross-file C call edges do not depend on a language server: the `--no-enrich` store carries 764 cross-file `Calls` and 550 cross-file `UsesMacro` edges, produced by this adapter's bare `dst_name` plus `resolve_one_file` in `crates/kin-index/src/linker.rs:806`. C is not in `ENRICHABLE_LANGUAGES`, so it never had an LSP sweep to lose.

Separately, `kin graph export --json` sorts each link's endpoints by entity id, `source < target` in 1658 of 1659 links on one store and 2163 of 2164 on the other, so direction cannot be read from an export. Every count here is undirected.

PR #1372, opened 2026-09-02
Bump kin-db to 0.7.86

troyjr4103:
Takes kin-db 0.7.86 (published 12:14Z), which carries 0.7.85's FIR-3096 salvage fix and 0.7.86's mapped snapshot read for the kernel memory work, beside the kin-model 0.7.23 pin already on main.

This is the registry receiver's own pin refresh from #1360, carried on a lane branch so it lands with a fresh check set (the wave PR's fast-gate runs were cancelled mid-flight) and so the receiver's reserved marker stays off main's history. Cargo.toml and Cargo.lock only; no new environment names.

Related: FIR-3096

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -141,7 +141,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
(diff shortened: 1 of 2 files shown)


troyjr4103:
Superseded: kin-db 0.7.87 is on the index and the registry receiver's own wave PR (#1360) now lands itself through kin-registry-wave-land.yml (kin #1371), so this hand-carried 0.7.86 pin would only race it.

PR #1417, opened 2026-09-03
Falsify the Windows cross-check against the kin#1405 shape

troyjr4103:
Throwaway, to falsify the new job on kin#1416. Close after reading.

This tree is main plus kin#1416's `windows-cross-check`, with `crates/kin-core/src/file_limit.rs` restored byte for byte from `ddc3e0621`, the sha kin#1405 landed. `target_soft_limit` sits at module scope while its only caller sits inside `#[cfg(unix)] mod imp`, so it is dead code on Windows and the workspace compiles with `-D warnings`. That pull request read green while all four Windows jobs failed to compile on main.

What has to happen for the new job to be evidence: `Windows cross-check` red here, naming

```
error: function `target_soft_limit` is never used
```

and green on kin#1416, which carries the identical job against a tree without the defect. A guard nobody has watched fail is not a guard.

No local gate ran on either branch. A benchmark round holds the compute quiet lock, so hosted CI is doing the grading.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1291,15 +1291,107 @@ jobs:
       - name: Report quarantined flakes
         run: python3 scripts/check-quarantine.py --report-junit target/nextest/default/junit.xml
 
+  # Windows compilation, graded on the pull request that breaks it.
+  #
+  # kin's four native Windows jobs all carry
+  # `if: ${{ github.event_name != 'pull_request' }}` (lines 201, 454, 533 and
+  # 638), so a pull request cannot see a Windows break at all: it lands, and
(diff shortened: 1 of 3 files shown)


troyjr4103:
Falsified, and the guard is evidence. `Windows cross-check` went red here in 1m49s with the exact error the native Windows jobs reported on main:

```
error: function `target_soft_limit` is never used
  --> crates/kin-core/src/file_limit.rs:57:4
   = note: `-D dead-code` implied by `-D warnings`
error: could not compile `kin-core` (lib) due to 1 previous error
Process completed with exit code 101.
```

The same job is green on kin#1416 in 2m37s from a cold cache. One file apart, opposite verdicts, so a green there means something. Closing; this branch existed only to watch the check fail.

PR #1420, opened 2026-09-03
Carry the semantic-primacy rescue into the entity-granular fuser

troyjr4103:
`locate`'s rescue for an embedding-only answer is dark on the profile users run. This carries it into the path they actually take, bounded so it cannot outrank an anchored file, and it ships OFF until an A/B earns it a default.

## Step zero, because a term removed on evidence must not be restored blind

Nothing removed it on evidence. `git log -S entity_fusion_default` returns three commits and `git log -S reciprocal_rank_fusion_entities` returns one, and that is the whole record.

The flip plan the call site cites is real and recoverable at `git show ba08f07e3:crates/kin-cli/docs/locate-entity-fusion-flip-plan.md`. It says "Status: **default OFF**, experiment wired, awaiting post-freeze A/B. Flipping the default is **out of scope** for the implementing lane", and it sets a four-clause gate: top-entity MRR beats run-to-run variance, symbol and line F1 hold or improve, file F1 does not regress, no passing gold drops to zero. `7ecefffaf` deleted it on 2026-06-11 in a three-file docs sweep that argues nothing about fusion.

kin#160 turned entity fusion on for real users on 2026-07-02 under a section headed "Validation (prepared, not run)". kin#841 widened it to accuracy-v2 on 2026-08-15 citing a 20-of-23 probe result, and the report it cites says in its own section 3 that the number was measured with entity fusion off.

The flip plan never mentions the semantic-primacy term (`grep -c -i primacy` returns 0 against a control that returns 34). So dropping it was not a decision anyone made. Restoring it corrects an omission. It still has to earn its default below.

## The defect

`accuracy-v2` sets `entity_fusion_default`, so the selector takes `entity_granular_fused_files`, which fuses through `reciprocal_rank_fusion_entities`. That combine is `1/(60 + rank + 1) + raw_norm * 0.05 + (signals - 1) * 0.02`. A key two arms found carries two rank terms plus the cross-signal bonus; a key one arm found carries one. About 0.034 apart at a middling rank, and semantic strength cannot buy it back. Measured on the demo's VS Code store, a file only the vector arm found scores about half a file both arms found: median 0.00552 over 119 files against 0.01016 over 335.

## The change

`apply_entity_semantic_primacy` adds `weight / (k + rank + 1)` to any key the embedding list ranked, the same shape as `reciprocal_rank_fusion_weighted`'s term, reading that function's own `KIN_LOCATE_SEMANTIC_PRIMACY_K` so one sweep moves both fusers. Index positions survive the re-keying: `entity_granular_fused_files` enumerates `ranked_lists` in order, so index 9 is the embedding list in both.

The graph-neighbourhood tiebreaker stays out, and the fuser's doc comment now says why for that term alone. It counts appearances across indices {1, 2, 4, 6}, which stay path-keyed here while 7 and 9 are entity-keyed, so the count would span two key spaces. One term is portable, one is not, and the old comment condemned both.

**The bound.** A key may consume at most `KIN_LOCATE_ENTITY_SEMANTIC_PRIMACY_HEADROOM` (0.9) of its gap to the unlifted top score. A capped key lands at `top * h + score * (1 - h)`, strictly below `top` for every `h < 1` and every `score < top`, and a key already at the top has no gap and takes no lift, so its score stays bit-identical. Fusion rank one cannot move, and since the projection gives each file its best entity's score, the top file out of fusion cannot move either.

Capping the gap rather than clamping to a fixed ceiling is what keeps the cap order-preserving. `top * h + score * (1 - h)` is strictly increasing in `score`, so two capped keys cannot collapse onto one value and be reordered by the key tiebreak.

**The weight is not monotone, and the A/B is designed around that.** Once a key's lift exceeds its headroom the cap binds, and capped keys keep their unlifted order, so a large enough weight caps the whole embedding set and returns it to the order it already had. Simulated over this exact combine on a fourteen-key fixture, the rank of an embedding-only key at embedding rank 1:

| weight | h=0.5 | h=0.9 | h=0.99 |
| -- | -- | -- | -- |
| 0.10 | 12 | 12 | 12 |
| 0.25 | 11 | 11 | 11 |
| 0.50 | 8 | 7 | 7 |
| 0.75 | 12 | **5** | **5** |
| 1.00 | 12 | 7 | 6 |
| 2.00 | 12 | 12 | 12 |
| 5.00 | 12 | 12 | 12 |
| 20.0 | 12 | 12 | 12 |

Unlifted rank is 12. Rank one holds in all twenty-four cells. The sweep reads a peak; it does not push the weight up expecting more effect.

## Pre-registered ship rule, written before any measurement

This is kin#1389's rule unchanged, as kin#1412 was graded against, plus this ticket's own clause. Every clause, or nothing ships.

1. rank one unchanged on all seven prose questions;
2. all twenty-two identifier lookups byte-identical on the entity ranking;
3. `CodeEditorWidget` holds entity rank 5 on VS Code, inside the five focals;
4. React's scheduling question keeps 3 of 4;
5. hiredis P2 keeps `redisReaderGetReply` on the ranking;
6. no single-letter generic in ripgrep's rg-walk focals;
7. **new:** `cursorTypeOperations.ts` enters the fifty-file list on VS Code.

If no arm passes every clause, this PR closes with the numbers and the ticket says why. That outcome is fine and it has already happened once on this seam.

Clause 1 is true by construction at the fusion boundary and only there. The lexical parity floor spends its budget ordered by parity and then by pre-floor score, and this change moves pre-floor scores in the tail, so which twelve files take the floor can change. Only the A/B settles clauses 1 through 7 in the answer.

## Honest expectation, recorded before measuring

The arithmetic says clause 7 is the one at risk. The lift differentiates by embedding rank, so it barely separates two keys that are both in the embedding list at nearby ranks; it separates strongly against keys the embedding arm never ranked. `cursorTypeOperations.ts` sits at embedding rank 124 of 541 and needs to pass roughly 205 files, most of which the embedding arm also ranked. The post-fusion pipeline is multiplicative, so a fusion-score gain does not map linearly onto the pre-floor rank, which is why this is measured rather than predicted.

## Gates

The box is under a compute quiet for an overnight benchmark, so nothing was built or tested locally. `rustfmt --edition 2021` ran and reported clean, which is a parse, not a build. `python3 scripts/gen_env_registry.py` regenerated the knob registry and `docs/env-vars.md` was updated to match by hand, so `env_doc_matches_registry` is the check to read. Hosted CI is the gate of record for everything else here.

Two things wait for the lift and this stays a draft until both are done:

- the falsification of each new test, by breaking the behaviour it guards and reading red. Three mutants: default the weight to a lifting value and `entity_fusion_semantic_primacy_ships_dark` must go red; drop the `.min(allowed)` cap and `entity_fusion_semantic_primacy_cannot_take_rank_one` must go red on rank one; skip the lift for keys the embedding arm ranked deeper than rank 0 and `entity_fusion_semantic_primacy_rescues_an_embedding_only_key` must go red.
- the seven-question A/B against the pre-registered rule.

FIR-3123. Splits out of FIR-3118.

## A second commit, and why

`Check & Test` is a no-op placeholder on a pull request here; its whole log is 2,456 bytes and its only step prints that it runs in the merge queue and on every main commit. The fast gate does the real grading and it runs `cargo nextest`, which gives every test its own process. `kin-core/src/test_env.rs` says in its own module doc that a shared-process `cargo test` is the authority when the two runners disagree, because nextest makes env-leak defects structurally invisible. So a new test that reads process-global env state another test set is graded on a PR only by the runner that cannot see it, and by the runner that can only after it lands.

`reciprocal_rank_fusion_entities` reads four knobs. The guard pinned the three this term introduces or reuses and left `KIN_LOCATE_RRF_RAW_WEIGHT` open. No test sets it today, which is exactly the state that makes the next one cheap to get wrong. The second commit pins all four.

## What CI actually proved

46 check runs on the first commit and 45 on the second, `total_count` equal to the array length both times so nothing paged, zero failures. Green alone would not be evidence that a new test ran, so the shard logs were read by test name: `rescues_an_embedding_only_key` PASS on shard 1, `ships_dark` PASS on shard 2, `cannot_take_rank_one` and `the_shipped_profile_fuses_at_entity_granularity` PASS on shard 3. Each asserts its baseline green before it measures an arm.

### DIFF
--- a/crates/kin-cli/src/commands/locate.rs
+++ b/crates/kin-cli/src/commands/locate.rs
@@ -12895,12 +12895,11 @@ fn entity_seed_keyed(
 /// PATH-keyed here while 7 and 9 are ENTITY-keyed, so that count would span two
 /// key spaces and never coincide.
 ///
-/// It does carry the path fuser's semantic-primacy term, off by default and
-/// applied by [`apply_entity_semantic_primacy`]. Index positions survive the
-/// re-keying: `entity_granular_fused_files` builds `keyed_lists` by enumerating
-/// `ranked_lists` in order, so index 9 is the embedding list in both, keyed by
-/// entity id. Without that term a key only the vector arm found carries a bare
-/// rank contribution, which is the exact case the term was written for.
+/// It also omits the path fuser's semantic-primacy term, which
+/// [`entity_granular_fused_files`] applies to this function's OUTPUT instead. The
+/// term has to be keyed by the file-level embedding ranking, and that list is not
+/// available here: index 9 of `keyed_lists` holds the embedding SEED entities,
+/// which are a different and much smaller population than `ranked_lists[9]`.
 fn reciprocal_rank_fusion_entities(
     keyed_lists: &[Vec<(String, String, f32)>],
     k: f32,
@@ -12948,7 +12947,6 @@ fn reciprocal_rank_fusion_entities(
             )
         })
         .collect();
-    apply_entity_semantic_primacy(&mut combined, keyed_lists);
     combined.sort_by(|a, b| {
         b.2.partial_cmp(&a.2)
             .unwrap_or(std::cmp::Ordering::Equal)
@@ -12957,9 +12955,13 @@ fn reciprocal_rank_fusion_entities(
     combined
 }
 
-/// Signal-list index of the embedding arm, in `ranked_lists` and in the
-/// `keyed_lists` [`entity_granular_fused_files`] derives from it in the same
-/// order.
+/// Signal-list index of the embedding arm in `ranked_lists`.
+///
+/// Note that the `keyed_lists` [`entity_granular_fused_files`] derives from
+/// `ranked_lists` does NOT hold the same content at this index: when the
+/// embedding seeds resolve, index 9 is replaced by the entity-keyed SEED list.
+/// So a term that needs the file-level embedding ranking must read
+/// `ranked_lists[9]`, not `keyed_lists[9]`.
 const EMBEDDING_SIGNAL_INDEX: usize = 9;
 
 /// Semantic-primacy lift for the entity fuser, applied in place to the unlifted
@@ -13002,35 +13004,29 @@ const EMBEDDING_SIGNAL_INDEX: usize = 9;
 /// weight grows, peaks, then falls back to its unlifted place. Sweep a range and
 /// read the peak; do not push the weight up expecting more effect.
 fn apply_entity_semantic_primacy(
-    combined: &mut [(String, String, f32)],
-    keyed_lists: &[Vec<(String, String, f32)>],
+    fused: &mut [(String, String, f32)],
+    embedding_ranked: &[(String, f32)],
 ) {
     let weight = locate_env_f32("KIN_LOCATE_ENTITY_SEMANTIC_PRIMACY_WEIGHT", 0.0);
-    if weight <= 0.0 {
+    if weight <= 0.0 || embedding_ranked.is_empty() {
         return;
     }
     let k = locate_env_f32("KIN_LOCATE_SEMANTIC_PRIMACY_K", 8.0);
-    // Clamped strictly below 1.0: at exactly 1.0 a lifted key reaches `top` and
+    // Clamped strictly below 1.0: at exactly 1.0 a lifted entry reaches `top` and
     // the sort's key tiebreak could put it first, which is the invariant this
     // whole term is bounded by.
     let headroom = locate_env_f32("KIN_LOCATE_ENTITY_SEMANTIC_PRIMACY_HEADROOM", 0.9).min(0.99);
-    let Some(embedding_list) = keyed_lists.get(EMBEDDING_SIGNAL_INDEX) else {
-        return;
-    };
-    let embed_rank: FxHashMap<&str, usize> = embedding_list
+    let embed_rank: FxHashMap<&str, usize> = embedding_ranked
         .iter()
         .enumerate()
-        .map(|(rank, (key, _, _))| (key.as_str(), rank))
+        .map(|(rank, (path, _))| (path.as_str(), rank))
         .collect();
-    if embed_rank.is_empty() {
-        return;
-    }
-    let top = combined
+    let top = fused
         .iter()
         .map(|(_, _, score)| *score)
         .fold(0.0f32, f32::max);
-    for (key, _, score) in combined.iter_mut() {
-        let Some(&rank) = embed_rank.get(key.as_str()) else {
+    for (_, path, score) in fused.iter_mut() {
+        let Some(&rank) = embed_rank.get(path.as_str()) else {
             continue;
         };
         let lift = weight / (k + rank as f32 + 1.0);
@@ -13040,6 +13036,11 @@ fn apply_entity_semantic_primacy(
             *score += applied;
         }
     }
+    fused.sort_by(|a, b| {
+        b.2.partial_cmp(&a.2)
+            .unwrap_or(std::cmp::Ordering::Equal)
+            .then_with(|| a.0.cmp(&b.0))
+    });
 }
 
 /// Entity-granular fusion (behind `KIN_LOCATE_ENTITY_FUSION`, ON for accuracy-v2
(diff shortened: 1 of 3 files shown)


troyjr4103:
Closing this without landing. The pre-registered rule's clause 7 fails at every weight, and the measurement says the semantic-primacy term is the wrong instrument for the file it was aimed at.

## What the measurement found

Run 11:19Z to 12:10Z on this lane's own copy of the demo's microsoft/vscode store, `KIN_EMBED_BACKEND=cpu`, isolated `KIN_HOME` per arm. The control reproduces FIR-3123 exactly: `cursorTypeOperations.ts` at score 0.010491 against the ticket's 0.01049, `cursor.ts` at 0.002095 against 0.00210, embedding ranks 124 and 85 of 541 on the nose.

**The first cut was inert on its target.** `cursorTypeOperations.ts` scored bit-identically at 0.010491406 in the control and at every weight to 2.5, and a probe at weight 50 scored 0.010491406 too. Bit-identical at 200x the shipped default is not a weak effect, it is no effect.

**Why: index 9 survives the re-keying but its content does not.** `entity_granular_fused_files` replaces that slot with `entity_seed_keyed(embedding_seeds, ...)`, one `entity:{id}` per seed entity, whenever the seeds resolve. The 541-entry file-level embedding ranking at `ranked_lists[9]` is a different and much larger population. So the claim I published in this PR body and on FIR-3123, that index positions survive so index 9 is the embedding list in both fusers, is true of the index and false of the content. That is the same class of reason the graph-neighbourhood bonus does not carry over. I have corrected it here and on the ticket.

The last commit re-keys the term by path off `ranked_lists[9]` and adds a characterization test for the displacement, which is what a unit test on the term alone could not catch.

## Re-keyed, the term works, and clause 7 still fails

Two inert control arms bracket the treatment and came back **bit-identical across all 735 candidates**, so every move below is real and none of it is store drift.

| file | control | weight 0.5 | weight 50 | control 2 |
| -- | -- | -- | -- | -- |
| `cursorTypeOperations.ts` | 248 | 239 | 295 | 248 |
| `cursor.ts` | 519 | **316** | **201** | 519 |
| `viewController.ts` | 29 | 29 | 31 | 29 |
| `viewModelImpl.ts` | 42 | 38 | 42 | 42 |
| `codeEditorWidget.ts` | 2 | 2 | 2 | 2 |
| `textModel.ts` | 59 | 58 | 50 | 59 |

Rank one holds at every weight, file and entity both: `snippetParser.ts` and `ViewModel.type`. The guard does what it was built to do.

The term reaches its target now: the fusion score moves 0.024743883 to 0.028531762 at weight 0.5, against a predicted `0.5 / (8 + 124 + 1)` of 0.003759, matching to five decimals. It just does not move the file far enough. Clause 7 wants `cursorTypeOperations.ts` inside the fifty; it goes 248 to 239.

**And weight 50 makes it worse, 248 to 295.** That is the non-monotone saturation this PR body predicted from a simulation before any measurement ran: once the cap binds, capped keys keep their unlifted order, so a large weight returns the embedding set to the ranking it already had. Predicted before, measured after.

## The same hypothesis, tested a second way

The path fuser's own term is correctly keyed by construction. Under `KIN_LOCATE_ENTITY_FUSION=0`, sweeping `KIN_LOCATE_SEMANTIC_PRIMACY_WEIGHT` 0, 0.5, 2, 8, 32 moves the file's final rank 570, 559, 513, 411, 319. Sixty-four times the shipped default never brings it near fifty. At the fusion stage it climbs 643 to 152 and the post-fusion pipeline pushes it back out.

Both routes agree. The term lifts all 541 embedding-ranked files, so it lifts the competition too, and a file at rank 124 gains only its differential against files at worse ranks. That is the mechanism this PR body named as the risk before measuring.

## What this leaves

The rule was pre-registered so the goalposts could not move after the data arrived, and it says every clause or nothing ships. Clause 7 fails. Nothing ships.

Worth recording for whoever picks this up, without arguing it into a ship:

- The mechanism is safe and it does work. `cursor.ts`, at embedding rank 85, moves 519 to 201 at weight 0.5 with rank one untouched.
- `cursorTypeOperations.ts` reaches fusion at all only through the embedding route, and it is absent from multihop in every run. It holds `TypeOperations.typeWithInterceptors`, where a typed character becomes an edit, and `viewModelImpl.ts` calls into `cursor.ts` directly. A graph arm that cannot reach a file one hop from a file it did find is the more likely defect, and no fusion tuning is the right instrument for it.
- The branch keeps all three commits, so a future lane inherits the corrected implementation, the guard, and the characterization test rather than starting over.

Full numbers, the void first sweep and what caught it, and the daemon-isolation shape any later lane should copy: lane `darklift` report, 2026-09-03.

PR #1421, opened 2026-09-03
Resume a startup rollout lease its holder can no longer hold

troyjr4103:
## What this fixes

A hosted daemon that dies mid-bootstrap leaves the publication-control record holding its startup rollout lease, with its own image digest recorded as the admitted reader. `bootstrap_runtime_if_absent` required that digest to equal the running daemon's before it would resume the lease. A replacement image never carries the digest of the one it replaces, so the record stayed owned by a daemon that no longer existed, and no later image could open readiness.

`runtime_spine_authority` then reported that lease as `RolloutActive` without ever consulting `expires_at`, while `acquire_rollout` had always treated an expired lease as takeover-eligible. The read path and the write path disagreed about what a live lease is.

## The evidence

Production's record, `gs://kin-ecosystem-kin-graphs-dev/v2/.kin-graph-publication-control.json`, read 2026-09-03T02:50Z:

```
active_lease.kind         rollout
active_lease.holder       kin-daemon-startup-bootstrap
active_lease.fence        1
active_lease.acquired_at  2026-09-02T06:20:46.739774451Z
active_lease.expires_at   2026-09-02T06:25:46.739774451Z
reader.identity           sha256:fcedf4f9...  (the v0.6.3 daemon)
```

That lease expired twenty hours before the cutover attempt that it blocked. kinlab's Promote To Prod run 33706197746 rolled the daemon back at 02:33Z; production served 502 from 02:20:32 to 02:33:47.

## The change

An expired startup lease now resumes on the surviving startup identifiers alone: holder, request id and the three repository lists. Nothing is taken on trust. The takeover runs through `acquire_rollout`, which refuses while the lease is still live, raises the fence so the previous holder's proof is stale, and re-fences every graph object; `complete_hosted_startup_rollout` then re-admits the reader under the running image. A completed record carries no active lease at all, so a settled fleet is never reopened, and `startup_bootstrap_never_auto_admits_over_an_existing_reader` still holds.

Every path that refuses because of a rollout now reports whether the lease is live or lapsed, through one shared `BlockingRolloutLease::observe`. Telling an operator that a rollout "remains active" when it died hours ago sends them to wait for a holder that is already gone. That message cost this fleet an attempt.

## Deployment consequence

The recovery route the old message named, `POST /authority/publication-control/rollout/release`, is not reachable in the deployed shape: the failing startup probe kills the process that would serve it. So this cannot be worked around in production. It has to ship inside the image before any promotion can pass, which makes it a v0.6.6 item, alongside the stock-Mac init fix. The hosted cutover then rides that release rather than a production control-plane operation on an unrehearsed path.

## Tests

Three cases in `publication_lease.rs`:

- `a_dead_startup_bootstrap_lease_does_not_brick_the_next_image` reproduces the production record exactly, with a foreign reader identity, and asserts the takeover raises the fence, staleness of the predecessor's proof, and reader re-admission under the new image.
- `a_live_startup_bootstrap_lease_still_refuses_a_foreign_image` is the negative control: expiry is the only thing that makes a startup lease recoverable, and the record is untouched while the lease runs.
- `an_expired_operator_rollout_reads_as_expired_and_is_left_alone` holds the boundary. An operator's rollout stays the operator's to resume or replace, but the refusal now says the lease died.

Not yet falsified locally: this box is under a compute quiet for an overnight bench round, so no builder was taken. `rustfmt` parses and formats clean. The falsification arm runs after the quiet lifts and the result is posted here. Draft until then.

### DIFF
--- a/crates/kin-daemon/src/bin/kin-daemon.rs
+++ b/crates/kin-daemon/src/bin/kin-daemon.rs
@@ -319,9 +319,11 @@ fn create_state(
                     Ok(kin_daemon::publication_lease::RuntimeSpineAuthority::Completed(
                         evidence,
                     )) => runtime.block_on(state.adopt_hosted_spine_rollout_fence(evidence)),
-                    Ok(kin_daemon::publication_lease::RuntimeSpineAuthority::RolloutActive) => {
-                        Err("a hosted rollout remains active; semantic readiness stays closed while the authenticated publication-control API resumes or replaces it".to_string())
-                    }
+                    Ok(kin_daemon::publication_lease::RuntimeSpineAuthority::RolloutActive(
+                        blocking,
+                    )) => Err(format!(
+                        "a hosted {blocking}; semantic readiness stays closed while the authenticated publication-control API resumes or replaces it"
+                    )),
                     Err(error) => Err(error.to_string()),
                 }
             };
--- a/crates/kin-daemon/src/publication_lease.rs
+++ b/crates/kin-daemon/src/publication_lease.rs
@@ -344,10 +344,58 @@ pub enum RolloutReleaseDisposition {
     CompletedExact,
 }
 
+/// The rollout lease occupying a publication-control record, as an operator
+/// needs to read it. Whether the lease is still live decides what to do about
+/// it: a live lease has a holder that will finish or lapse on its own, while an
+/// expired one has no holder left to wait for and needs the release route.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct BlockingRolloutLease {
+    pub holder: String,
+    pub fence: u64,
+    pub expires_at: DateTime<Utc>,
+    pub expired: bool,
+}
+
+impl BlockingRolloutLease {
+    /// Read a rollout lease as of `now`. Every path that refuses because of a
+    /// rollout builds its refusal from here, so none of them can drift into
+    /// calling a lapsed lease active.
+    pub(crate) fn observe(active: &ActivePublicationLease, now: DateTime<Utc>) -> Self {
+        Self {
+            holder: active.holder.clone(),
+            fence: active.fence,
+            expires_at: active.expires_at,
+            expired: active.expires_at <= now,
+        }
+    }
+}
+
+impl std::fmt::Display for BlockingRolloutLease {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.expired {
+            write!(
+                formatter,
+                "rollout fence {} held by {}, whose lease expired at {} and was never released",
+                self.fence,
+                self.holder,
+                self.expires_at.to_rfc3339()
+            )
+        } else {
+            write!(
+                formatter,
+                "rollout fence {} held by {} until {}",
+                self.fence,
+                self.holder,
+                self.expires_at.to_rfc3339()
+            )
+        }
+    }
+}
+
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum RuntimeSpineAuthority {
     Completed(kin_spine::SpineRolloutFenceEvidence),
-    RolloutActive,
+    RolloutActive(BlockingRolloutLease),
 }
 
 #[derive(Clone, Debug, Eq, PartialEq)]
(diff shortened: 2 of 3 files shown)


troyjr4103:
Owed falsification arm, written out so it does not need this session.

This box was under a compute quiet for the overnight bench round with zero builders declared and swap at 14.96 of 16 GiB, so no slot was taken and the crate tests have not run. `rustfmt` parses and formats the three edited files clean, which proves syntax and nothing else. Hosted CI is green over the full forty-seven-run set with zero failures.

After the quiet lifts, from the umbrella root:

```
.kin-coord/bin/heavy-slot.sh rolloutlease kin -- \
  cargo test -p kin-daemon --features gcs publication_lease
```

The three new cases are `a_dead_startup_bootstrap_lease_does_not_brick_the_next_image`, `a_live_startup_bootstrap_lease_still_refuses_a_foreign_image` and `an_expired_operator_rollout_reads_as_expired_and_is_left_alone`.

Falsify each by breaking exactly the behavior it guards, and require the named assertion to go red rather than any assertion:

1. Revert the `|| active.expires_at <= now` clause in the resume predicate at `publication_lease.rs:754`. Case one must fail at its `expect("an expired startup lease must be takeover-eligible")`, and cases two and three must still pass. That is the production deadlock, so a mutation that leaves case one green means the case is not reproducing the record.
2. Drop the `active.expires_at <= now` guard so the predicate accepts a foreign identity unconditionally. Case two must fail at `assert!(intruder.bootstrap_runtime_if_absent().unwrap().is_none())`. This is the arm that proves the fix did not simply delete the invariant.
3. Change `expired: active.expires_at <= now` in `BlockingRolloutLease::observe` to a constant `false`. Cases one and three must both fail on their `blocking.expired` assertions.
4. Remove the `active.holder == STARTUP_BOOTSTRAP_HOLDER` clause. Case three must fail, because an operator rollout would then be hijacked by startup.

If a mutant survives, suspect a second defence catching the input one step earlier before suspecting the mutation, and answer it with a new input only the target can catch rather than by deleting a guard.

Post the output here and take the PR out of draft only once every arm above has been run and the mutants have been reverted.

troyjr4103:
Falsification arm run. Baseline green, all four mutants caught, but only after the arm found a hole in the tests themselves.

## Round one found a surviving mutant

Deleting the `holder == STARTUP_BOOTSTRAP_HOLDER` clause from the startup resume predicate left every case green. That clause was untested, and nothing in the suite would have told us if startup recovery began hijacking an operator's rollout.

The cause was a second defence, not a weak mutation. The operator rollout the cases acquire differs from a startup lease in **both** the holder and the request id, so the request id refuses it one step earlier and the holder is never consulted. Fixed by adding an input only the holder clause can refuse, in fc73054, rather than by touching the guard: an operator rollout that reuses the startup request id, so every startup identifier matches except the holder. An operator chooses its own request id and nothing stops it choosing the startup one, which is exactly why the holder is checked.

## Round two, at fc73054

```
=== falsify kin#1421 :: 2026-09-03T12:04:33Z ===
head     fc7305412 Give the startup holder check an input only it can refuse
BASELINE PASS
--- mutant: revert the expiry clause, so a dead startup lease is unrecoverable again
    CAUGHT: a_dead_startup_bootstrap_lease_does_not_brick_the_next_image
--- mutant: accept a foreign identity unconditionally, deleting the no-auto-admit guard
    CAUGHT: a_live_startup_bootstrap_lease_still_refuses_a_foreign_image
--- mutant: report every blocking lease as live
    CAUGHT: a_dead_startup_bootstrap_lease_does_not_brick_the_next_image
--- mutant: drop the startup-holder clause, so startup hijacks an operator rollout
    CAUGHT: an_operator_rollout_reusing_the_startup_request_id_is_still_not_startup
0 modified file(s); head fc7305412
VERDICT: baseline green and all four mutants caught.
```

Ran through `heavy-slot.sh` on rust-heavy-c, one builder, released on exit at rc=0. Every mutation refuses unless its anchor matches exactly once, and the tree was asserted unmodified after each revert.

The second mutant is the one worth reading: deleting the no-auto-admit guard turns the live-lease case red, so this change did not weaken the invariant that `startup_bootstrap_never_auto_admits_over_an_existing_reader` protects. It only lets an expired startup lease be taken over.

Out of draft.

PR #1427, opened 2026-09-03
FALSIFICATION ARM, do not land: put the whole entity back on the last-known-good entry

troyjr4103:
**Not for landing. Close this once its shard has reported.**

This is the falsification arm for the guard on #1419. It restores the copy that branch removes:
`LkgEntry` carries the whole `Entity` again beside the fingerprint, and `record` fills both, which is
today's shape under a new name.

Two tests must go red under it, and nothing else in the workspace may:

- `kin-reconcile::lkg_retained_bytes the_lkg_retains_nothing_that_scales_with_an_entity_s_strings`,
  on the arm difference. The two arms carry 64-byte and 8,192-byte doc summaries over 256 entities,
  so an entry that reaches the entity carries about two megabytes more in the large arm.
- `kin-reconcile lkg::tests::an_entry_is_exactly_a_fingerprint`, on the size equality.

A guard that stays green under this arm is not a guard, which is why the arm is run rather than
asserted.

Related: FIR-3064

### DIFF
--- a/crates/kin-reconcile/src/cross_file.rs
+++ b/crates/kin-reconcile/src/cross_file.rs
@@ -41,6 +41,7 @@
 //! stays independent of repository size.
 
 use std::collections::{BTreeSet, HashMap, HashSet};
+use std::sync::Arc;
 
 use kin_index::{
     bare_entity_name, link_cross_file_incremental_with_completeness, FileParseCompletenessMap,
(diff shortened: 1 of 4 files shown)


troyjr4103:
Read and done. Both guards went red under this arm: the retained-bytes guard on its arm difference, by exactly 2,080,768 bytes, which is the whole doc-summary difference to the byte, and an_entry_is_exactly_a_fingerprint on the size. Fast gate lint and policy stayed green, so the arm weakened the code and not a test. #1419 is unchanged.

PR #1433, opened 2026-09-03
Pin the supervisor's own token read as config IO

troyjr4103:
Main went red at 9f18585d2 on the zero file-search invariant: kin#1426 made the supervisor token required by default, and the client read of ~/.kin/supervisor.token in crates/kin-daemon/src/supervisor.rs (supervisor_client_auth_token, line 1191) is std::fs. The guard runs in the check job, which grades main and not pull requests, so #1426 landed green and reddened main.

The read is the daemon's own credential file, config IO under the rule's carve-out, never repository content and never a semantic answer. This adds the function to the file's existing function-scoped allowlist entry with the reason spelled out.

Ran: `python3 scripts/verify-zero-file-search.py .` on this tree: PASSED, 202 files scanned, 11 with function-scoped exemptions. Falsified: with the pin removed the same run reports line 1191 and FAILED; restored, PASSED. The check job will not grade this PR either, so the local run above is the proof and main's next push run is the confirmation.

### DIFF
--- a/scripts/zero-file-search-allowlist.json
+++ b/scripts/zero-file-search-allowlist.json
@@ -298,7 +298,7 @@
     },
     {
       "file": "crates/kin-daemon/src/supervisor.rs",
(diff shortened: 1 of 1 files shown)


troyjr4103:
Superseded: kin#1434 landed the same function pin at 17752b39a while this was on CI.

PR #1445, opened 2026-09-03
fix(daemon): keep the hosted publication probe metadata-only and memoize source bodies

troyjr4103:
The hosted daemon was making 13.28M Cloud Storage requests a day against a bucket holding 0.6 GB. Two read paths account for effectively all of it, and neither is the lease loop. The control record took zero object writes in the 24 hours measured, so nothing here changes how leases are written.

## The publication probe was downloading snapshots

`PublicationGatedStorageBackend` forwards seventeen `StorageBackend` methods and did not forward `load_snapshot_cursor`. The trait default recovers the whole authority and throws the graph away. `DaemonState::probe_repo_publication` runs it on every hosted request, from a call site whose own comment describes it as "one metadata-only backend read".

The request counts prove it rather than the source reading does. `ReadObject` ran 14,184 ahead of `GetObjectMetadata` across 15,520 `ListObjects`, which is one excess body read per list. A forwarded override would have produced one excess HEAD per list, in the other direction.

## Source bodies were refetched per request

kin-db's `load_source_blob_bounded` memoized nothing across requests, so 6,834 objects absorbed 6.6M reads a day. That is roughly 970 fetches of each object daily and 423 GB of egress.

`SourceBodyMemoBackend` wraps the GCS backend at `open_gcs_backend`. The key carries the repository as well as the digest so one repo never serves another's bodies, absence is never memoized, and a bounded read re-applies the caller's ceiling on a hit.

## Expected effect

Roughly 45k requests a day against the 13.28M measured. That figure is arithmetic rather than a measurement, because nothing here is deployed yet.

## Tests

Both fixes are pinned by tests that assert on requests made rather than answers returned, since a caller on the expensive path returns the correct answer too. The gate test was proven to bite by removing the forward and watching it fail.

`cargo test -p kin-daemon --all-features` passes 1,415 with 0 failures. `kin-precheck kin` ran 20 gates and passed 20. Doctests are clean. `kin-parity` did not run and its command is recorded in the lane report.

## Follow-up

This is one instance of a class, tracked as FIR-3147. The same decorator inherits trait defaults for fourteen other methods, and the larger problem there is capability erasure rather than cost: wrapping a fully capable backend produces one that reports itself incapable. `load_prepared_workspace_graph` returning `Ok(None)` forever is the worst of them, because a permanent no is indistinguishable from a genuine cache miss, so prepared-state reuse does not fail on the hosted path, it quietly never happens.

That work should land against main after this rather than stack on it, so the change with measured cost evidence ships first.

### DIFF
--- a/crates/kin-daemon/src/bin/kin-daemon.rs
+++ b/crates/kin-daemon/src/bin/kin-daemon.rs
@@ -282,7 +282,7 @@ fn create_state(
                     // the primary graph until the GCS fence has completed.
                     let state = DaemonState::open_with_backend_and_publication_control(
                         layout,
-                        Box::new(backend),
+                        backend,
                         repo_id,
                         allowed_repo_ids,
                         Arc::clone(&publication_control),
@@ -296,7 +296,7 @@ fn create_state(
             };
             let state = DaemonState::open_with_backend_and_publication_control(
                 layout,
-                Box::new(backend),
+                backend,
                 repo_id,
                 allowed_repo_ids,
                 Arc::clone(&publication_control),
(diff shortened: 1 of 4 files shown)


troyjr4103:
Superseded by #1447, same commit 3a2880490. Reopened from `chore/lane-gcs-request-cost-kin` because `kin-lane merge land` requires a PR come from the lane's own branch, and this one was opened from `fix/gcs-request-cost`. The rename happened to satisfy the pre-push branch-prefix hook, whose accepted list is not the same set as the merge tool's landable-branch shape. No code change.

PR #1455, opened 2026-09-04
Let a release mint and promote on its preflight record alone

troyjr4103:
## What changes

A stranger that cannot run no longer holds a release. It labels one.

Founder decision, 2026-09-03: "I dont even want that to block. If I run out of limit. Ship anyways. Stranger is nice to have." Followed by the ruling to flip everything: the tag, the release and GitHub Latest all move on the machine preflight alone, with the pending label kept on the record so a reader can see which releases were measured.

## The bug this closes

The mint's proof gate has run `KIN_RELEASE_REQUIRE=preflight` since 2026-09-02 (`release-tag.yml:1817`, `release.yml:2474`). The listing that picks the sha it grades did not move with it. `release-tag.yml` selected on `(map(.name) | unique | length) == 2`, so a sha carrying only `preflight.json` was never proposed to the gate that would have admitted it. The mint then printed "no reviewed main commit carrying X has release proof records yet", wrote `needed=false` and exited 0. That path is a deliberate no-op rather than a decline, so it escalated nothing. Half the rail was relaxed and the half that picks the sha was not, silently, and nothing guarded that expression either way.

## Both binding sites, plus the flip

**`.github/workflows/release-tag.yml`**
- The candidate listing proposes any sha carrying `preflight.json`. The comment above it and the `fully-evidenced` echo now say what the code does.
- The mint records the state it tagged on. It has computed `stranger_state` since 2026-09-02 and wrote it only to `GITHUB_OUTPUT`, where nothing in the workflow read it back.

**`.github/workflows/release.yml`**
- The `Promote stable tag` step no longer exits early on `STRANGER_STATE != complete`. It warns, writes a step summary section naming the gap, and promotes. `promoted=true` on both paths.
- The release-body stamp stays, minus one sentence. It claimed "It stays a prerelease and is not GitHub Latest until that record lands", which the flip makes false, and a false claim on a public release page is worse than a missing one. It now reads: "It ships and is promoted anyway, so this notice is how you tell it apart from a release that was measured."

**`scripts/select-release-candidate.py`**
- `STRANGER_COMPLETE` / `STRANGER_PENDING` constants, in the vocabulary `check-release-proof-artifacts.mjs` already returns and both workflows already `case` on.
- Mint-eligible on `preflight.json` alone. The cut stands down rather than arming a newer sha it would only race.
- The `stranger` decision stays **above** that stand-down, so `release-cut.yml`'s stranger job (gated on `decision == 'stranger'`) still runs while the archives it must judge exist. A naive rewrite that short-circuits earlier makes both that decision and the archives refusal dead code; falsification arm 2 below is exactly that mutation.
- The expired-archives refusal still refuses. Its reason now says first contact can never be recorded for that candidate and the mint may tag it anyway, instead of implying the release is held.
- `stranger_state` is a top-level `Decision` field rather than a `details` key, so the JSON document and the `GITHUB_OUTPUT` line are the same value read from the same place. `details` carries the `mintable` list beside it.

**`.github/workflows/release-cut.yml`** republishes `stranger_state` as a `select` job output.

**`docs/release-bot.md`** lines 96, 111 and 163 state the new rule.

## Where a reader finds the label

Three places, all durable, none of them a log line:

1. The mint's own step summary, the "Release tag minted" table, now carries a `first contact` row reading either `**pending** (no stranger.env)` or `complete on arms <arms>, driven on the <driver> endpoint`.
2. The release body, behind the `<!-- kin-first-contact-proof -->` marker, unchanged except for the sentence the flip falsified.
3. `release.yml`'s promote step summary, which now reads "Promoted with first contact pending" instead of "GitHub Latest held".

The mint creates a lightweight tag ref, not an annotated tag object, so there is no tag annotation to write into; making it annotated would change what `gh api .../git/ref/tags/<tag> -q .object.sha` returns and break the mint's own post-verify. The step summary is the mint's record.

## Falsification

Every new assertion goes red when the behavior it guards is broken. The three workflow assertions carry `expect_assertion` arms that run on every CI invocation of the authority suite, and each was additionally run against the real file on disk:

```
ARM 1  revert mintable to require both records
       AssertionError: unexpected grading of aaaa...  (the sha should never have been graded)
ARM 2  move the mintable stand-down above the candidate block  (the ordering trap)
       AssertionError: 'stand-down' != 'stranger'
ARM 3  drop stranger_state from release-cut.yml's select outputs
       AssertionError: release-cut.yml's select job does not republish stranger_state
ARM 4  revert the archives refusal reason
       AssertionError: 'the mint may still tag it' not found in '...'
ARM 5  restore the both-records listing in release-tag.yml
       AssertionError: release-tag.yml's candidate listing requires a stranger record again...
ARM 6  restore the Latest hold in release.yml
       AssertionError: release.yml holds a release back from GitHub Latest again...
ARM 7  delete the STRANGER_STATE env binding in release-tag.yml
       AssertionError: release-tag.yml computes a stranger state and reads it back nowhere...
```

Every file was restored byte-identical afterwards (`diff -q` against pre-arm copies) and both suites re-run green.

The record test is behavioral, not a grep. `assert_the_mint_records_the_state_it_tagged_on` executes the real readback step twice through the existing `execute_tag_readback` harness, once per state, and compares the two step summaries. Two identical summaries fail. That harness needed the step's first-contact env taught to it, since it runs under `set -u`.

## Local gates run

```
python3 scripts/test-select-release-candidate.py     Ran 71 tests ... OK   (68 before)
python3 scripts/test-release-workflow-authority.py   green
node --test scripts/release-promotion-plan.test.mjs scripts/read-promotion-plan.test.mjs \
            scripts/check-release-proof-artifacts.test.mjs    # tests 82  pass 82  fail 0
ruby -ryaml                                          release-tag.yml, release-cut.yml, release.yml all parse
bash -n + a real run of the new summary block in both states
jq                                                   the new listing expression lists a preflight-only sha; the old one did not
```

`bin/kin-precheck kin` is running through the heavy slot as this opens; its cargo gates are slower than hosted CI and the release this unblocks is time-critical, so the rest is left to CI. These are Python, YAML and Markdown changes; no Rust is touched.

## Known follow-on, not fixed here

`release-promote.yml` selects candidates on `prerelease === true` (`scripts/release-promotion-plan.mjs`, `selectCandidates`). Now that nothing is held, it finds nothing, so a release promoted with a pending notice does not get that notice stripped if a stranger record lands later, and the "Release published without first-contact proof" alarm goes quiet. The notice stays true about what was known at publication. Worth its own ticket.

### DIFF
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -90,6 +90,7 @@ jobs:
       candidate: ${{ steps.select.outputs.candidate }}
       rc_run: ${{ steps.select.outputs.rc_run }}
       move: ${{ steps.select.outputs.move }}
+      stranger_state: ${{ steps.select.outputs.stranger_state }}
       reason: ${{ steps.select.outputs.reason }}
     steps:
       # The literal trusted context value, as in kin-registry-wave-land.yml.
@@ -872,8 +873,9 @@ jobs:
           {
             echo "## Both proof records are filed for \`${CANDIDATE}\`"
             echo ""
-            echo "The mint selects the newest fully evidenced sha in the version range on its"
-            echo "next evaluation, which is its cron or any completed CI run for a main push."
+            echo "The mint already selects any sha in the range carrying \`preflight.json\`. With"
+            echo "\`stranger.env\` filed too, the release it cuts also claims first-contact coverage"
+            echo "and is eligible for GitHub Latest instead of shipping as a prerelease."
           } >> "$GITHUB_STEP_SUMMARY"
 
   # No runner, so say so loudly and hand over the exact command. This is the
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -873,9 +873,10 @@ jobs:
           {
             echo "## Both proof records are filed for \`${CANDIDATE}\`"
             echo ""
-            echo "The mint already selects any sha in the range carrying \`preflight.json\`. With"
-            echo "\`stranger.env\` filed too, the release it cuts also claims first-contact coverage"
-            echo "and is eligible for GitHub Latest instead of shipping as a prerelease."
+            echo "The mint already selects any sha in the range carrying \`preflight.json\`, and the"
+            echo "release it cuts becomes GitHub Latest either way. With \`stranger.env\` filed too,"
+            echo "that release also gets to claim first-contact coverage instead of carrying a"
+            echo "pending notice saying nobody measured what a first-time user meets."
           } >> "$GITHUB_STEP_SUMMARY"
 
   # No runner, so say so loudly and hand over the exact command. This is the
--- a/.github/workflows/release-promote.yml
+++ b/.github/workflows/release-promote.yml
@@ -5,17 +5,25 @@ name: Release Promotion
 
 # Finishes a release whose first-contact proof arrived after its tag.
 #
-# The release chain has two tiers. release-tag.yml mints a tag on the machine
-# preflight alone, so a candidate is never held hostage to a driver; release.yml
-# publishes it as a prerelease and writes into its body that first-contact proof
-# is pending. GitHub Latest, which is what an installer resolves when a stranger
-# types the install line, still requires the complete stranger record.
+# release-tag.yml mints a tag on the machine preflight alone and release.yml
+# publishes and promotes it, writing into its body that first-contact proof is
+# pending. Nothing waits for the stranger any more, so what this workflow
+# finishes is the RECORD rather than the release: the pending notice has to come
+# back out once the record lands, or a measured release goes on claiming it was
+# never measured.
 #
-# Without this workflow that second tier would be a dead end: the record lands
-# on the evidence branch hours later and nothing would notice. A push to the
+# Without this workflow that would be a dead end: the record lands on the
+# evidence branch hours later and nothing would notice. A push to the
 # release-evidence branch cannot trigger anything either, because that branch is
 # an orphan carrying no workflow files, so GitHub resolves no workflow from the
 # pushed ref. Hence a sweep.
+#
+# KNOWN GAP, tracked as FIR-3152. selectCandidates in
+# scripts/release-promotion-plan.mjs still filters on `prerelease === true`,
+# which was how a held release was recognised. Nothing is held now, so this
+# sweep finds nothing, the notice is never stripped, and the
+# "Release published without first-contact proof" alarm never fires. Fixing it
+# means selecting on the pending marker instead of on the prerelease flag.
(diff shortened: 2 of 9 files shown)


troyjr4103:
`bin/kin-precheck kin` finished clean against the exact pushed sha.

```
kin-precheck: repo=kin tree=.../worktrees/strangermint-kin sha=d3fe298ab5bc50f5c4aa58223e14b1bba05a96bb branch=chore/lane-strangermint-kin
kin-precheck: 20 gates enumerated from the check job of .github/workflows/ci.yml
PASS     guard:check_private_repo_coupling.sh
PASS     guard:check_runtime_boundaries.sh
PASS     guard:test-release-policy-gate.sh
PASS     guard:test-windows-authority-legs.sh
PASS     guard:zero_file_search_guard.sh
PASS     guard:check-kin-vfs-compat.mjs
PASS     guard:check-rc-build-drift.mjs
PASS     guard:test-assertion-reachability.py
PASS     guard:test-guard-duplicate-release-run.py
PASS     guard:test-homebrew-release-gate.py
PASS     guard:test-install-checksum.py
PASS     guard:test-private-repo-coupling.py
PASS     guard:test-release-tag-evaluate-dispatch.py
PASS     guard:test-release-workflow-authority.py
PASS     guard:test-select-release-candidate.py
PASS     guard:verify-hydration-semantics.py
PASS     guard:verify-zero-file-search.py
kin-precheck: 20 gates ran, 20 passed, 0 failed, on kin d3fe298ab5bc50f5c4aa58223e14b1bba05a96bb
```

The second commit on this branch corrects a defect I found reviewing my own diff, before CI could.

The flip to promote-and-label left nine sentences saying a stranger-pending release stays a prerelease and waits for GitHub Latest. Three are read by people rather than by code: the mint's step summary claim and the two `::warning::` lines the mint and the release raise on the pending path. A fourth rides in the selector's decision JSON and its `GITHUB_OUTPUT` reason. Two of the nine predate this PR. A record describing a gate the code no longer has is worse than no record, because it reads as the explanation for something that never happens.

Correcting that prose then defeated one of the new assertions, and the assertion's own falsification arm caught it one commit after the arm was written. `assert_the_mint_records_the_state_it_tagged_on` searched the whole step summary for the word "pending", and the corrected complete-branch claim contains that word in the sentence saying the release does NOT carry a pending notice. It now reads the `| first contact |` row alone and additionally refuses a proven row that says pending. Same family as the comment-absorbed mutant noted below: a check satisfied by prose ABOUT the thing it checks.

Two trap classes this work paid for are catalogued in kin-ecosystem#301, and the umbrella doc correction landed as kin-ecosystem#300. The follow-on this PR creates is FIR-3152.

PR #1475, opened 2026-09-04
Finish the first-contact record after a release is promoted (FIR-3152)

troyjr4103:
## What this fixes

FIR-3152, plus the founder's ruling that the stranger must never block a release.

The promoter went blind the moment `release.yml` stopped holding Latest, and nothing said so. `selectCandidates` keyed on `prerelease === true`, which was how a held release was recognised. Once no release was ever a prerelease again, the sweep selected nothing, the pending notice was never stripped, and the missing-proof alarm could not fire because nothing was ever waiting. **v0.6.6 is the first release in exactly that state and it sits there now**, carrying a notice saying nobody measured it, with no path that would ever take the notice back out.

## Two states, two actions

A proven candidate now goes to one of two lists, and the split is not cosmetic.

`promote` is a release still held as a prerelease: npm ordering checks, notice out, flip to Latest, readback. Unchanged.

`clearNotice` is a release already Latest whose record has since landed: notice out, nothing else. Running the promote path on one of those would end in `gh release edit --latest` on whichever tag the sweep reached, which for anything but the newest release is a rollback, and would then fail its own readback assertion.

## The alarm gets a second threshold, not a second issue

A held release is stuck and six hours is the right impatience. A promoted one shipped and is merely unmeasured, which since 2026-09-04 is the normal state of every release the stranger has not reached. One threshold for both would ring on every healthy release six hours after it shipped and train every reader to skim the one issue that matters. Held stays at 360 minutes, promoted gets 1440, and the alarm body gained a `state` column because a table whose rows mean two different things has to say which is which.

## The strip is now coupled to its writer

Every fixture in `read-promotion-plan.test.mjs` built its own copy of the notice text. `release.yml` could reword it, `stripPendingNotice` could stop matching, and the whole suite would stay green while a promoted release kept claiming its proof was pending forever. That wording did change, in #1455. The new case reads the `<<NOTE` heredoc out of the workflow and strips that.

It also asserts the extraction found a notice before comparing anything, because an extraction that silently found nothing would compare two empty values and pass. That is the shape this fleet catalogued at `docs/traps.md` after it printed PASS over a fixture that never built.

## The hosted stranger can no longer redden the rail

Founder, 2026-09-04: "the damn smoke should never block release it is a nice to have ONLY ... it should be mainly run locally not in the CI."

`release-cut.yml`'s stranger job takes `continue-on-error: true` and writes a summary line naming the state whatever happened. A driver out of limit, an offline runner or a container that would not start is not a fact about the candidate.

**The trap that makes that dangerous, handled:** `continue-on-error` masks `needs.<job>.result` to `success` for every downstream job, including one whose upstream failed. `publish-stranger` gated on `needs.stranger.result == 'success'`, so under the flag alone it would have run over a failed arm and published nothing while reporting success. It now keys on the stranger job's own `recorded` output, which is written only after a record exists on disk.

`docs/release-bot.md` now documents `bin/kin-stranger run --archive` and `--npm` as the ordinary path with the hosted job as the optional convenience. The umbrella's `docs/fleet-operations.md` gets the same in a companion PR.

## Falsification

Every new arm was broken and went red on the exact behavior. Files restored byte-identical afterwards, suites green again.

```
ARM 1  release.yml's notice drifts from what the stripper matches
       not ok - stripPendingNotice strips the notice release.yml actually writes
ARM 2  the heredoc extraction finds nothing (the vacuous-pass shape)
       not ok - error: 'release.yml no longer writes the notice with a <<NOTE heredoc'
ARM 3  selectCandidates keyed on the prerelease flag alone again (the original bug)
       not ok - selectCandidates takes exactly the shape this design creates
ARM 4  an already-Latest release sent down the promote path (the rollback)
       not ok - a proven release that is already Latest is cleared, never re-promoted
ARM 5  one threshold for both states
       not ok - a promoted release gets a longer fuse than a held one
```

Arms 6 and 7 live in the authority suite as `expect_assertion` arms and run on every invocation: removing `continue-on-error` from the stranger job, and pointing `publish-stranger` back at `needs.stranger.result`.

## Gates run locally

```
node --test release-promotion-plan + read-promotion-plan   # tests 35  pass 35  fail 0   (was 30)
python3 scripts/test-release-workflow-authority.py         green, including the two new arms
ruby -ryaml                                                release-cut, release-promote, release.yml all parse
```

`bin/kin-precheck kin` is running as this opens; its tail goes in a comment. No Rust is touched.

### DIFF
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -647,6 +647,23 @@ jobs:
     needs: select
     if: needs.select.outputs.decision == 'stranger' && vars.KIN_STRANGER_RUNNER != ''
     runs-on: ${{ vars.KIN_STRANGER_RUNNER }}
+    # The stranger never reddens this rail. Founder decision, 2026-09-04: "the
+    # damn smoke should never block release it is a nice to have ONLY ... it
+    # should be mainly run locally not in the CI." A driver that ran out of
+    # limit, a runner that went offline, a container that could not start: none
+    # of those is a fact about the candidate, and a red cut run for any of them
+    # is a rail that looks broken while behaving exactly as designed. The
+    # stranger's home is `bin/kin-stranger` against the archive or the published
+    # npm bytes; this hosted job is the optional convenience.
+    #
+    # `continue-on-error` has TWO axes and only one of them is wanted here. It
+    # keeps the run green, and it ALSO masks `needs.stranger.result` to
+    # 'success' for every downstream job, including a stranger that failed. So
+    # `publish-stranger` below keys on this job's own `recorded` output instead,
+    # which is written only after a record exists on disk. Gating it on
+    # `.result` under this flag would publish nothing over a failed arm and
+    # report success for doing it.
+    continue-on-error: true
     # Three concurrent arms plus image and container setup. The arms run
     # concurrently, so the wall clock is one phase 1 plus one phase 2, not three
     # of each.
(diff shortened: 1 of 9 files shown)


troyjr4103:
`bin/kin-precheck kin` finished clean against the pushed sha.

```
kin-precheck: repo=kin tree=/Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/strangermint-kin sha=e8d296bfafeb0ea04b9575024fc7631cfb513772 branch=chore/lane-strangermint-promoterecord
PASS fmt cargo fmt -- --check
PASS clippy cargo clippy --all-targets --all-features -- <45 allows from .github/clippy-allowed-lints.txt>
PASS guard:check_private_repo_coupling.sh bash scripts/check_private_repo_coupling.sh
PASS guard:check_runtime_boundaries.sh bash scripts/check_runtime_boundaries.sh
PASS guard:test-release-policy-gate.sh bash scripts/test-release-policy-gate.sh
PASS guard:test-windows-authority-legs.sh bash scripts/test-windows-authority-legs.sh
PASS guard:zero_file_search_guard.sh bash scripts/zero_file_search_guard.sh
PASS guard:check-kin-vfs-compat.mjs node scripts/check-kin-vfs-compat.mjs
PASS guard:check-rc-build-drift.mjs node scripts/check-rc-build-drift.mjs
PASS guard:check-required-contexts.py python3 scripts/check-required-contexts.py
PASS guard:test-assertion-reachability.py python3 scripts/test-assertion-reachability.py
PASS guard:test-guard-duplicate-release-run.py python3 scripts/test-guard-duplicate-release-run.py
PASS guard:test-homebrew-release-gate.py python3 scripts/test-homebrew-release-gate.py
PASS guard:test-install-checksum.py python3 scripts/test-install-checksum.py
PASS guard:test-private-repo-coupling.py python3 scripts/test-private-repo-coupling.py
PASS guard:test-release-tag-evaluate-dispatch.py python3 scripts/test-release-tag-evaluate-dispatch.py
PASS guard:test-release-workflow-authority.py python3 scripts/test-release-workflow-authority.py
PASS guard:test-select-release-candidate.py python3 scripts/test-select-release-candidate.py
PASS guard:verify-hydration-semantics.py python3 scripts/verify-hydration-semantics.py
PASS guard:verify-zero-file-search.py python3 scripts/verify-zero-file-search.py
PASS node-test node --test <15 file(s) named by the check job>
kin-precheck: 21 gates ran, 21 passed, 0 failed, on kin e8d296bfafeb0ea04b9575024fc7631cfb513772
```

Invoked by absolute path with an explicit tree, because the relative form through heavy-slot exits 126 having graded nothing and prints no tally.

PR #1482, opened 2026-09-04
Fix Firestore post-status retry semantics

troyjr4103:
## Summary

- Distinguish successful Firestore reads, which require a complete JSON body, from successful atomic writes, whose received status is authoritative.
- Prevent Commit and head-CAS replay after a successful status with a truncated informational body.
- Keep three exact real-socket response-semantics tests in the required fast-gate context.

## Verification

```text
cargo test -p kin-spine --features firestore --locked firestore::transient_retry_tests -- --nocapture --test-threads=1
16 passed; 0 failed; 131 filtered out

python3 scripts/test-release-workflow-authority.py
exit 0

actionlint .github/workflows/ci.yml
exit 0

cargo fmt --all -- --check
exit 0

git diff origin/main...HEAD --check
exit 0

cargo clippy --all-targets --all-features -- <repository reviewed lint-debt allows>
exit 0
```

## Falsification

Changing only the Commit call from the informational-body policy to the required-body policy made the exact no-replay socket regression fail with exit 101. Restoring the implementation made that same command pass 1/1. This proves the regression observes the replay boundary.

The branch is based on exact main SHA `a77b7c857b4a49dcd005a65e4ef51a17e3357bdd`. The exact PR head is `bf08478c0563560553bfe4b1657539fe03e79ccf`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1123,6 +1123,31 @@ jobs:
           # shellcheck disable=SC2086
           cargo clippy --all-targets --all-features -- $allows
 
+      # The deployed Firestore client needs a body to retry a query, but a
+      # successful Commit has already acknowledged its atomic write and must
+      # not replay just because its informational body truncates. The default
+      # workspace profile leaves this feature off. Keep the three real socket
+      # contracts in this required context, after its all-features build, with
+      # an exact listing assertion so a stale filter cannot grade nothing.
+      - name: Test Firestore post-status response semantics
+        shell: bash
+        run: |
+          set -euo pipefail
+          listing="$(cargo test -p kin-spine --features firestore --locked -- --list | tr -d '\r')"
+          for test_name in \
+            firestore::transient_retry_tests::query_documents_retries_a_body_that_fails_to_arrive_over_the_wire \
+            firestore::transient_retry_tests::commit_write_batch_does_not_replay_after_a_truncated_success_body \
+            firestore::transient_retry_tests::get_document_does_not_retry_a_permission_denied_with_a_truncated_body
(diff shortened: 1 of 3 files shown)


troyjr4103:
Independent exact-head review (read-only, opus-1m reviewer under the launch captain) at head bf08478c0563560553bfe4b1657539fe03e79ccf: GO. Full review: `.kin-coord/reports/showhn-review-1482-20260904.md` in the umbrella.

Checkable evidence for the one claim hosted CI does not itself grade, the focused suite on this exact head:

```
.kin-coord/bin/heavy-slot.sh review1482 kin -- cargo test -p kin-spine --features firestore --locked firestore::transient_retry_tests -- --test-threads=1
test result: ok. 16 passed; 0 failed; 0 ignored; 0 measured; 131 filtered out; finished in 1.56s
```

Boundary check: firestore.rs holds five Firestore call sites, all through the retry helper; the three reads (list_all_documents, get_document, query_documents) parse their success body, so Required is necessary there; the two writes (commit_write_batch and the head CAS) ignore the body on success by construction. The falsification claim holds by construction as well: the no-replay test scripts one response and the listener closes after it, so a required-body Commit retries into connection-refused and the `.expect` exits 101.

Two follow-ups, neither blocking: the diff also makes a truncated 403 fail fast and stops retrying a truncated 409 carrying ABORTED (the safe direction; worth a sentence in the summary), and in scripts/test-release-workflow-authority.py FIRESTORE_RESPONSE_BODY_CONTRACT_COMMAND is a strict prefix of the listing constant, so the execution-line condition cannot fail on its own and needs a mutation arm.

PR #1484, opened 2026-09-04
Refuse a commit while a merge transaction is open (FIR-3171)

troyjr4103:
Refuses `kin commit` while a merge transaction is open, before forced admission can advance the
workspace generation the transaction is bound to. FIR-3171 part one.

## The defect

The v0.6.6 stranger parked a twenty-conflict merge, hand-authored a resolution into the working
copy, and ran `kin commit` to take it. Kin accepted it:

```
$ kin commit -m "hand-authored merge resolution"
Created semantic change 5a6e6de0... on branch 'refs/heads/main' (7 entities, 0 relations, 1 artifacts)
exit 0
```

An ordinary single-parent commit published onto `refs/heads/main` while an unresolved merge
transaction was held against `refs/heads/main`. They then settled all twenty conflicts and the
publish failed:

```
$ kin resolve --do-continue
Error: ... HTTP 409 Conflict and published nothing: workspace 0dd72637-... has moved since this
merge opened, so publishing it would not be the merge that was resolved; abandon it with
`kin resolve --abort` and merge again
```

Twenty settled conflicts, gone. Git refuses the same commit outright with "you have unmerged paths".

## The mechanism

`command_commit` forced filesystem admission before it dispatched repository commit planning. A
merge record is bound to the workspace generation that opened it, so a successful unrelated commit
advanced that generation and made the parked transaction stale. The refusal has to run after the
command coordination gate is taken and before forced admission, which is where this puts it.

The open-merge summary rides in the existing publication-keyed authority admission cache rather than
in a second authority open, so a commit on a repository with no open merge pays nothing new. The
structural counter that proves one complete authority open per publication still passes; it is named
in the evidence below because it is the test that would catch a regression here.

The refusal names the exact transaction, how many of its conflicts are settled, and both ways out.

## Evidence

Every command below ran through the fleet heavy slot against this branch's worktree. Local gate
output is quoted rather than described, because hosted CI does not run any of it.

### The regression fails without the fix

The mutation removes the guard call itself, not an assertion. `crates/kin-daemon/src/api.rs:6222`,
`refuse_commit_during_open_merge(&state)?;`, replaced by a comment. The build proves the mutation
landed rather than being absorbed by a second defence:

```
warning: function `refuse_commit_during_open_merge` is never used
   --> crates/kin-daemon/src/api.rs:696:4
```

```
.kin-coord/bin/heavy-slot.sh vcsedge kin -- bash -c 'cargo build -p kin-daemon --bin kin-daemon \
  && cargo test -p kin-cli --test repository_merge_resolution \
  a_commit_while_a_merge_is_in_progress_is_refused_before_admission -- --nocapture'
```

2026-09-04T16:13:02Z, rc=101:

```
thread 'a_commit_while_a_merge_is_in_progress_is_refused_before_admission' panicked at
crates/kin-cli/tests/repository_merge_resolution.rs:929:5:
commit must not publish around an open merge: stdout=Created semantic change
a38f1506151e5c0af720d7c7844565aba3f7c99d168ea3819c7cb8ffe4dc5afe on branch 'refs/heads/main'
(2 entities, 0 relations, 1 artifacts)
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 25 filtered out; finished in 6.12s
```

The unguarded daemon published a single-parent change onto the merge's own target branch, which is
the reported defect exactly.

### The regression passes with it, and so does the whole merge suite

The guard was restored by the inverse edit rather than by `git checkout --`, and
`git status --porcelain` printed nothing afterwards, so the tree went back to exactly these bytes.
The daemon was rebuilt between the two arms on purpose: the CLI fixture's compatibility probe keys
on commit build identity and will accept a previously built daemon when only dirty source differs,
so a CLI-only rerun would have graded the old bytes.

```
.kin-coord/bin/heavy-slot.sh vcsedge kin -- bash -c 'cargo build -p kin-daemon --bin kin-daemon \
  && cargo test -p kin-cli --test repository_merge_resolution -- --nocapture'
```

2026-09-04T16:15:37Z, rc=0:

```
test a_commit_while_a_merge_is_in_progress_is_refused_before_admission ... ok
test result: ok. 26 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 25.10s
```

That suite covers the new refusal beside restart durability, stale identity, concurrent settlement,
completion, abort, path collision and the non-user semantic move.

### Whole crates

`cargo test -p kin-daemon` (2026-09-04T15:17:03Z) exited 0 with the daemon library reporting 1,368
passed, 0 failed and 3 ignored measurement cases, and every daemon integration binary passing.
`cargo test -p kin-cli` (15:22:41Z) exited 0 across the library and every CLI integration binary.
The structural counter that would catch a second authority open,
`the_admission_pair_costs_one_open_per_publication`, passed on its own at 15:11:02Z: the open-merge
summary rides in the existing publication-keyed cache, so a commit on a repository with no open
merge pays nothing new.

### Parity, both runs

`bin/kin-parity` resolved and printed the tree before grading it: checkout
`.kin-coord/worktrees/vcsedge-kin`, head `13bdf2f9738c4898e98bb22b8002faba3036b789`, branch
`chore/lane-vcsedge-kin`, `dirty False`. Release build exited 0 in 696.0s and the binary stamps the
checkout's own HEAD.

First run, 15:53:29Z:

```
kin-parity: magic       PASS                          26 pass / 0 fail / 0 unreadable
kin-parity: brownfield  FAIL                  479.0s  10 pass / 1 fail / 1 unreadable
kin-parity: firstrun    PASS-WITH-ALLOWANCES  104.6s   9 pass / 2 fail / 0 unreadable
kin-parity: FINAL FAIL in 1435.0s  (NON-CITABLE)
```

Both first-run failures are on parity's own allowance list by name (FIR-2580, and FIR-2501 for the
FIR-2554 doctor projection row), as is brownfield check 4 (FIR-2464). The one unallowed red was
brownfield check 9, FIR-2639: "2 staging path(s) survived the re-run unreferenced".

That check is load-sensitive, not this change. It PASSED at this exact head 28 minutes earlier in
another lane's run (`.kin-coord/parity/20260904T152544Z/parity.log:119`) and at
`48f84cd7f206a31464509cd5a135bc0e3e71cd27` on 09-03 (`20260903T002908Z/parity.log:592`), both times
with the sentence "the re-run reclaimed 2.8 MB of staging and said so; nothing is left beside the
repository". The failing run carried a kin-daemon build and a 26-test CLI suite in a second builder
slot while brownfield ran. The check kills `kin init` mid-ladder on a poll and asserts on what the
re-run reclaims, which is exactly the shape that moves under CPU contention.

Second run, same binaries, `--skip-build --only brownfield`, 16:18:08Z, rc=0:

```
CHECK 9 FIR-2639 PASS trend-unknown the re-run reclaimed 2.8 MB of staging and said so; nothing is
  left beside the repository
kin-parity: brownfield  PASS-WITH-ALLOWANCES  423.0s  11 pass / 0 fail / 1 unreadable
kin-parity: FINAL PASS-WITH-ALLOWANCES in 423.5s  (NON-CITABLE)  allowances: brownfield/4 FIR-2464
```

Both runs are here rather than only the second, because a reader who sees only the pass cannot tell
a flake that was chased down from one that was never seen.

### Precheck, and the gate it caught

The first run, on the pre-repair commit, was red on one gate, and it was right:

```
FAIL     guard:verify-zero-file-search.py     python3 scripts/verify-zero-file-search.py  (exit 1)
         | [VIOLATION] crates/kin-daemon/src/api.rs contains forbidden filesystem access:
         |   Line  650: lease.metadata(), (filesystem metadata probe)
         |   Line 8861: .metadata() (filesystem metadata probe)
         | Allowlist validation FAILED:
         |   entry crates/kin-daemon/src/api.rs allow_match '.metadata()' occurs 2 times in
         |   scanned code (want 1)
```

`lease.metadata()` reads a repository-v6 authority lease and returns graph-owned truth, not a
filesystem probe, but it shares a method name with one in the deny set, so the allowlist pins the
exact occurrence count per file and any new call has to be re-justified. This change had added the
second one to the daemon RPC surface.

The repair is in the diff, above, and it is not a bump on `api.rs`. That file's own allowlist reason
says it "is the answer authority every agent, editor, and CLI reaches, so it is scanned in full and
only the explicit IO-boundary bodies are excused". `repository_merge_state.rs` already carries a pin
whose reason is "Every `.metadata()` in this module is called on a repository-v6 authority lease and
returns graph-owned authority metadata", and it already owns `workspace_merge_record`, the function
this code was calling. So the accessor moved there as `open_merge_summary(lease, workspace_id)`,
`api.rs` dropped its local struct and now keeps one `.metadata()`, and the merge-state pin goes 2 to
3 with the third named in its reason. It takes no extra authority open: the existing lease is passed
through, and `record_authority_open()` is counted inside
`LocalRepositoryAuthorityContext::open` (`local_repository_authority.rs:64`).

The pin was falsified rather than trusted. At its pre-bump value the guard fails and names the new
call by count:

```
Allowlist validation FAILED:
  entry crates/kin-daemon/src/repository_merge_state.rs allow_match '.metadata()' occurs 3 times
  in scanned code (want 2)
```

and at 3 it passes with its own deny-set self-test green:

```
Deny-set self-test passed: 30 matcher cases (6 negative controls) and 6 binding-learner cases
(2 negative controls).
Verification PASSED: Zero File-Search Invariant holds.
```

The run of record, on this head, 2026-09-04T16:41:51Z, rc=0:

```
kin-precheck: repo=kin tree=.../worktrees/vcsedge-kin sha=b79320b3f22f4660704bc2bc78ebd7c216a1cfc7
              branch=chore/lane-vcsedge-kin
kin-precheck: 21 gates ran, 21 passed, 0 failed, on kin b79320b3f22f4660704bc2bc78ebd7c216a1cfc7
```

No unknown script and no DROP, so the gate list is enumerated from this tree's ci.yml rather than
remembered. The merge suite was re-run against the moved accessor and is 26 passed, 0 failed.

## Base

Opened from `chore/lane-vcsedge-kin` at `b79320b3f22f4660704bc2bc78ebd7c216a1cfc7`, a few commits
behind `origin/main` and sharing no file with anything that landed since it was cut. The falsification
arm and the crate runs quoted above were taken on `13bdf2f97`, the same change before the
zero-file-search repair moved one lease accessor between modules; the merge suite was re-run on this
head afterwards and precheck's run of record is on this head.

## What this does not do

This is part one. `kin resolve` still accepts no hand-authored body, so there is no
`--take-worktree` and no `git add` equivalent, and the stale-conflict guidance after an aborted
merge is unchanged. Those stay open on FIR-3171.

### DIFF
--- a/crates/kin-cli/tests/repository_merge_resolution.rs
+++ b/crates/kin-cli/tests/repository_merge_resolution.rs
@@ -901,6 +901,68 @@ fn a_second_merge_while_one_is_in_progress_is_refused() {
     assert!(record.state.is_in_progress());
 }
 
+/// A commit cannot publish around a parked merge. Forced admission advances
+/// the workspace generation, so allowing the commit would make the recorded
+/// restore point stale and strand every resolution already carried by it.
+#[test]
+fn a_commit_while_a_merge_is_in_progress_is_refused_before_admission() {
+    let root = tempdir().expect("temp root");
+    let repo = conflicting_repository(root.path());
+    let runtime = common::IsolatedDaemonRuntime::new(&repo);
+    let layout = initialize_kin_repo(&runtime, &repo);
+
+    parked_merge(
+        &run_kin(&runtime, &repo, &["merge", "feature"]),
+        "kin merge",
+    );
+    let main_before = branch_change(&layout, "main");
+    let parked = persisted_record(&layout).expect("the merge is parked");
+    let parked_hash = hex::encode(parked.hash.as_bytes());
+    let hand_authored = b"pub fn hand_authored_resolution(value: u64) {}\n";
+    fs::write(repo.join("src/lib.rs"), hand_authored).expect("write a hand-authored resolution");
+
+    let refused = run_kin(
+        &runtime,
+        &repo,
+        &["commit", "-m", "hand-authored merge resolution"],
+    );
+    assert!(
+        !refused.status.success(),
+        "commit must not publish around an open merge: stdout={} stderr={}",
+        String::from_utf8_lossy(&refused.stdout),
+        String::from_utf8_lossy(&refused.stderr)
+    );
+    let stderr = String::from_utf8_lossy(&refused.stderr);
+    assert!(
+        stderr.contains("HTTP 409 Conflict"),
+        "the open transaction is a conflict rather than a daemon fault: {stderr}"
+    );
+    assert!(
+        stderr.contains(&parked_hash),
+        "the refusal names the exact merge transaction: {stderr}"
+    );
+    assert!(
+        stderr.contains("kin resolve --do-continue") && stderr.contains("kin resolve --abort"),
+        "the refusal names both ways out of the open transaction: {stderr}"
+    );
+
+    assert_eq!(
+        branch_change(&layout, "main"),
+        main_before,
+        "the refused commit does not move the target branch"
+    );
+    assert_eq!(
+        persisted_record(&layout),
+        Some(parked),
+        "the refusal happens before admission and leaves the exact merge record intact"
+    );
+    assert_eq!(
+        fs::read(repo.join("src/lib.rs")).expect("read the authored resolution"),
+        hand_authored,
+        "the refusal preserves the operator's working-copy resolution"
+    );
+}
+
 /// Settling names one identity at a time, and a name that matches nothing is
 /// refused rather than quietly settling the wrong conflict.
 #[test]
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -223,8 +223,9 @@ pub(crate) struct ProjectionAuthorityCache {
     held: std::sync::Mutex<Option<HeldProjectionAuthority>>,
     query: std::sync::Mutex<Option<HeldQueryAuthority>>,
     command: std::sync::Mutex<Option<HeldCommandAuthority>>,
-    /// The admission pair every reconcile task needs, which is not a fourth
-    /// wrapper over the authority but two small values read out of one.
+    /// The admission values every reconcile task needs, plus the small
+    /// open-merge summary a commit gate needs. This is not a fourth wrapper
+    /// over the authority, only values read out of one.
     ///
     /// `current_authority_admission` wants a root bundle and a compiled
     /// admission matcher, and neither can differ between tasks inside one
@@ -277,6 +278,7 @@ struct HeldAdmission {
     published: LocalPublicationIdentity,
     roots: kin_model::RootBundle,
     policy: Option<kin_index::ResolvedAdmissionMatcher>,
+    open_merge: Option<crate::repository_merge_state::OpenMergeSummary>,
 }
 
 impl ProjectionAuthorityCache {
@@ -336,30 +338,15 @@ impl ProjectionAuthorityCache {
         });
     }
 
-    fn reuse_admission(
-        &self,
-        published: &LocalPublicationIdentity,
-    ) -> Option<(
-        kin_model::RootBundle,
-        Option<kin_index::ResolvedAdmissionMatcher>,
-    )> {
+    fn reuse_admission(&self, published: &LocalPublicationIdentity) -> Option<HeldAdmission> {
         lock_recover(&self.admission)
             .as_ref()
             .filter(|held| &held.published == published)
-            .map(|held| (held.roots.clone(), held.policy.clone()))
+            .cloned()
     }
 
-    fn install_admission(
-        &self,
-        published: LocalPublicationIdentity,
-        roots: kin_model::RootBundle,
-        policy: Option<kin_index::ResolvedAdmissionMatcher>,
-    ) {
-        *lock_recover(&self.admission) = Some(HeldAdmission {
-            published,
-            roots,
-            policy,
-        });
+    fn install_admission(&self, admission: HeldAdmission) {
+        *lock_recover(&self.admission) = Some(admission);
     }
 
     fn reuse_command(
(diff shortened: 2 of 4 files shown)


troyjr4103:
Independent exact-head review (review1484, read-only, opus-1m) on b79320b3f22f4660704bc2bc78ebd7c216a1cfc7: **Verdict: GO**. Report: `.kin-coord/reports/showhn-review-1484-20260904.md` in the umbrella.

What was checked: the gate fires only when `workspace_merge_record` finds a record for that exact workspace with `state.is_in_progress()` true; it returns 409 naming the transaction hash, both refs, settled and total counts; it reuses the lease already taken for `roots` (api.rs:638-642) so no second authority open; it sits before the forced admission (api.rs:6203) so the parked record is untouched on refusal; cache staleness is closed because `merge_transactions` folds into `local_state_root` (kin-db 0.7.101 repository.rs:8499) and the cache key is a per-call hash of the durable publication record.

Falsified by the reviewer: the zero-file-search allowlist pin goes red at 2 with the exact ci.yml invocation ("occurs 3 times in scanned code (want 2)", lines 81, 125, 477), passes at 3 on the head, rc=0; the api.rs entry is byte-identical to base.

Gap noted, accepted: the red arm was run on 13bdf2f97 rather than the head (the accessor moved modules between them, gate logic and call site unchanged; merge suite 26 of 26 and precheck 21 of 21 re-run on the head). Hosted checks: 47 listed, 47 total, 30 success, 17 skipped, zero failures, CLEAN.

Landing condition from the captain: parity verdict on this head green, then the lane enqueues and lands.

PR #1485, opened 2026-09-04
Qualify the first-contact answers Kin cannot certify

troyjr4103:
A stranger's first questions get answers Kin cannot back, and each one reads as confident. This
bundle makes six of those surfaces state their own limits. It changes no answer that was already
honest, and it does not widen what Kin can prove.

## What was wrong, and where

**An absent concept received a certified answer.** `qualifies_populated_answers` excludes
`semantic_locate`, so a populated locate page gets no `negative` block, the absence gate reads
`Silent`, and a silent input certifies nothing but refuses nothing either. The daemon does record an
`all_fallback` degradation, and `describes_advice_not_run_quality` deliberately filters it out of the
run-quality verdict, correctly: the run lost no capability. The result was that
`semantic_locate("password hashing and session token expiry")` on a repository with no auth code came
back with four confident neighbours and `_kin.verdict.state: certified`, `limiting_factor: null`.

Answered at the absence gate, which is the seam that owns "how far can these rows be trusted", rather
than by folding a query's shape into a verdict about how well the query ran. A fallback-only page for
a prose query now carries `kind: relevance_unverified`, `interpretation: nearest_neighbors_only` and a
`relevance_floor_unmeasured` reason, and the `_kin` verdict follows it to `inconclusive`. It does not
say the concept is absent, because nothing in the response can tell an absent concept from a correct
neighbour, and that is the whole reason it cannot be certified.

This replaces a test that asserted the opposite. Its argument was that `all_fallback` fires on prose
queries whose top hits are correct, so a qualifier there is one an agent learns to ignore. That holds
against the `unnamed_ranking` wording, which asserts a named symbol was not found and would be false
over a correct hit. It does not hold against this one, which asserts only that the ranking publishes
no measured relevance threshold, a sentence equally true either way. The argument is now answered in
the code rather than deleted with the test.

**A schema required properties it did not define.** `trim_schema` filters `properties` and the
top-level `required` against the keep list and touches nothing else, so `get_context_pack` shipped an
`anyOf` offering `entity_id`, `entities` and `question` while defining only the first, on all three
belt profiles. Two of its three documented ways to be called were required by the served schema and
absent from it. Separately, the clipped-chain remediation tells a caller to widen `limit_per_step`
while the belt had trimmed that property away, so the advice was unfollowable and a five-wide cap
could not be raised through the advertised schema at any budget.

**Status replayed an unrelated clock.** `kin_graph_status` can answer from the last settled sample
when a live one cannot be taken. The counters said so under `stale`, while `_kin.freshness` still
carried the HEAD admission clock, so a six-minute-old `entity_count: 0` sat beside
`state: recorded, age_seconds: 6`. `_kin.freshness` now carries the selected graph's own stale
observation, and the stdio contract rejects any response where the two disagree in either direction.

**A covering-test count read as a claim about the working copy.** `entity_impacts[].covering_tests` is
a graph-observed lower bound, and nothing beside the number said so, so a `0` from a missing
local-variable property edge published as a confident zero. Every serialized row now carries
`covering_tests_bound: graph_observed_lower_bound`, through one annotation seam that serves both the
direct impact response and the impact nested inside `semantic_review`. **This is disclosure, not a
repair: the number is still 0.** The enrichment gap under it is untouched.

**A correct install looked broken.** Outside a Kin repository there is no root for a live shell hook
to project, and the projection row still read `Stale` and counted toward `needs_attention`. The
post-fix footer derived its own attention predicate rather than calling the one the summary uses, so
`unsupported` rows the summary correctly called not applicable reappeared under "Still needs manual
steps". The readiness line now also names which rows are expected first-run work and which are host
limits.

**Uncertainty had to survive a budgeted cut.** The budget already excludes `_kin` and `negative` and
rebuilds the verdict on any pass that cut rows. What was missing was a regression holding that
property over a first-contact impact response, so the working-copy-behind caveat, the response-bounded
factor and the new covering-test bound are all asserted after 120 rows are cut to a 5,000-character
ceiling.

## Two things found while verifying the fixes

The `limit_per_step` short description told agents to read `truncated_steps`, a response field that
exists nowhere in Kin; the real disclosure is `clipped_steps`, carrying `fanout_truncated` and
`fanout_dropped`. A fix for "no nonexistent parameters shown to users" had shipped a nonexistent field
name of its own. Corrected, and the class now has a check rather than a spot fix:
`no_short_property_description_names_an_identifier_the_tool_never_publishes` asserts that every
backtick-quoted identifier in the per-tool short-description table appears in that tool's own
registered description or schema. It passes over the whole table, so that was the only one.

The schema defect is likewise answered by a sweep rather than by naming two properties:
`no_belt_profile_serves_a_schema_that_requires_a_property_it_does_not_define` walks all three belt
profiles and asserts no served schema requires a property its own `properties` map omits, with an
anti-vacuity control that fails if it stops reading combinator branches, which is the only place the
defect can live.

## Verification

Claims hosted CI does not itself check are recorded here as the command and its output, never as
prose. Eight new tests, one rewritten and two extended, were falsified across fourteen pairings, one
mutant at a time, each mutant the exact reverse of a product change; files were restored from text held
in memory and the tree proved byte-identical afterwards.

Every gate below ran on this branch's head `6c530a366d1dd34182f2a1416929c806ec0a0ca7`, through
`.kin-coord/bin/heavy-slot.sh answertruth kin` from the umbrella root, each printing the worktree and
target it resolved before grading.

```text
$ heavy-slot.sh answertruth kin -- cargo fmt --all -- --check
heavy-slot: answertruth command exit rc=0 at 2026-09-04T17:00:19Z

$ heavy-slot.sh answertruth kin -- cargo test -p kin-mcp
kin-lane: run lane=answertruth repo=kin worktree=.../worktrees/answertruth-kin
   target=.../cargo-target/answertruth-kin embed=cpu
test result: ok. 891 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 20.42s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
heavy-slot: answertruth command exit rc=0 at 2026-09-04T17:01:23Z

$ heavy-slot.sh answertruth kin -- cargo test -p kin-cli
running 2145 tests
test result: ok. 2144 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 165.30s
(2989 passed, 0 failed across every binary in the crate)
heavy-slot: answertruth command exit rc=0 at 2026-09-04T17:13:31Z

$ heavy-slot.sh answertruth kin -- bin/kin-precheck kin --repo-dir kin=<this worktree>
kin-precheck: repo=kin tree=.../worktrees/answertruth-kin sha=6c530a366d1dd34182f2a1416929c806ec0a0ca7 branch=chore/lane-answertruth-kin
kin-precheck: 21 gates ran, 21 passed, 0 failed, on kin 6c530a366d1dd34182f2a1416929c806ec0a0ca7
heavy-slot: answertruth command exit rc=0 at 2026-09-04T17:15:42Z
```

`kin-precheck` names the three check-job steps it deliberately does not reproduce, so the 21 is a stated
subset rather than a claim to have run the whole job: the language-server provisioning script, the
release policy gate whose failure routing needs GitHub event context that a local run cannot supply
truthfully, and the hydration falsification that only means anything against the poisoned copy the
workflow builds first.

`bin/kin-parity` ran on the same tree before the rebase, at `a77b7c857` dirty, and the rebase took in
only `ci.yml`, `release-cut.yml`, `crates/kin-spine/src/firestore.rs`, `docs/release-bot.md` and two
release scripts, none of which this bundle touches:

```text
kin-parity: build       release        1023.5s
kin-parity: magic       PASS            166.4s  26 pass / 0 fail / 0 unreadable
kin-parity: brownfield  PASS-WITH-ALLOWANCES  400.4s  11 pass / 0 fail / 1 unreadable
kin-parity: firstrun    PASS-WITH-ALLOWANCES  106.2s  9 pass / 1 fail / 1 unreadable
kin-parity: FINAL PASS-WITH-ALLOWANCES in 1698.4s  (NON-CITABLE)
            allowances: firstrun/9 FIR-2580, firstrun/3 FIR-2501, brownfield/4 FIR-2464
heavy-slot: answertruth command exit rc=0 at 2026-09-04T16:59:04Z
```

All three allowances are pre-existing tracked tickets and none names a file in this bundle.

Falsification, fourteen pairings, thirteen red:

```text
MUTANT: the absence gate stops qualifying a fallback-only prose page
  prose_query_over_fallback_hits_refuses_to_certify_relevance                  FAILED  rc=101
  a_prose_locate_page_of_neighbours_never_certifies_on_the_compact_surface     FAILED  rc=101
MUTANT: the belt trims limit_per_step back out of the trace schema
  agent_default_advertises_every_answer_recovery_and_input_alternative         FAILED  rc=101
MUTANT: the context-pack keep list returns to its shipped form
  no_belt_profile_serves_a_schema_that_requires_a_property_it_does_not_define  FAILED  rc=101
  agent_default_advertises_every_answer_recovery_and_input_alternative         FAILED  rc=101
MUTANT: impact rows stop carrying the covering-test bound
  impact_rows_carry_one_based_presentation_lines_beside_the_raw_span           FAILED  rc=101
  impact_bounds_and_working_copy_caveats_survive_response_budgeting            ok      rc=0    SURVIVED
MUTANT: the replayed sample stops owning the envelope's freshness
  replayed_graph_status_marks_the_standard_envelope_stale                      FAILED  rc=101
MUTANT: a budgeted cut stops downgrading the verdict and the absence object
  impact_bounds_and_working_copy_caveats_survive_response_budgeting            FAILED  rc=101
MUTANT: the projection row stops reading whether a repository is here
  a_live_hook_outside_a_repository_is_not_an_attention_row                     FAILED  rc=101
MUTANT: the post-fix manual footer returns to its own predicate
  the_post_fix_manual_list_excludes_not_applicable_rows                        FAILED  rc=101
MUTANT: the readiness line stops naming expected first-run work
  the_printed_count_is_the_tally_and_both_come_from_one_predicate              FAILED  rc=101
MUTANT: the short description names a response field again
  no_short_property_description_names_an_identifier_the_tool_never_publishes   FAILED  rc=101
MUTANT: the producer renames the value the hand-built fixture spells
  impact_bounds_and_working_copy_caveats_survive_response_budgeting            FAILED  rc=101
```

The survivor is a wrong pairing in the harness, not a test that cannot fail, and it is left in the
record rather than removed. The budget test builds its impact rows by hand, so deleting the producer's
annotation leaves the field present in that fixture. What that test guards is that a budgeted cut does
not strip the caveats from the rows it kept, and the mutant targeting exactly that, removing
`crate::verdict::mark_response_bounded` from the budget loop, turns it red. The other half, that the
producer adds the bound at all, is guarded by `impact_rows_carry_one_based_presentation_lines_beside_the_raw_span`,
which the first mutant does turn red. The key and its value are now one constant read by the producer
and both tests, so a rename cannot let the hand-built fixture drift green.

## What this does not fix

Three things, stated so nothing downstream reads more into this than it says.

`covering_tests: 0` for a genuinely covered entity is still `0`. The label beside it,
`covering_tests_bound: graph_observed_lower_bound`, says the number is a graph-observed lower bound
and not a claim about every test in the working copy. The enrichment gap that produces the zero,
coverage resolving through a constructor call but not through property access on an inferred local,
is untouched here.

A stranger report also names `impact_analysis` as telling a caller to widen `limit_per_step`. That
half is unclaimed. The only advice in this repository naming that parameter is `spine_clipping_gap`,
gated on `spine_clipped_steps`, and only `trace_data_flow` publishes that field
(`crates/kin-mcp/src/handlers/entities.rs:4086`). Nothing here changes `impact_analysis`, and the
report should not be read as covering it.

A prose locate page now qualifies its relevance whether or not its top hit is correct, because no
field in the response separates the two. That is a deliberate widening of when the caveat appears,
and it is the price of not certifying what cannot be certified.

## One thing this bundle is next to but does not fix

`bin/kin-parity`'s firstrun suite carries an allowance, FIR-2501 with its own ticket FIR-2554, saying
`kin doctor`'s projection row cannot go green under Kin's own shell hook: the hook strips the preload
from every `kin` invocation on purpose, and the row measures whether the shim is loaded into the doctor
process, which is the one process the hook exists to keep unshimmed. That is the inside-a-repository
half of the row this bundle touches. Nothing here changes it. The new branch is gated on
`!inside_repository`, and parity reproduced the identical allowance after the change.

## Landing note

`Product Acceptance` carries `if: ${{ github.event_name != 'pull_request' }}`, so it reports
`skipped` here and main's own push run is the grader for these changes.

### DIFF
--- a/crates/kin-cli/src/commands/health.rs
+++ b/crates/kin-cli/src/commands/health.rs
@@ -1267,7 +1267,9 @@ fn check_projection_mode() -> HealthCheck {
         }
     };
     let cwd = env::current_dir().unwrap_or_default();
-    let repo_root = kin_core::KinLayout::discover(&cwd)
+    let layout = kin_core::KinLayout::discover(&cwd);
+    let inside_repository = layout.is_some();
+    let repo_root = layout
         .map(|layout| layout.root().to_path_buf())
         .unwrap_or(cwd);
     let report = crate::commands::projection::report_for(
@@ -1278,7 +1280,7 @@ fn check_projection_mode() -> HealthCheck {
     );
     let outside = probe_outside_repo(home_dir().ok().as_deref(), report.shim.engaged);
     let hook = probe_shell_hook(&kin_home, detect_shell(), &report.shim.path);
-    projection_mode_check_for(&report, env::consts::OS, &outside, &hook)
+    projection_mode_check_for_context(&report, env::consts::OS, &outside, &hook, inside_repository)
 }
 
 /// Whether Kin's own shell hook is live in the shell that started this process.
@@ -1939,6 +1941,35 @@ fn projection_mode_check_for(
              available here"
         },
     )
+}
+
+/// Qualify the projection row by whether this doctor run has a repository to
+/// project. Kept as a separate seam so first-contact context can be tested
+/// without asking a fixture to discover the process working directory.
+fn projection_mode_check_for_context(
+    report: &crate::commands::projection::ProjectionReport,
+    os: &str,
+    outside: &OutsideRepoProbe,
+    hook: &ShellHook,
+    inside_repository: bool,
+) -> HealthCheck {
+    let check = projection_mode_check_for(report, os, outside, hook);
+    if !inside_repository && hook.is_live() && matches!(check.status, HealthStatus::Stale) {
+        return HealthCheck::new(
+            "projection_mode",
+            "Projection in force",
+            HealthStatus::Unsupported,
+            format!(
+                "this command is not inside a Kin repository, so there is no repository root for \
+                 the live shell hook to project; the hook itself is live; underlying probe: {}",
+                check.detail
+            ),
+        )
+        .with_platform_note(
+            "Run `kin doctor` from inside the Kin repository whose projection you want to check.",
+        );
+    }
+    check
 }
 
 /// Report staging an interrupted `kin init` left on disk, and where it stopped.
@@ -7975,6 +8006,37 @@ mod tests {
                 .is_some_and(|fix| fix.contains("new interactive shell")),
             "the no-hook fix is still the shell, and it is still the right one: {stale:?}"
         );
+    }
+
+    /// Outside a Kin repository there is no graph-backed root for a live hook
+    /// to project. That is expected context, not a stale install.
+    #[test]
+    fn a_live_hook_outside_a_repository_is_not_an_attention_row() {
+        let mut unbound = correctly_hooked_report();
+        unbound.live.unengaged_here_only = false;
+
+        let outside = projection_mode_check_for_context(
+            &unbound,
+            "macos",
+            &not_probed(),
+            &hook_live(),
+            false,
+        );
+        assert!(
+            matches!(outside.status, HealthStatus::Unsupported),
+            "outside a repository no projection can be in force, got {:?}: {}",
+            outside.status,
+            outside.detail
+        );
+        assert!(!needs_attention(&outside));
+        assert!(outside.detail.contains("not inside a Kin repository"));
+
+        // The falsification control: the same probe from inside a repository
+        // is still stale because its bound root is not serving this process.
+        let inside =
+            projection_mode_check_for_context(&unbound, "macos", &not_probed(), &hook_live(), true);
+        assert!(matches!(inside.status, HealthStatus::Stale));
+        assert!(needs_attention(&inside));
     }
 
     /// The negative controls for the green above. Each one removes a single
--- a/crates/kin-cli/src/commands/setup.rs
+++ b/crates/kin-cli/src/commands/setup.rs
@@ -13196,7 +13196,7 @@ fn readiness_line(report: &crate::commands::health::HealthReport) -> ReadinessLi
     } else {
         "Each row above carries the fix it needs."
     };
-    let sentence = if waiting.len() == 1 {
+    let mut sentence = if waiting.len() == 1 {
         format!("1 check needs attention: {}. {advice}", labels.join(", "))
     } else {
         format!(
@@ -13205,6 +13205,28 @@ fn readiness_line(report: &crate::commands::health::HealthReport) -> ReadinessLi
             labels.join(", ")
         )
     };
+    let pending = waiting
+        .iter()
+        .filter(|check| matches!(check.status, HealthStatus::Pending))
+        .map(|check| check.label.as_str())
+        .collect::<Vec<_>>();
+    if !pending.is_empty() {
+        sentence.push_str(&format!(
+            " Pending rows are expected first-run work: {}.",
+            pending.join(", ")
+        ));
+    }
+    let degraded = waiting
+        .iter()
+        .filter(|check| matches!(check.status, HealthStatus::Degraded))
+        .map(|check| check.label.as_str())
+        .collect::<Vec<_>>();
+    if !degraded.is_empty() {
+        sentence.push_str(&format!(
+            " Degraded rows are host limits: {}.",
+            degraded.join(", ")
+        ));
+    }
     ReadinessLine {
         ready: false,
         severe,
@@ -14161,14 +14183,7 @@ pub async fn doctor(fix: bool, install_language_servers: bool, json: bool) -> Re
     println!();
     print_human_report(&after);
 
-    let still_manual: Vec<&crate::commands::health::HealthCheck> = after
-        .checks
-        .iter()
-        .filter(|c| {
-            !matches!(c.status, crate::commands::health::HealthStatus::Healthy)
-                && c.manual_fix.is_some()
-        })
-        .collect();
+    let still_manual = manual_attention_checks(&after);
     if !still_manual.is_empty() {
         println!();
         println!("Still needs manual steps:");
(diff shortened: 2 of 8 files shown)


troyjr4103:
Exact-head review on 6c530a366d1dd34182f2a1416929c806ec0a0ca7 by the launch captain (captain-0904-124805) after the opus reviewer died at a usage limit mid-review: **Verdict: GO**. Report: `.kin-coord/reports/showhn-review-1485-20260904.md` in the umbrella (the partial opus notes are kept beside it and agree).

Checked against the diff: eight files, Cargo.lock and Cargo.toml untouched. Each of the six classes does what the body says and no more: the prose-locate gate refuses to certify relevance without ever claiming absence and keeps a named-symbol page certified as its control; the belt keep lists now define every property the served anyOf requires, guarded by two sweeps with anti-vacuity floors; `_kin.freshness` carries the selected graph's own stale observation and the stdio contract rejects disagreement in either direction; every impact row carries `covering_tests_bound: graph_observed_lower_bound` through one seam with shared constants; the projection row downgrades to not-applicable only outside a repository with a live hook, with the inside-a-repository control still stale; the budget test holds both caveats across a 120-row cut. `clipped_steps` confirmed as the real field (trace_data_flow.rs:570); `truncated_steps` survives only in the new test's doc comment. Fourteen falsification pairings read; the survivor explanation holds by inspection. No em dash in any added user-facing string.

One MEDIUM follow-up, not a blocker: `graph_freshness_reading` (verdict.rs:796) is deliberately Silent for every freshness state, a decision written for the admission clock, so the new `Stale` observation reaches the freshness block and the `stale` payload but not `_kin.verdict`; a replayed status can still read `certified` when completeness is complete. Ticket filed.

Hosted checks at 17:30Z: 47 of 47 listed, none adverse, two still running (Fast gate test shard (1), Analyze (rust)). Landing armed through merge enqueue; it lands when those conclude green.

PR #1486, opened 2026-09-04
Say which half of readiness refused, and how long it waited

troyjr4103:
## Why

Kubernetes records a probe's status code and nothing else. So a daemon that answers 503 on `/readiness` tells an operator exactly one thing: it is not ready.

On 2026-09-04 that was not enough. KinLab promotion run [33886343178](https://github.com/firelock-ai/kinlab/actions/runs/33886343178) rolled a hosted daemon that bound port 4219 at 464 seconds into its start, then failed its rollout 134 seconds later having never reported ready. The pod log carries no line saying whether it was still uninitialized or whether the spine authority was refusing:

```
15:09:28Z     container started
15:17:12.992Z INFO kin_daemon::api: daemon API server listening port=4219
15:19:27.064Z error: deployment "kin-daemon" exceeded its progress deadline
15:19:28Z     Replicas: 1 desired | 1 updated | 1 total | 0 available | 1 unavailable
```

Kubelet aggregates repeated probe events, so the last recorded probe result is 15:14:56Z and nothing on record says what `/readiness` answered after the API bound. The post-mortem could name a mechanism and could not confirm it. This is the line that would have answered it.

## What

The refusal branch of `readiness` logs one line naming which half said no, with two durations:

```
WARN kin_daemon::api: readiness refused initialized=false hosted_spine_ready=true warming=false
     refused_on="initialization" authority_wait_ms=0 spine_gate_wait_ms=0
```

The gate wait has to be recorded where it happens, so `acquire_spine_read_authority` stores it in one relaxed atomic immediately after taking the publication gate and `readiness` reads it back. That is a single store on a path that already awaits a lock.

It is worth having as its own number rather than folded into the total. The periodic hosted refresh passes (`ingest_repo_into_spine_under_publication`, `refresh_all_hosted_cross_repo_edges_with_before_commit_hook`, `refresh_all_cross_repo_edges_under_publication`) hold that same `RwLock` for write, and a tokio `RwLock` is write-preferring, so a readiness probe queued behind them is a specific and checkable explanation rather than a guess. In the run above those passes ran back to back with no idle gap from 15:15:21Z until SIGTERM at 15:19:41Z.

Nothing decides anything on these numbers. They are reported, not consulted, and readiness returns exactly what it returned before.

WARN rather than DEBUG: a hosted daemon that cannot serve is the operator's problem whether or not anyone raised the log level first, and the line stops as soon as the daemon is ready.

## The tests are a pair, deliberately

`a_refused_readiness_says_which_half_refused` captures the formatted tracing output through a `MakeWriter` and fails if the line is missing, stops naming which half refused, or drops the gate wait.

`a_ready_daemon_logs_no_readiness_refusal` fails if a ready daemon logs a refusal at all.

Without the second, an unconditional line would satisfy the first while emitting a warning every few seconds for the life of every healthy pod, burying the one case that matters. Neither test is worth much alone.

One fixture detail, because it bit: `test_state()` opens a real repo and `DaemonState::open` sets `is_initialized` from whether a snapshot loaded, so the fixture arrives ready. The first run of the new test failed on that rather than on the behaviour, asserting a 503 from a state that was already 200. It now sets the flag back by hand with a comment saying why.

## Falsification

CI runs the tests; it does not run these. Each arm breaks one thing, alone, and was checked for **which** assertion fired.

```
ARM A  the tracing line deleted entirely
       a_refused_readiness_says_which_half_refused ... FAILED
       a_ready_daemon_logs_no_readiness_refusal ... ok
       "a 503 from /readiness must say so in the log; Kubernetes keeps only the status code, so an
        unlogged refusal is an unanswerable rollout failure."
       captured output contained no readiness line at all

ARM B  the line emitted on the ready branch too
       a_refused_readiness_says_which_half_refused ... ok
       a_ready_daemon_logs_no_readiness_refusal ... FAILED
       "a ready daemon must not log a refusal. The probe runs every few seconds for the life of the
        pod, so an unconditional line would bury the one case that matters."
       captured: readiness refused initialized=true hosted_spine_ready=true ...

ARM C  the refused_on field dropped
       a_refused_readiness_says_which_half_refused ... FAILED
       "the refusal must name which half refused, so a reader does not have to guess between
        initialization and the spine authority."
       captured: readiness refused initialized=false hosted_spine_ready=true warming=false
                 authority_wait_ms=0 spine_gate_wait_ms=0

ARM D  spine_gate_wait_ms dropped
       a_refused_readiness_says_which_half_refused ... FAILED
       "the refusal must report how long the authority check waited for the publication gate, which
        is the measurement that separates a slow open from starvation behind the refresh passes."
       captured: readiness refused initialized=false hosted_spine_ready=true warming=false
                 refused_on="initialization" authority_wait_ms=0

CONTROL  restored
       test result: ok. 6 passed; 0 failed
```

Each arm was confirmed to have compiled before its red was believed (`error[E` and `could not compile` both zero, one test binary executed). An earlier pair of arms edited by regex over the braced block did **not** compile, and `cargo test` gives a compile failure and an assertion failure the same exit code 101, so they read as red while proving nothing. They were redone with the block matched as an exact literal.

Green on this branch, through the fleet's builder slot:

```
$ cargo test -p kin-daemon readiness
test api::tests::a_ready_daemon_logs_no_readiness_refusal ... ok
test api::tests::a_refused_readiness_says_which_half_refused ... ok
test api::tests::readiness_returns_200_when_initialized ... ok
test api::tests::readiness_returns_200_for_initialized_empty_repository ... ok
test api::tests::readiness_reports_the_warming_state_it_is_in ... ok
test api::tests::readiness_is_answered_while_a_blocking_warm_up_holds_the_runtime ... ok
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 1367 filtered out

$ cargo fmt --all -- --check
(clean)
```

FIR-3167, FIR-3179.

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -3880,11 +3880,13 @@ async fn readiness(State(state): State<Arc<DaemonState>>) -> impl IntoResponse {
     let initialized = state
         .is_initialized
         .load(std::sync::atomic::Ordering::Relaxed);
+    let authority_started = Instant::now();
     let hosted_spine_ready = if initialized && state.hosted_spine_readiness_required() {
         state.acquire_spine_read_authority().await.is_some()
     } else {
         true
     };
+    let authority_wait = authority_started.elapsed();
     let warming = state.spine_warming() || (initialized && !hosted_spine_ready);
 
     if initialized && hosted_spine_ready {
@@ -3896,6 +3898,34 @@ async fn readiness(State(state): State<Arc<DaemonState>>) -> impl IntoResponse {
             }),
         )
     } else {
+        // Say which half refused, and how long the authority check spent
+        // waiting, because a refusal that says nothing is how a rollout gets
+        // read as a broken daemon.
+        //
+        // On 2026-09-04 a hosted daemon bound port 4219 at 464 s into its
+        // start, then failed its rollout 134 s later having never reported
+        // ready. Nothing in that pod's log said whether it was still
+        // uninitialized or whether the spine authority was refusing, and
+        // Kubernetes records only the status code, so the post-mortem could
+        // name a mechanism and not confirm it. This is the line that would
+        // have answered it.
+        //
+        // WARN rather than DEBUG on purpose: a hosted daemon that cannot
+        // serve is the operator's problem whether or not anyone raised the
+        // log level first, and this stops as soon as the daemon is ready.
+        tracing::warn!(
+            initialized,
+            hosted_spine_ready,
+            warming,
+            refused_on = if !initialized {
+                "initialization"
+            } else {
+                "hosted spine authority"
+            },
+            authority_wait_ms = authority_wait.as_millis() as u64,
+            spine_gate_wait_ms = state.last_spine_gate_read_wait().as_millis() as u64,
+            "readiness refused"
+        );
         (
             StatusCode::SERVICE_UNAVAILABLE,
             Json(ReadinessResponse {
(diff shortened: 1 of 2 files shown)


troyjr4103:
Independent exact-head review (review1486, read-only, opus-1m) on 67f22e9796aaaa8e94d0123371342eaa0c9b49f4: **Verdict: GO**. Report: `.kin-coord/reports/showhn-review-1486-20260904.md` in the umbrella. Checked: two files, +190 -0, no response behavior changed; both zero-file-search gates pass at the head under the ci.yml invocations; the relaxed atomic is written at one site and read at one with no branch on it; the WARN fires only on the 503 branch; both new tests pass in hosted CI (run 33900087727: a_refused_readiness_says_which_half_refused 0.079 s, a_ready_daemon_logs_no_readiness_refusal 0.660 s), and the four falsification tails in the body print per-test result lines, so the compile trap the lane caught earlier is not in what it reported; hosted 47 of 47, 30 success 17 skipped, CLEAN; 2 behind main with disjoint api.rs hunks (#1484 at 223-654 and 6197, this PR at 3880-3930 and 42799), no rebase needed. Two MEDIUM findings, not blocking, one follow-up PR before the next release cut (spinefix, rows D and E in its report sections 23 and 24): on the initialization branch both duration fields print measured-looking zeros because the gate was never consulted, and no test covers the hosted branch where the numbers are real; and the WARN is not gated to hosted, so a no-snapshot local first start (kin init then kin daemon) writes about five lines a second at the default filter until the first reconcile. Captain ruling: land now; the follow-up gates the WARN to hosted, labels or omits the durations on the initialization branch, and moves the positive test onto a hosted spine fixture.

PR #1490, opened 2026-09-04
Report unavailable enrichment after language-server installation

troyjr4103:
A language-server install could end with a green refresh acknowledgement even when the running daemon returned `enrichment_available: false`. Setup now requires an explicit `true` before acknowledging enrichment and keeps restart guidance for unavailable, unconfirmed, or failed requests. An install with no running daemon no longer asks for an unnecessary restart.

The wizard and `doctor --fix --install-language-servers` share this path. This does not stop a daemon automatically or change language-server discovery.

Fixes FIR-3189.

Validation used the final one-file working tree, based on `a73168325d6b6808de3158ffaf6a0dfebca7771b`, before committing. The verified `setup.rs` SHA-256 is `a9f4b16a83370f7d2e3b0df893867a144492a3c372beeafd026f0a933a17471c`.

```text
.kin-coord/bin/heavy-slot.sh doctorready kin -- cargo test --locked -p kin-cli --lib commands::setup::tests -- --test-threads=1

test result: ok. 191 passed; 0 failed; 0 ignored; 0 measured; 1956 filtered out; finished in 17.74s
heavy-slot: doctorready command exit rc=0 at 2026-09-04T20:33:30Z

.kin-coord/bin/heavy-slot.sh doctorready kin -- /Users/troyfortinjr/GitHub/kin-ecosystem/bin/kin-precheck kin --repo-dir kin=/Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/doctorready-kin
kin-precheck: 21 gates ran, 21 passed, 0 failed, on kin a73168325d6b6808de3158ffaf6a0dfebca7771b DIRTY
heavy-slot: doctorready command exit rc=0 at 2026-09-04T20:29:42Z
```

Precheck ran through the lane wrapper with its explicit worktree path. It included the workflow's all-targets/all-features clippy command and allow-list. Its declared omissions were runner language-server provisioning, GitHub-event release-policy routing, and hydration falsification against a workflow-built poisoned copy.

The two new tests use isolated HTTP response fixtures and an absent-daemon repository. They cover available, unavailable, missing/malformed acknowledgement, HTTP failure, and no-daemon states. Both tests were falsified by restoring the defect they guard:

```text
HTTP-success mutant:
  left: SweepRequested
 right: Unavailable
test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 2145 filtered out; finished in 0.03s

No-daemon restart mutant:
assertion failed: observed.message().is_none()
test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 2145 filtered out; finished in 0.03s

RESTORED fixed setup.rs byte for byte
```

The restored 191-test setup suite passed after both mutation arms. No real Kin daemon or language server was started by the new tests.

### DIFF
--- a/crates/kin-cli/src/commands/setup.rs
+++ b/crates/kin-cli/src/commands/setup.rs
@@ -13297,47 +13297,68 @@ fn fix_verdict(unfinished: &[UnfinishedRepair]) -> Result<()> {
     )
 }
 
-/// Provision the language servers behind Kin's cross-file reference edges, and
-/// report each one in the operator's own words.
-///
-/// Shared by the wizard and by `kin doctor --fix` so a person gets the same
-/// disclosure, the same prompt, and the same restart advice on both paths. The
-/// returned lines are the "what was applied" list both surfaces already print;
-/// anything not applied is printed here instead, because a declined or failed
-/// install is a fact the operator needs and an empty applied-list is not.
-/// Ask a running daemon to re-probe and re-enrich after a server was installed.
-///
-/// Readiness taken once latches: a daemon that probed before the install would
-/// keep reporting that language unavailable for the rest of its life, and that
-/// stale answer is an input under an agent-facing verdict. The sweep route
-/// re-probes at its start, so one call refreshes the verdict and produces the
-/// edges the new server can finally resolve.
-///
-/// Returns whether a daemon was actually poked, so the caller can fall back to
-/// telling the user to restart when there was none.
-async fn refresh_running_daemon_after_install(cwd: &std::path::Path) -> bool {
+/// What a daemon confirmed after a language server was installed.
+#[derive(Debug, PartialEq, Eq)]
+enum PostInstallDaemonRefresh {
+    NoDaemon,
+    SweepRequested,
+    Unavailable,
+    Unconfirmed,
+    Failed,
+}
+
+impl PostInstallDaemonRefresh {
+    fn message(&self) -> Option<String> {
+        let reason = match self {
+            Self::NoDaemon => return None,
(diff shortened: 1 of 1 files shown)


troyjr4103:
Exact-head review evidence for #1490 at 4d678f03b, posted by the launch captain (captain-0904-154852).

Read in full: `setup.rs` only. `refresh_running_daemon_after_install` now returns a typed outcome instead of a bool: a sweep response carrying `enrichment_available: true` is the only case that drops the restart advice; `false` (Unavailable), a missing field (Unconfirmed), a string `"true"` (Unconfirmed, the typed check), a request failure or error body (Failed) all keep `RESTART_AFTER_INSTALL` in the printed line, and no running daemon (NoDaemon) prints nothing new. That is FIR-3189's corrected mechanism: the green tick after `--fix` came from the sweep's success status being read as enrichment, and the row now says why a restart is still needed.

Independent verification on the same head by the Codex readiness captain (`.kin-coord/reports/kin-ready-doctorready-20260904.md`): 191 setup tests green, precheck 21 of 21, two deliberate mutations caught by exact assertions with a sibling control green and the exact bytes restored. The test server cases in the diff cover every enum arm above.

Hosted checks on this head at 20:36Z: 47 listed of 47, 21 success, 17 skipped, 1 neutral, 8 in progress, 0 adverse.

Verdict: GO. Marked ready and enqueued; it lands when the remaining checks conclude green. The release rail is held (release-tag.yml disabled by the captain), so this joins v0.6.8's range and mints only when the composition is proven. Closes FIR-3189.

PR #1492, opened 2026-09-04
Count semantic changes, not every hash a read surface prints

troyjr4103:
Product Acceptance went red on main (run 33912645879, alarm kin#1489) for `kin diff` behaving
correctly. `ref_grammar:short_prefix FAIL FIR-3015 prefix 'adfc' matches 2 changes and kin diff
resolved it anyway, so it picked one without saying which`. It did not. The second "change" was an
entity revision id.

## Why the grader was wrong

`kin diff` resolves its endpoint through `ref_grammar::resolve` (`diff.rs:518-519`), once, and a
test in `ref_grammar.rs` pins that it holds no parser of its own. That resolver already refuses an
ambiguous prefix, and it scans `lease.snapshot().changes`, of which `kin log` walks a subset
(`log.rs:101-150`). So two ids `kin log` printed sharing a prefix would both be seen and the
command would have refused. In that run it exited 0.

The count came from `Suite.changes()`, which collected every `\b[0-9a-f]{64}\b` out of `kin log`
and `kin blame`. `blame.rs:134-137` prints `revision.revision_id` in the first column, and
`EntityRevisionId` is a `Hash256` (kin-model 0.7.26 `ids.rs:265`), so it renders exactly like a
change id. A prefix matching one change and one revision was counted as two changes.

Reproduced locally against a debug `kin` at a73168325, on the same three-commit fixture the suite
seeds:

```
GRADER changes() count=6
REAL change ids from kin log: 3
NOT-A-CHANGE values the grader counted: 3
  dea088e2f946e3840295deb68696b0b011a7f27e269bf5b07efac01d9ad1dfb6
  8e7cdfc3fb62bd65fabe66dbc94743c173258a3d738682a5bbd9f28b6226001c
  dc19e52b30ea6593ad5a9f9f8be50e4573ade1bf0a8ec87c0f9ed24112a6ca61
  $ kin diff dea0 HEAD -> rc=1
DIFF revision-prefix dea0 -> rc=1
  $ kin diff dea088e2f946e3840295deb68696b0b011a7f27e269bf5b07efac01d9ad1dfb6 HEAD -> rc=1
DIFF full revision id -> rc=1
```

`kin diff` refuses a revision id at four characters and at sixty-four. There was nothing to fix in
the product's answer.

After this change, same probe, fresh fixture:

```
AFTER changes() count=3
naive findall over the SAME blame stdout: 6 values
values the old parse would have counted as changes and this one does not: 3
  REVISION 3b902f48088f3302a87f1aeba867ab77b9b7bd1c255fcde3d1e25a40ffeaf11b
  REVISION 404c860bc72d0d5edb8a2af1ffd223dd341670f58606e326204235c7f59457ba
  REVISION 703b4dc4a180e03d25a7b1bf1a5b198933dc50442151fe711827ee41ad7144bd
```

## What changed

`scripts/acceptance/ref_grammar_repro.py` reads each surface where it says what a value IS.
`kin log -n 20 --json` names `change_id`; on a blame row the column decides rather than the shape.
stdout only, because `answer()` folds stderr in and stderr carries tracing warnings. Both surfaces
stay, so a parse keyed on one still cannot pass the day that surface changes how it prints.

`crates/kin-cli/src/commands/ref_grammar.rs` gains the tests the ambiguity arm never had.
`resolve_change_prefix` takes an iterator of change ids rather than the authority lease, because
change ids are hashes and a test that seeds a repository and waits for two of them to collide on
four hexadecimal characters runs about once in tens of thousands of runs. Two behaviours change,
both in the refusal and neither in what resolves: collecting stopped at the preview width BEFORE
the sort, so which candidates a refusal named came out of HashMap iteration order and one command
could print two different failures; every match is collected now, which also makes the count exact
rather than a floor, with the remainder reported.

The docstrings that licensed the old parse are corrected. They put three change ids colliding on
four hexadecimal characters at one run in five. It is nearer one in tens of thousands, so the
fixture reaches the unique arm and the ambiguity refusal is graded where a collision can be built.

## Falsification

Each arm was applied, built and run, then reverted.

Resolve the first candidate instead of refusing:

```
test commands::ref_grammar::tests::a_unique_short_prefix_still_resolves ... ok
test commands::ref_grammar::tests::an_ambiguous_short_prefix_is_refused_naming_the_selector_and_the_count ... FAILED
test commands::ref_grammar::tests::an_ambiguous_refusal_reads_the_same_whatever_order_the_candidates_arrive_in ... FAILED
test result: FAILED. 6 passed; 2 failed
```

Refuse every prefix, which is the control that the two above are not satisfied by a resolver that
refuses everything:

```
---- commands::ref_grammar::tests::a_unique_short_prefix_still_resolves stdout ----
exactly one change begins with beef: cannot resolve 'beef': FALSIFICATION ARM: refuse every prefix
test result: FAILED. 7 passed; 1 failed
```

Restore the early exit before the sort:

```
---- ...an_ambiguous_refusal_reads_the_same_whatever_order_the_candidates_arrive_in stdout ----
the count is exact, not a floor: cannot resolve 'adfc': 'adfc' is ambiguous: 5 semantic changes
begin with it (adfc00000000, adfc00000000, adfc00000000, adfc00000000, and 1 more); use more characters
test result: FAILED. 7 passed; 1 failed
```

Restore `HEX64.findall` over the whole blame row, which is the parse this replaces:

```
SELFTEST change_set/names-changes-not-revisions BROKEN
  expected=['8ebcb1c6...', '29872726...', 'c4e634b2...']
  got=['8ebcb1c6...', '29872726...', 'c4e634b2...', '57b4b328...', '7f051ba6...', '1a9b0a62...']
SELFTEST change_set/blame-alone-is-a-backstop BROKEN
SELFTEST FIR-3015 30 case(s), 2 broken
rc=1
```

Restored, both green:

```
running 8 tests
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 2141 filtered out

SELFTEST FIR-3015 30 case(s), 0 broken
```

Live suite against locally built `kin` and `kin-daemon`:

```
CHECK printed_id FIR-3015 PASS FIR-3015 kin diff resolves the id kin history printed (b23c13f8c390) and still refuses a fabricated one, naming it (deadbeefcafe)
CHECK relative FIR-3015 PASS FIR-3015 kin diff walks parent hops (HEAD~1 answers) and refuses a hop past the root (HEAD~99)
CHECK parity FIR-3015 PASS FIR-3015 both surfaces agree on all 12 measured form(s): 10 resolved by each, 2 refused by each
CHECK refusal_names_it FIR-3015 PASS FIR-3015 the refusal names the endpoint that failed: dededede...
CHECK short_prefix FIR-3015 PASS FIR-3015 the unique 4-character prefix 'b23c' resolves
```

FIR-3015.

### DIFF
--- a/crates/kin-cli/src/commands/ref_grammar.rs
+++ b/crates/kin-cli/src/commands/ref_grammar.rs
@@ -268,35 +268,42 @@ pub(crate) fn parse_change_id(input: &str) -> Result<SemanticChangeId> {
 
 /// The one change whose id starts with `prefix`, or a refusal that says why not.
 ///
-/// The candidate set is the authority snapshot's own change map, which both
-/// callers already hold a lease on. That map IS the repository's history, so a
-/// prefix is matched against every change kin holds rather than against the
-/// slice reachable from one branch: an id an operator read out of
+/// `candidates` is the authority snapshot's own change map, which both callers
+/// already hold a lease on. That map IS the repository's history, so a prefix is
+/// matched against every change kin holds rather than against the slice
+/// reachable from one branch: an id an operator read out of
 /// `kin history --ref <branch>` resolves here whatever branch they are standing
 /// on. Nothing on disk is consulted; the ids come from graph-owned authority.
+///
+/// It arrives as an iterator of ids rather than as the lease itself so that the
+/// refusal below can be graded on a collision that was CONSTRUCTED. Change ids
+/// are hashes, so two of a small repository's own ids sharing four hexadecimal
+/// characters is a one-in-tens-of-thousands accident, and a check that waits for
+/// one is a check that never runs. Taking the ids alone lets the tests below
+/// hand this two that collide by construction, which is the only reason the
+/// ambiguity arm is graded at all.
 ///
 /// Ambiguity is refused rather than resolved to whichever candidate came first.
 /// Picking one would be the worst outcome available: it answers, it looks right,
 /// and it silently describes a different point in history than the operator
 /// meant.
 fn resolve_change_prefix(
-    lease: &kin_db::AuthorityReadLease<kin_db::RepositoryAuthorityState>,
+    candidates: impl Iterator<Item = SemanticChangeId>,
     original: &str,
     prefix: &str,
 ) -> Result<SemanticChangeId> {
-    let mut matches = Vec::new();
-    for change_id in lease.snapshot().changes.keys() {
-        if change_id.to_string().starts_with(prefix) {
-            matches.push(*change_id);
-            if matches.len() > AMBIGUITY_PREVIEW {
-                break;
-            }
-        }
-    }
-    // Sorted so an ambiguity refusal names the same candidates in the same order
-    // every run. A HashMap's iteration order is not stable and an error message
-    // that reshuffles between two runs of one command reads like two different
-    // failures.
+    // Every match is collected before any is named, and the sort is what makes
+    // the refusal reproducible. The early exit this used to take once
+    // `AMBIGUITY_PREVIEW` candidates were in hand ran BEFORE that sort, so which
+    // ids a refusal named came from the change map's iteration order, which a
+    // HashMap does not promise: two runs of one command could name two different
+    // sets and read like two different failures. The filter is a string prefix
+    // over a set bounded by the repository's own history, so collecting it whole
+    // costs nothing worth that ambiguity, and it makes the count an exact number
+    // rather than a floor.
+    let mut matches: Vec<SemanticChangeId> = candidates
+        .filter(|change_id| change_id.to_string().starts_with(prefix))
+        .collect();
     matches.sort();
     match matches.len() {
         0 => Err(ref_error(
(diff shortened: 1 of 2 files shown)


troyjr4103:
Exact-head review of 9288f6621e43318962a57953da2a42e31ca239f8 by an independent reviewer (read-only lane, full report in the umbrella at `.kin-coord/reports/showhn-review-1492-20260904.md`).

Verdict: GO.

Verified on the head: the diff is exactly `crates/kin-cli/src/commands/ref_grammar.rs` and `scripts/acceptance/ref_grammar_repro.py`; the two production callers of the shared resolver (`diff.rs:519`, `ref_lookup.rs:191`) keep the authority change-key iterator; a unique prefix and a full-length id resolve as before and an ambiguous prefix is an `Err` naming the selector and the exact count (`ref_grammar.rs:304-336`); candidate order cannot change the refusal text (`ref_grammar.rs:851-880`); the grader feeds only `kin log --json` change ids and blame's CHANGE column into its set (`ref_grammar_repro.py:585-588`) with the control that sampled revision ids are present in the text and absent from the set (lines 881-907), and it keeps its silent-pick detector (lines 343-353); the focused Rust suite passed 8 of 8 through the builder slot, the parser self-test passed 30 matcher cases with 6 negative controls, both zero-file-search guards passed as ci.yml runs them; an independent falsification (ambiguity arm replaced with `Ok(matches[0])`) turned the two ambiguity tests red and the file was restored to a clean tree. Hosted: 47 of 47 check runs listed, none failed, none pending; Product Acceptance is PR-skipped by design and is graded on main after landing.

Findings: none at HIGH, MEDIUM or LOW.

troyjr4103:
Captain delta review of 825a8c4a39a09e0c7170008a430ef8d71eb58b2c over the reviewed head 9288f6621e43318962a57953da2a42e31ca239f8.

The delta is one file, `scripts/acceptance/ref_grammar_repro.py`, 6 insertions and 3 deletions, every changed line a `#` comment (the SELFTEST_BLAME note now says the dashed rule was shortened from 140 to 78 dashes and carries no hash). The Python AST of the script is identical before and after the delta (`ast.dump` equality), so no behaviour and no parse changed, and the lane's self-test re-ran at 30 cases, 0 broken. The independent review's GO on 9288f6621 carries to this head. Enqueued; hosted checks on this head land it when green.

PR #1493, opened 2026-09-04
Report entity change by content, not by artifact

troyjr4103:
## What this fixes

The vcs stranger arm's own top finding on the public v0.6.7 bytes. Entity-level change detection
was artifact-based in `kin conflicts`, `kin diff` and `kin log`, so a two-function change reported
fourteen merge conflicts where four exist and answered `Entities: +1 ~11 -0`. `kin conflicts
--show` printed empty `base -> ours:` and `base -> theirs:` for ten of the fourteen, contradicting
the same command's own list. The stranger's words: "the difference between a merge you can read
and a merge you cannot."

The cause is in the source, not in the model. `reconciler` stamps the whole FILE's blob hash into
every entity's `metadata.extra`, and editing one function moves the byte span of every entity
below it. Every surface that reported a change compared the whole `Entity`, so every entity in a
touched file compared unequal.

## One filter, not three

kin already had two content filters and neither was reachable from the surfaces that report
changes. `kin blame` filtered on `behavior_hash` in `ref_lookup.rs` and named what it filtered.
The merge projection path had a second, richer one in `entities_agree`, whose own doc comment
names this bug ("twenty-two entity conflicts for one edited function"). `kin conflicts`, `kin diff`
and `kin log` used neither.

This promotes ONE predicate, `entity_content_agrees`, into `kin-core::workspace_semantics`, the
lowest crate `kin-cli` and `kin-daemon` both already depend on. Blame, history, diff, log and merge
composition all call it. `entities_agree` becomes a one-line delegate.

## What did not change

Nothing minted changes. `diff_workspace_semantics` still authors deltas on exact equality, because
the revisions a file-level edit records are real: they are what the file did. What changed is
reporting one of them as a change to an entity that did not change, and every surface now names
what it withheld, which is the contract `kin blame` already held itself to.

The merge decision table is unchanged. Only how it measures "did this side move" becomes
content-based, so an entity neither branch edited composes instead of conflicting and carries this
branch's value, exactly as settling it `--ours` did before. Artifact conflicts are untouched, so a
file both branches changed is still a decision the reader makes. `apply_resolution`, the projection
path and every refusal are untouched, including the FIR-3171 gate that refuses a commit while a
merge transaction is open.

## Fourteen to four, on the stranger's own shape

`a_two_function_merge_conflicts_two_entities_and_two_artifacts` rebuilds the stranger's merge:
`ledger/reporting.py` with 5 entities, `tests/test_reporting.py` with 7, both files changed on both
branches, `format_totals` and `test_search_requires_every_term` edited differently on the two sides.
Every other entity carries the same source on all three sides and differs only in the file blob
stamp and the byte span.

Before, with exact `Entity` equality in `compose`'s decision table:

```
assertion `left == right` failed: only the two edited functions conflict, got
["test_search_requires_every_term", "test_totals_by_category", "test_format_totals",
 "test_search_matches_description", "totals_by_category", "reporting", "ENTRIES",
 "format_totals", "search_entries", "totals_by_month", "test_reporting", "test_totals_by_month"]
  left: 12
 right: 2
```

Twelve entity conflicts plus two artifact conflicts is fourteen, which is the stranger's list
verbatim. After:

```
running 15 tests
test repository_merge::tests::a_two_function_merge_conflicts_two_entities_and_two_artifacts ... ok
test repository_merge::tests::an_entity_edited_differently_on_both_sides_still_conflicts ... ok
test repository_merge::tests::a_one_sided_edit_composes_to_the_side_that_edited_it ... ok
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 1361 filtered out
```

Four conflicts: `format_totals`, `test_search_requires_every_term`, and both files.

## Every new test falsified

Each was broken, run red, restored and run green. The tails are the actual output.

**`compose`'s content-based decision table.** Broken by putting `our_side == their_side`,
`our_side == base_side` and `their_side == base_side` back:

```
thread 'repository_merge::tests::a_two_function_merge_conflicts_two_entities_and_two_artifacts'
panicked at crates/kin-daemon/src/repository_merge.rs:3265:9:
assertion `left == right` failed: only the two edited functions conflict, got [...12 names...]
  left: 12
 right: 2
```

Restored:

```
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 1361 filtered out
```

**`kin diff` and `kin log`.** Broken by replacing the content comparison in `diff_entities` with
`if false` and the one in `unchanged_entity_deltas` with `false`:

```
---- commands::diff::tests::a_two_function_commit_reports_the_functions_it_changed stdout ----
assertion `left == right` failed: one function body moved, not eleven
  left: 12
 right: 1

---- commands::log::tests::a_two_function_commit_counts_the_functions_it_changed stdout ----
  left: 0
 right: 10

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 2147 filtered out
```

Restored:

```
running 8 tests
test commands::diff::tests::a_two_function_commit_reports_the_functions_it_changed ... ok
test commands::diff::tests::a_diff_that_withholds_nothing_says_nothing_about_withholding ... ok
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 2139 filtered out

running 2 tests
test commands::log::tests::a_two_function_commit_counts_the_functions_it_changed ... ok
test commands::log::tests::a_change_that_withholds_nothing_says_nothing_about_withholding ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 2147 filtered out
```

**The predicate itself.** `a_file_level_stamp_and_a_shifted_span_are_not_a_change` goes red if the
whole `Entity` is compared. Its positive controls are `an_edited_body_is_a_change` and
`a_rename_and_a_one_sided_absence_are_both_changes`, so a predicate that returned `true`
unconditionally fails:

```
running 3 tests
test workspace_semantics::tests::an_edited_body_is_a_change ... ok
test workspace_semantics::tests::a_file_level_stamp_and_a_shifted_span_are_not_a_change ... ok
test workspace_semantics::tests::a_rename_and_a_one_sided_absence_are_both_changes ... ok
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 825 filtered out
```

Every collapse test is paired with a control that fails if the fix removed detection rather than
noise: an entity both branches edited still conflicts, a one-sided edit composes to the branch that
made it rather than silently to `ours`, and a diff or log that withholds nothing says nothing about
withholding.

## The second and third commits: an untouched entity follows its published file

**Said plainly, because it matters for reading this PR.** On `origin/main` before this branch,
`compose` used exact `==` and `entities_agree` had exactly one caller,
`decisions_a_side_drops`. So THIS PR is what moved entity composition from whole value to content,
and THIS PR is what opened the stale-span class as a consequence. The alignment step is what
closes it. It is not a pre-existing bug this branch happens to fix; it is the cost of the collapse,
paid in the same PR.

The first commit collapsed the conflict set and left that defect, which review caught and hosted
CI reproduced on its own. Composition decides entities and artifacts on separate passes, so the
two can pick different sides for one file. For an entity BOTH branches edited that is deliberate:
`project_artifacts_from_settled_entities` makes the file follow the settlement. For an entity
NEITHER branch edited it is not. Its content agrees on every side, so it never conflicts and never
gets a settlement, and composition took `ours`. When that file published from `theirs`, the graph
served ours' byte span and ours' file blob stamp for theirs' bytes.

Nothing downstream caught it. The entity delta is empty because content did not move, replay checks
only tree equality, and publish installs the stored delta without re-deriving anything. That is the
stale-span class, arriving through the merge.

`align_unchanged_entities_with_their_artifacts` runs after the artifacts are final, in both paths:
after composition on a clean merge, and after `apply_resolution` and the settlement projection on a
resolved one, because that projection can still flip a file. It is a standalone function called
after those, never edited into them. It moves only entities whose content agrees across the sides,
so an entity a reader settled keeps the side they chose.

**The regression fixture already existed.** `a_bulk_artifact_settle_does_not_override_a_named_entity_settle`
in `crates/kin-cli/tests/repository_merge_resolution.rs:1384` failed on the first head with "the
published graph span 27..43 does not select `mate` in the merged 40-byte artifact", which is this
defect end to end through a real daemon. On this head:

```
     Running tests/repository_merge_resolution.rs
test a_bulk_artifact_settle_does_not_override_a_named_entity_settle ... ok
test result: ok. 26 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 24.31s
```

**Falsified** by making `align_unchanged_entities_with_their_artifacts` return early. The tail is
the defect itself, ours' offsets and ours' blob stamp served for theirs' bytes:

```
---- an_entity_neither_branch_edited_follows_the_side_its_file_publishes stdout ----
assertion `left == right` failed: the entity publishes the span of the file that publishes
  left: Some(SourceSpan { ... start_byte: 3200, end_byte: 3240, start_line: 32 ... })
 right: Some(SourceSpan { ... start_byte: 4800, end_byte: 4840, start_line: 48 ... })

---- an_artifact_flipped_by_a_settlement_carries_its_unchanged_entities stdout ----
assertion `left == right` failed: an entity nobody edited follows the flip its file took
  left:  ... metadata: EntityMetadata { extra: {"artifact_blob": String("20")} } ...
 right:  ... metadata: EntityMetadata { extra: {"artifact_blob": String("30")} } ...

test result: FAILED. 16 passed; 2 failed; 0 ignored; 0 measured; 1361 filtered out
```

Restored:

```
test repository_merge::tests::an_entity_neither_branch_edited_follows_the_side_its_file_publishes ... ok
test repository_merge::tests::an_artifact_flipped_by_a_settlement_carries_its_unchanged_entities ... ok
test repository_merge::tests::alignment_leaves_a_settled_entity_on_the_side_it_was_settled_to ... ok
test result: ok. 18 passed; 0 failed; 0 ignored; 0 measured; 1361 filtered out
```

The third is the control: a pass that rewrote every entity from the published side would satisfy
the first two while discarding the reader's own `--ours` decision.

### The third commit closed two more sides of the same hole

Review found that the alignment only ever moved an entity to `theirs`, and joined an entity to its
file by the path the entity carried before the merge. Two shapes escaped it. A file settled
`--base` published base's bytes under ours' span and ours' blob stamp. A file one side MOVED kept
ours' span over the moved bytes, because the pre-merge path is not in the merged tree and the join
found nothing.

It now takes whichever side the merged tree's value actually came from, and resolves the file by
the identity the merged tree carries. Every side's own span is read, base included and on the same
footing, because the side that moved the file names a different path there. **A path more than one
artifact claims resolves to none:** a contested path is a conflict of its own with its own
settlement, and choosing a claimant here would answer it silently. The join is byte-keyed off maps
built once per side rather than a scan that rendered both sides of every comparison to a `String`
for every entity.

Falsified, base side:

```
---- a_file_settled_to_base_carries_the_base_spans_of_what_nobody_edited stdout ----
assertion `left == right` failed: an entity nobody edited follows its file to base
  left:  ... start_byte: 3200 ... metadata: EntityMetadata { extra: {"artifact_blob": String("20")} }
 right:  ... start_byte: 1600 ... metadata: EntityMetadata { extra: {"artifact_blob": String("10")} }
```

Falsified, identity join, by matching the entity's pre-merge span path against the merged tree:

```
---- a_file_moved_on_one_side_carries_its_moved_spans stdout ----
assertion `left == right` failed: an entity nobody edited follows the file that moved
  left:  ... span: SourceSpan { file: FilePathId("ledger/reporting.py"), start_byte: 3200 ...
 right:  ... span: SourceSpan { file: FilePathId("ledger/reports/reporting.py"), start_byte: 4800 ...
```

### The clean-merge call site, end to end

The alignment has two call sites and only the settled one had end-to-end coverage. It now has both.
`merge_of_a_one_sided_edit_publishes_its_siblings_at_the_merged_offsets` is the most ordinary merge
there is: one branch enlarges one function, the other never touches that file. Every sibling below
the edit has to publish the offsets of the bytes the merge published.

Falsified by removing the `three_way` call:

```
---- merge_of_a_one_sided_edit_publishes_its_siblings_at_the_merged_offsets stdout ----
assertion `left == right` failed: beta publishes the offsets of the bytes the merge published,
not ours' 19 against theirs' 69
  left: 19
 right: 69
```

**Two warnings for whoever checks this.** First, force a daemon rebuild before trusting any
falsification of these integration tests. The harness reuses a compatible `kin-daemon` binary keyed
on build sha and dirty flag, so a falsified source with an unchanged dirty stamp reads as green;
run `cargo build -p kin-daemon --bin kin-daemon` first. That is how the arm above read green until
I forced it. Second, a fixture built on `git mv` proves nothing here: a move imports as a removal
plus an addition, so those entities are new identities and compose correctly without the alignment.
My first attempt at this fixture used one and stayed green under the removed call. The edit case is
the one with teeth.

### Ordering against #1498

`#1498` added the authored-file replay after `project_artifacts_from_settled_entities`, which is
where the alignment also goes. The rebase keeps both with **#1498's replay first**, because
`replay_authored_files` rewrites the merged entities and the merged tree, and the alignment reads
that tree to decide which side each untouched entity follows. Running it first would decide against
a tree that is about to move. #1498's unconditional replay-equality check is untouched.

## The capabilities merge row

`crates/kin-cli/tests/fixtures/git-replacement-capabilities-v1.json` is included into
`kin capabilities`, and its merge note said composition decides each identity by whole value. This
PR makes that false for entities, so the note now names the seven fields content means and says the
byte span and the containing file's blob hash are projection facts, so two sides differing only in
those no longer conflict and the entity composes from the side its file publishes from. The
whole-artifact and no-textual-merge limits are kept verbatim. The inherited semantic-overlay
warning is kept, because the source at this head still emits it:
`crates/kin-daemon/src/repository_merge.rs:324` computes `overlay_pending` and lines 330 to 358
build the refusal naming "a pending semantic overlay". The `conflicts` row is untouched; its prose
describes the record's contents, not conflict granularity.

```
     Running tests/capability_evidence_tests.rs
test result: ok. 2 passed; 0 failed
     Running tests/command_capabilities.rs
test result: ok. 16 passed; 0 failed
```

## Where the checks live

The classes here are all expressible as crate tests, which is where they belong, because kin's
`Product Acceptance` job carries `if: ${{ github.event_name != 'pull_request' }}` and so does not
grade this PR. These run through `ci.yml` and would fail the PR that reintroduced the over-report.

## Local gates

Run from the umbrella root against this worktree, through the fleet's heavy slot:

```
PASS     clippy                               cargo clippy --all-targets --all-features -- <45 allows>
... 21 rows ...
kin-precheck: 21 gates ran, 21 passed, 0 failed, on kin 116e0034cfc56995265e97a65205d7753da63351
```

`kin-parity` was not run; hosted CI is the gate of record for the full workspace.

One thing worth saying plainly: the precheck read 21 of 21 green on the first head while
`repository_merge_resolution` was red, because that gate list is `ci.yml`'s check job and does not
cover that test binary. A lane touching merge composition should run the integration test binaries
too, not only precheck.

## One test run I cannot account for

An early run of the combined library suites reported `test result: FAILED. 2149 passed; 1 failed;
1 ignored`. I cannot name the test: my own script piped that run through `grep "test result:"`, so
the failure line and the failures block were discarded before I saw them. Re-running the identical
tree came back fully green, 2150 passed on kin-cli, 1378 on kin-daemon, 827 on kin-core, zero
failures. So this is a non-reproduction, not a diagnosis. I am not claiming it was pre-existing and
I am not claiming it was flaky infrastructure. I am saying it happened once, it did not happen
again on the same code, and I have no evidence about what it was.

## Local gates on the rebased head

Every command from the umbrella root through the fleet's heavy slot, against this worktree, `rc`
captured on its own line rather than through a pipe:

```
kin-precheck: 21 gates ran, 21 passed, 0 failed, on kin 116e0034cfc56995265e97a65205d7753da63351

cargo test -p kin-daemon --lib repository_merge          rc=0   20 passed, 0 failed
cargo test -p kin-cli --test repository_merge            rc=0   19 passed, 0 failed
cargo test -p kin-cli --test repository_merge_resolution rc=0   34 passed, 0 failed
cargo test -p kin-cli --test capability_evidence_tests   rc=0    2 passed, 0 failed
cargo test -p kin-cli --test command_capabilities        rc=0   16 passed, 0 failed
```


troyjr4103:
Exact-head review of 20b35a73a03a64d6e7244ad185865d353d414a93 by an independent reviewer (read-only lane, full report in the umbrella at `.kin-coord/reports/showhn-review-1493-20260904.md`).

Verdict: NO-GO on that head.

1. HIGH. In a source-only unrelated edit, a content-equal entity composed from ours (`repository_merge.rs:1909-1910`) while the exact artifact composer selected theirs because ours equalled base (`1911-1914`; artifact predicate `633-649`); `diff_workspace_semantics` emitted no delta for it (`704-710`), replay checked tree equality only (`737-754`), and publish installed the stored delta without re-derivation, so a stale span and metadata could be published for the other side's bytes under a clean envelope. Hosted CI confirmed it on the same head: `kin-cli::repository_merge_resolution::a_bulk_artifact_settle_does_not_override_a_named_entity_settle` failed (published span 27..43 does not select `mate` in the merged 40-byte artifact) in Fast gate test shard (1) and the aggregate Fast gate.

Verified on that head: scope was exactly diff.rs, log.rs, ref_lookup.rs, kin-core workspace_semantics.rs and repository_merge.rs; `apply_resolution` and the post-projection replay call were byte-identical to main.

The lane's new head e0818778d adds an alignment step that moves every content-agreeing entity to the side whose file publishes, in both composition paths, with the failing fixture green by name; that head is under its own exact-head review before any landing.

troyjr4103:
Exact-head review of e0818778d6ebb09446817fceff974823822a4792 (the second head) by an independent reviewer (read-only lane, full report in the umbrella at `.kin-coord/reports/showhn-review-1493c-20260904.md`).

Verdict: NO-GO on this head, on one finding, which is a missing guard rather than a defect. A third head is already being prepared.

1. MEDIUM. `align_unchanged_entities_with_their_artifacts` has two call sites, `repository_merge.rs:687-695` in `three_way` (every merge that composes without a reader) and `:1041-1049` in `publish_resolved_merge` (every merge a reader settled), and only the second is observed by any test. Commenting out only the first leaves every relevant suite green (kin-daemon `repository_merge` 18, kin-cli `repository_merge` 18, kin-cli `repository_merge_resolution` 26), while the control arm (the early return re-applied inside the function, with the reused daemon binary deleted first) turns the end-to-end fixture red with the prior NO-GO text, so the suite is sensitive to the function and the clean-merge call site has no guard. The undefended shape is the plainest instance of the class: edit one function on a branch, leave the target's copy untouched, merge with no reader, and every sibling entity in that file would carry a base-era span and blob stamp over theirs' bytes if the call were dropped by a refactor, a revert, or a bad merge against #1498, which also edits this file. The fix is a kin-cli end-to-end fixture for a clean one-sided merge beside the existing one, falsified by removing the `three_way` call.

2. LOW. The entity-to-artifact join reads ours' span path first (`:2011-2018`), so an entity whose ours-side path is not in the merged tree is skipped in silence; not reachable today because reconcile matches identity within one file, so a file move mints new ids. The comment at `:2009-2010` says base is read for an entity whose file moved, which the chain does not do.

3. LOW. The join is a linear scan with a String allocation per comparison over a BTreeMap that buys only dedup; a byte-keyed map is O(1) and allocates nothing.

4. LOW. `kin diff` and `kin log` name what they withheld but offer no `--all-revisions` equivalent, while `diff.rs:1010-1013` claims blame's contract.

Verified on the head: the delta from 20b35a73a is exactly `repository_merge.rs` (+308) and the fixture's merge row (one line); `apply_resolution`, `project_artifacts_from_settled_entities` and the whole publish tail after the alignment call are byte-identical to origin/main by sha256; hosted 47 of 47 listed equals total, 0 failed, 0 pending at 22:10:50Z, with `Fast gate test shard (1)`, `Fast gate build and tests` and CodeQL all success and mergeStateStatus CLEAN; local through heavy-slot: `repository_merge_resolution` 26 passed with the named fixture green by name, kin-daemon `repository_merge` 18 with the three new tests by name, `capability_evidence_tests` 2, `command_capabilities` 16; both zero-file-search guards pass with negative controls; the capability row replaced one sentence and added four, its seven fields match `workspace_semantics.rs:45-53`, reconcile stamps `blob_hash` into every entity of a changed file, and the overlay warning still emits at `:332, 337, 344, 358`.

Correction to the first review's table: origin/main's compose used `==` and `entities_agree` had one caller (`decisions_a_side_drops`), so this PR is what moved entity composition from whole value to content and opened the stale-span class, and the alignment commit is what closes it. The third head carries the clean-merge fixture, a side-agnostic alignment with a published-identity join, and fixtures for a base artifact resolution and a theirs-only file move; the same reviewer re-pins to it.

troyjr4103:
Independent exact-head review at 116e0034c, read-only, by review1493c (worker-opus). Posted by the captain because the review brief bars the lane from writing to the PR. Full report in the umbrella at `.kin-coord/reports/showhn-review-1493d-20260904.md`.

Verdict GO. Every finding from the two prior reviews is closed, each with a falsification arm that goes red when the fix is removed, and the third head also closed a hole neither prior review reached.

Falsification, three arms, each one edit against an otherwise clean tree and each restored. Base dropped from side selection: `a_file_settled_to_base_carries_the_base_spans_of_what_nobody_edited` red at repository_merge.rs:3789, start_byte 3200 against 1600 and artifact_blob 20 against 10, 19 passed 1 failed. Identity join swapped for the pre-merge span path: `a_file_moved_on_one_side_carries_its_moved_spans` red at :3870, ledger/reporting.py 3200 against ledger/reports/reporting.py 4800, 19 passed 1 failed. `three_way` call site removed, with the kin-daemon binary deleted and rebuilt first because the harness compat probe keys on sha-with-dirty and cannot tell two dirty trees apart: `merge_of_a_one_sided_edit_publishes_its_siblings_at_the_merged_offsets` red at repository_merge.rs:446, left 19 right 69, daemon binary stamped 23:24:22Z after the edit. Each arm reddens exactly its own test and nothing else, so none is a blunt mutation.

Ordering against #1498 verified by reading: `project_artifacts_from_settled_entities` ends :1038, the authored-file replay is :1040-1068 and reassigns `merged_artifacts` at :1067, the alignment is :1074-1083 and reads that map at :1081. Load-bearing, not cosmetic. It also composes the other way: an authored file's bytes match no side, so the not-attributable branch at :2128-2133 leaves those entities alone rather than pulling them back to a side.

Untouched boundaries by sha256, against the rebase base e68435c3b: `apply_resolution` 42e94b26..., `project_artifacts_from_settled_entities` 0de212bd..., and the whole publish tail below the alignment call b91e58ef..., all three identical on both refs.

Local, through the fleet heavy slot against the lane worktree at this sha: kin-precheck 21 gates enumerated, 21 passed, 0 failed; kin-daemon --lib repository_merge 20 passed; kin-cli --test repository_merge 19 passed; --test repository_merge_resolution 34 passed; capability_evidence_tests 2; command_capabilities 16. Both zero-file-search guards pass with their own negative controls, rc captured off a pipeline. Hosted at 23:17:30Z: 47 listed equals total_count, 0 failed, 0 pending, all seven required contexts green, Product Acceptance skipped as expected on a PR, mergeStateStatus CLEAN read twice.

Capability row checked against source: the merge note is byte-identical to the second head's, its delta against the rebase base is one replaced sentence plus four added with ten of eleven kept verbatim, the seven named fields are exactly `entity_content_agrees` at workspace_semantics.rs:45-53, reconcile does stamp the file blob hash into every entity of a changed file at reconciler.rs:816 then :897 matched and :944 new with no condition, and the overlay citation holds at :324 and :330-358. The one caveat raised on the second head is gone: "composes from the side its file publishes from" was stronger than that code, which only moved to theirs; this code takes any of the three, so the sentence is now exactly true.

The PR body's unexplained single test run is honest as written. It names the mechanism that destroyed the evidence, its own grep through the result line, and explicitly declines both pre-existing and flaky. The reviewer did not reproduce it; the reviewer's runs on this head are green and there is no evidence about that run either way.

Two non-blocking observations. The moved-file unit test drives a move that preserves the artifact identity, which the current importer does not emit, and the PR body says so and warns the next reader, so it is defensive against a shape the model allows rather than one reconcile produces; it claims nothing end to end. And `kin diff` and `kin log` still cannot list what they withheld, which their comments now correctly say; worth a follow-up ticket, not a block.

PR #1500, opened 2026-09-04
Name every readiness refusal, and only the ones worth naming

troyjr4103:
## What this fixes

FIR-3188, from review1486's two MEDIUM findings on #1486, plus a third refusal source that the work uncovered and that narrows what #1486 proved.

**The probe is answered before the handler on an unadmitted hosted daemon.** `hosted_reader_admission` exempts only the publication-control routes and `/health`. `/readiness` is not exempt, so on a hosted daemon whose reader is not yet admitted the middleware answers the kubelet's probe with 503 `hosted graph reader admission refused` before `readiness` runs. In that state #1486's line never fired; it fired only for an admitted reader whose spine authority then failed closed. The kubelet keeps the status code and discards the body, which is exactly the shape FIR-3179 is about.

One qualification, from review1500, stated so this body does not overclaim: the refusal was not traceless on a live pod. The middleware's 503 carries `hosted graph reader admission refused: ...` in its body, and the control-plane sidecar in the same pod logs the daemon's 503s from its own health poll, so an operator reading that container's log could see the reason. What was missing was the line on the daemon's side, on the probe route the kubelet's decision keys on, which is what this adds. This was found the honest way: a hosted test that asserted the handler's line captured nothing while a marker emitted from the test body itself was captured, so the subscriber was live and the handler had never run.

**Durations printed as measured zeros on the initialization half** (review finding 1). When `initialized` is false the authority check never runs, and the old line still printed `authority_wait_ms=0 spine_gate_wait_ms=0`, which reads as "the gate was fine" when the gate was never asked.

**The WARN was not gated to hosted** (review finding 2). A local daemon answers 503 for its whole first reconcile while the CLI polls every 200 ms, so a no-snapshot first `kin daemon` wrote about five warnings a second with nothing an operator could act on.

## The change

Three refusal sources, each named in the log, none twice:

```
readiness refused refused_on="reader admission" error=...                      hosted_reader_admission, probe route only
readiness refused ... refused_on="initialization"                               readiness, no durations
readiness refused ... refused_on="hosted spine authority" authority_wait_ms=.. spine_gate_wait_ms=..   readiness
```

- `hosted_reader_admission` logs when, and only when, the refused route is a readiness probe route (`is_readiness_route`: `/readiness`, `/ready`, with the `/v2` prefix stripped like its siblings). A busy unadmitted daemon does not write a line per refused request; the probe arrives every few seconds and stops the moment the reader is admitted.
- `readiness` fires nothing unless `hosted_spine_readiness_required()` is true, carries the timing as `Option<Duration>` that is `Some` only when `acquire_spine_read_authority` ran, and emits two different lines so the initialization half carries no durations at all.

Response bodies and status codes are unchanged everywhere.

## Tests

`hosted_test_state()` builds the daemon as production composes it, the same way `publication_control_api_drives_the_rollout_to_release` does: a `LocalFileBackend` root with a snapshot for each of the five fleet repos, `PublicationControl` over an in-memory store, `open_with_backend_and_publication_control`, a durable spine over a cold `FakeSpineStore`, and an assertion that `hosted_spine_readiness_required()` is true. Not `hosted_in_memory_spine_allowed`, which turns the hosted half off.

- `a_hosted_daemon_refused_by_the_spine_authority_reports_both_waits`: the handler called directly on the fixture (`initialized` true, reader unadmitted, cold spine, so `acquire_spine_read_authority` returns `None`). 503, `refused_on="hosted spine authority"`, both durations present. This is the branch review finding 1 said no test covered.
- `a_hosted_daemon_refused_on_initialization_says_so_with_no_durations`: same fixture with `is_initialized` set back to false. 503, `refused_on="initialization"`, and neither duration field appears.
- `an_unadmitted_hosted_reader_refuses_the_probe_and_says_so`: through the full `router_with_publication_control_auth` with no bearer, as a kubelet sends it. 503, the middleware's `refused_on="reader admission"` line, and no handler line, so one refusal is never reported twice with two reasons.
- `a_local_daemon_still_initializing_logs_no_readiness_refusal`: `test_state()`, `is_initialized` false. 503 and nothing logged.
- `a_ready_daemon_logs_no_readiness_refusal`: unchanged from #1486.

The handler tests call `readiness(State(state))` directly because that is the only way to reach the handler's halves on an unadmitted reader; the middleware path has its own test above.

## Falsification

CI runs the tests; it does not run these. Each arm breaks one thing alone, in a copy restored afterwards, and each was checked for which assertion fired and that the arm compiled (`error[E` and "could not compile" both zero).

```
ARM A  the hosted gate removed (the handler logs on every 503, local included)   compiled: yes
       a_local_daemon_still_initializing_logs_no_readiness_refusal ... FAILED
       "a local daemon answers 503 for its whole first reconcile while the CLI polls it every 200 ms;
        a WARN there is a wall of noise on every first `kin daemon`, not a signal."
       the other four: ok

ARM B  durations printed on the initialization half (authority_wait_ms = 0, spine_gate_wait_ms = 0)   compiled: yes
       a_hosted_daemon_refused_on_initialization_says_so_with_no_durations ... FAILED
       "on the initialization half the authority check never ran, so no duration may be printed:
        a zero there reads as \"the gate was fine\" when the gate was never asked."
       the other four: ok

ARM C  the spine-authority line deleted   compiled: yes
       a_hosted_daemon_refused_by_the_spine_authority_reports_both_waits ... FAILED
       "the refusal must name the spine authority half. Captured output was: \"\""
       the other four: ok

ARM D  the middleware line deleted   compiled: yes
       an_unadmitted_hosted_reader_refuses_the_probe_and_says_so ... FAILED
       "a probe refused by reader admission is a 503 the kubelet cannot explain; the middleware must
        say so, because the handler behind it never runs. Captured output was: \"\""
       the other four: ok   (and that empty capture is the proof the handler does not run on this path)

CONTROL  restored: 5 of 5 ok, worktree back to one modified file
```

Each arm was applied to the formatted file by locating its block structurally, printing the enclosing function's brace balance before the build, so a non-compiling mutant could not pass for a red one.

## What this does not do

It does not shorten the open, and it does not prove the starvation hypothesis (FIR-3179). It makes every 503 a hosted readiness probe can receive say why, in the log the kubelet does not discard. The Firestore endpoint is still not selectable at runtime (FIR-3181), so the fixture here is the in-process seam, not an emulator.

FIR-3188. Related FIR-3179, FIR-3167.

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -1783,6 +1783,13 @@ fn is_process_liveness_route(path: &str) -> bool {
     path == "/health"
 }
 
+/// The routes a kubelet readiness probe is pointed at. Both are served by
+/// `readiness`; `/ready` is the older spelling clients still poll.
+fn is_readiness_route(path: &str) -> bool {
+    let path = path.strip_prefix("/v2").unwrap_or(path);
+    path == "/readiness" || path == "/ready"
+}
+
 /// A hosted daemon may expose the authenticated rollout surface while its
 /// reader admission is absent or moving, but it must not serve graph authority
 /// under an unadmitted image. This keeps first bootstrap and recovery possible
@@ -1801,6 +1808,20 @@ async fn hosted_reader_admission(
     if let Some(control) = state.publication_control.as_ref() {
         if let Err(error) = control.assert_runtime_admitted(kin_db::GraphSnapshot::CURRENT_VERSION)
         {
+            // This answers the kubelet's readiness probe before the readiness
+            // handler ever runs, and the kubelet keeps only the status code.
+            // A hosted daemon whose reader is not yet admitted therefore fails
+            // its rollout with nothing in its log saying so, unless it is said
+            // here. Only the probe route logs, so a busy unadmitted daemon does
+            // not write a line per refused request; the probe arrives every
+            // few seconds and stops the moment the reader is admitted.
+            if is_readiness_route(request.uri().path()) {
+                tracing::warn!(
+                    refused_on = "reader admission",
+                    %error,
+                    "readiness refused"
+                );
+            }
             return (
                 StatusCode::SERVICE_UNAVAILABLE,
                 Json(json!({
@@ -3914,13 +3935,20 @@ async fn readiness(State(state): State<Arc<DaemonState>>) -> impl IntoResponse {
     let initialized = state
         .is_initialized
         .load(std::sync::atomic::Ordering::Relaxed);
-    let authority_started = Instant::now();
-    let hosted_spine_ready = if initialized && state.hosted_spine_readiness_required() {
-        state.acquire_spine_read_authority().await.is_some()
+    let hosted = state.hosted_spine_readiness_required();
+    // `None` means the authority check did not run, which is the case on the
+    // initialization half and on every local daemon. Carrying the timing as an
+    // Option keeps a branch that never consulted the gate from ever printing a
+    // measured-looking zero for it.
+    let mut authority_wait: Option<Duration> = None;
+    let hosted_spine_ready = if initialized && hosted {
+        let started = Instant::now();
+        let ready = state.acquire_spine_read_authority().await.is_some();
+        authority_wait = Some(started.elapsed());
+        ready
     } else {
         true
     };
-    let authority_wait = authority_started.elapsed();
     let warming = state.spine_warming() || (initialized && !hosted_spine_ready);
 
     if initialized && hosted_spine_ready {
@@ -3932,8 +3960,8 @@ async fn readiness(State(state): State<Arc<DaemonState>>) -> impl IntoResponse {
             }),
         )
     } else {
-        // Say which half refused, and how long the authority check spent
-        // waiting, because a refusal that says nothing is how a rollout gets
+        // Say which half refused, and when the authority check ran, how long
+        // it waited, because a refusal that says nothing is how a rollout gets
         // read as a broken daemon.
         //
         // On 2026-09-04 a hosted daemon bound port 4219 at 464 s into its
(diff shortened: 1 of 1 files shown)


troyjr4103:
Independent review at the exact head `b7bc4e764ed99b7934016ad69653990c4a8d7228`, read-only.
Verdict: GO. No blocking finding.

**Ownership.** Four sites, by hunk header: the new predicate `is_readiness_route`, inside
`hosted_reader_admission`, inside `readiness`, and inside `mod tests`. One file, 290 insertions and
44 deletions, no other function touched. Zero File-Search holds by reading and not only by the
gates: the one new production predicate is a string test, the middleware hunk adds a `warn!` inside
the existing `Err` arm and returns the same 503 body, and the `readiness` hunks move a clock and
split one line into two. The only filesystem IO added is the test fixture's, at an explicit test
boundary. `hosted_spine_readiness_required()` is now called on every probe and is an Option check
plus a cfg(test) atomic (`state.rs:6602-6611`), so the probe path pays nothing.

**The falsification reproduces.** The arms were re-run independently in a lane worktree whose
`api.rs` is byte-identical to this head (sha256 `6ddabf58b49abd52731dfc621190790a2cf73d9f944da4753cb06aca1457dabd`),
on a tree that was clean before and after. Each arm reddened its own target, compiled clean, and
left the other four green; the control passed five of five.

```
ARM A target-FAILED=1 compile-errors=0 other-ok=4 | test result: FAILED. 4 passed; 1 failed
ARM B target-FAILED=1 compile-errors=0 other-ok=4 | test result: FAILED. 4 passed; 1 failed
ARM C target-FAILED=1 compile-errors=0 other-ok=4 | test result: FAILED. 4 passed; 1 failed
ARM D target-FAILED=1 compile-errors=0 other-ok=4 | test result: FAILED. 4 passed; 1 failed
CONTROL  test result: ok. 5 passed; 0 failed
```

Arm D's tail, the one that proves the new middleware test can fail:

```
panicked at crates/kin-daemon/src/api.rs:43167:9:
a probe refused by reader admission is a 503 the kubelet cannot explain; the middleware must say
so, because the handler behind it never runs. Captured output was: ""
```

**Hosted CI, re-read over the full set.** `check-runs?per_page=100` at this head: `total_count=47`,
`listed=47`, none uncompleted, none concluded failure, cancelled, timed out or action required. The
seven required contexts on main are green by name: DCO Sign-off, PR text hygiene, cargo-deny,
gitleaks (full history), Fast gate lint and policy, Fast gate build and tests, MCP surface contract.
`Product Acceptance` is skipped, which is what AGENTS.md says it must be on a kin pull request.
CLEAN and MERGEABLE. The base is 12 behind, and no main-side commit since the merge base touches
`readiness`, `hosted_reader_admission`, `is_process_liveness_route`, `last_spine_gate_read_wait`,
`hosted_spine_readiness_required` or `acquire_spine_read_authority`, so the CLEAN is semantic and
not only textual.

**Findings.**

1. MEDIUM, non-blocking. "A busy unadmitted daemon does not write a line per refused request" is
   guarded by nothing. An arm that removes exactly that gate, so the middleware logs on every
   refused route, leaves the whole `kin-daemon` lib suite green: `1373 passed; 0 failed`. There is
   no second defence catching it earlier; no test drives a hosted unadmitted daemon at a non-probe
   route while capturing warnings. The handler's equivalent noise gate IS guarded, and arm A
   reddens when it goes, so the asymmetry is the finding: the "only the ones worth naming" half at
   the middleware can be refactored away with every gate staying green. One test closes it: hit a
   non-probe route on the hosted fixture with an unadmitted reader and assert nothing is logged.
2. LOW. `probe_hosted_readiness`'s doc comment says `/readiness` is exempt from the bearer "see
   `is_publication_control_route`". That function (`api.rs:1776`) matches only
   `/authority/publication-control` and its subtree. The exemption is `is_public_route`
   (`api.rs:1587`), which `daemon_auth` calls at `api.rs:1656`. The same block also states its
   purpose twice.
3. LOW. `hosted_test_state()` leaks two `TempDir`s per call through `Box::leak`, so Drop never runs
   and both directories outlive the test process; three tests use the fixture, so one
   `cargo test -p kin-daemon` leaves six behind. The `EnvVarGuard` is already returned in the tuple
   to keep it alive, and the directories can ride along the same way.
4. LOW, calibration only, no change requested. The body's framing is log-scoped and accurate:
   nothing in the log said why. Worth recording that the reason was already readable on the same
   pod. `/health` is exempt from the admission middleware, and its response carries
   `reader_admitted` and `reader_admission_error` (`api.rs:819-821`, filled at `3853-3857`), with an
   unadmitted reader flipping the top-level status to `attention`. What this PR changes is a
   live-read fact into a logged one, which is the half a post-mortem on a dead pod can use.

PR #1499, opened 2026-09-04
Parse a session's bytes where the session publishes them

troyjr4103:
Closes the live-daemon half of FIR-3200. The restart window it does not close is FIR-3208.

## Measurement, taken before the fix

A one-line Rust fixture is committed, a disposable session workspace is materialized, 17
lines are prepended to the file inside the session dir (the shape `kin exec` produces),
`/reconcile` publishes it, then `/commands/commit` runs, then the published span is read
back out of the graph.

Broken:

```
committed start_line=0
reconcile modified=1 enriched=0 after_reconcile_start_line=0
commit status=200 OK body={"entity_count":0,"relation_count":0,"file_count":1}
after_commit start_line=0
assertion `left == right` failed: the published span must move with the bytes the session changed
  left: 0
 right: 17
```

Read the commit line. It returns **200 with `entity_count: 0` and `file_count: 1`**. The
tree delta is recorded and the semantic delta is not, so a function that now begins on line
17 keeps answering at line 0, and every read surface serves that under an envelope with
nothing to report. That is the brownfield stranger's finding on `psf/requests`, reproduced.

Fixed:

```
committed start_line=0
reconcile modified=1 enriched=1 after_reconcile_start_line=17
commit status=200 OK body={"entity_count":1,"relation_count":0,"file_count":1}
after_commit start_line=17
test result: ok. 1 passed
```

The span moves at the reconcile, before the commit runs at all, and the commit that follows
carries `entity_count: 1`, so the semantic change reaches history rather than being absent
from it.

## Mechanism

Three facts, read from source rather than inferred.

`reconcile_session_workspace` admits a session's exact tree and publishes it. Its summary
carried `semantic_files_enriched: 0` and `semantic_enrichment_failures: 0` as literals, and
those two were the only writers of those fields anywhere in the daemon. The publication
moved bytes and parsed none of them.

`command_commit` forces one complete admission through
`sync_filesystem_with_graph_deferring_tree_publication`.

That admission early-returns on an empty tree delta, and the whole semantic enrichment loop,
including `reconciler.reconcile_file_change`, sits below that return.

The first fact makes the tree delta empty for the third. The publication already made the
working copy current, so the commit's admission finds nothing to move, returns, and the file
whose bytes changed is never offered to a parser.

## Fix

The parse now happens where the bytes were published. `reconcile_session_workspace` calls
`readmit_semantics_for_paths` for the paths its publication moved, under the same gate that
published them, and reports the real counts in the two fields that were hardcoded.

[...]

PR #1496, opened 2026-09-04
Support native commit amend with exact head leases

troyjr4103:
Adds `kin commit --amend` through the CLI, daemon and native transaction planner. Amend records the complete working state as a replacement change while retaining the original immutable change, original author and every ordered parent. It preserves the message unless `--message` is supplied and records the current caller separately as the transaction actor.

The CLI selects an exact head. The daemon validates that selection before admission and again during planning, then publishes with an exact ref lease and the existing atomic workspace transaction. Detached amend advances only its workspace. Unborn workspaces, open merges and conflicting entity leases refuse. Ordinary empty commits still refuse. Path and hunk selection remain unavailable and are disclosed in the capability record.

Focused tests and falsification passed before rebasing onto `1e990a4e46cab5d77a05f8b2dd45ecc3e1f09d76`. Formatting and source guards pass on the rebased tree. Full local precheck, capability tests and parity are pending. This draft is not ready to merge.

Focused local commands and output:

```text
cargo test --locked -p kin-daemon --lib amend -- --test-threads=1
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 1373 filtered out; finished in 24.10s

cargo test --locked -p kin-cli --bin kin commit_amend -- --test-threads=1
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 34 filtered out; finished in 0.01s

cargo test --locked -p kin-cli --test commit_amend -- --test-threads=1
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.22s

cargo test --locked -p kin-daemon --lib repository_commit::tests -- --test-threads=1
test result: ok. 25 passed; 0 failed; 0 ignored; 0 measured; 1357 filtered out; finished in 34.40s

cargo test --locked -p kin-daemon --lib a_second_commit_on_an_unchanged_tree_is_refused_and_records_nothing -- --test-threads=1
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1381 filtered out; finished in 1.24s

cargo test --locked -p kin-cli --lib commands::commit::tests -- --test-threads=1
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 2130 filtered out; finished in 5.68s
```

[...]

PR #1501, opened 2026-09-04
Carry a session's unpaid parse across a restart

troyjr4103:
Closes FIR-3208. Sits on top of #1499, which landed at 23:40:02Z as `f6a29e329` and closed
the live-daemon half of FIR-3200; this branch is rebased onto that squash and carries only
its own two commits.

## What #1499 left open, and why it could not close it there

#1499 makes a session's publication parse the bytes it moves, under the gate that published
them. That fixes a live daemon and nothing more. Measured:

```
a_daemon_reopened_after_a_session_reconcile_answers_at_the_moved_span
assertion failed: a reopened daemon must load the spans the session's bytes hold
  left: 0
 right: 17
```

It fails at the first read after the restart, before the commit is even reached. The reason
is structural rather than incidental. Entities reach durable authority only inside a
semantic change; a session reconcile deliberately publishes the workspace tree with no
history and no ref move, which is what keeps a session's edit pending until someone commits
it; so there is no change for the re-derived semantics to ride, and the next daemon rebuilds
from a history that never carried them. A synchronous `save_snapshot` immediately after the
re-derivation was probed and did not change the result, so the snapshot is not the carrier
either.

Closing it needs something that outlives the daemon and tells the next commit it owes a
parse.

## The record

`crates/kin-daemon/src/semantic_debt.rs`. One entry per path, carrying the body hash the
parse is owed for, persisted under the store root beside the two markers that already serve
this path, `unpublished-enrichment.json` and the LSP-enriched marker.

Bound to a hash, not to a path alone, because a path alone cannot say whether the debt is
still real. The working copy moves on, other writers publish over the same path, a commit
lands. Path plus body answers all three: an entry whose tree entry no longer names that body
was overtaken by a later transition, and the admission that observed that transition already
enriched it. Those settle for free instead of being re-parsed forever.

Written at the publication. Drained by the commit's admission. Settled only when the
commit's transaction reaches authority, because a commit derives its change from the live
graph, so everything the drain put back into that graph is in history by then. Settling at
the drain instead would clear the record for a parse the next crash could still lose.

Also drained once per daemon life by the reconcile loop, before it answers anything. A path
whose bytes reached authority without a parse is a wrong answer this daemon can know about,
and waiting for the next commit would serve the previous parse to every query in between.

### Ingestion IO

Every filesystem call this adds is ingestion IO at an explicit boundary and none of it is a
semantic answer. `semantic_debt::outstanding` reads the record, `record` and `settle` rewrite
it, `settle_all` removes it. The record says which paths are owed a parse; the parse itself
reads graph-owned CAS through the exact tree, never the filesystem, and the host is touched
only by the existing `host_entry_matches_graph` identity guard.

## A hole I put in this myself, and the test that would have caught it

I first put the drain inside the `deltas.is_empty()` arm, mirroring the early return it
replaces. That is wrong, and wrong in the same shape as the original defect.

The commit settles the whole record on success. An admission whose own transition is
non-empty would skip the drain and settle anyway, so one unrelated one-line edit would
discard a parse another file was still owed, and that file would stay stale with the record
claiming it was paid. The admission's own transition says what THIS pass moved; it never says
what an earlier one left owing. Drawing a conclusion from the wrong question is exactly what
the empty-delta early return was doing in the first place.

The drain now runs on both paths, before the empty-transition conclusion. I found this by
reading the settle point rather than from a failure, so
`an_unrelated_edit_does_not_settle_a_debt_nothing_paid` is the test it should have failed on:
publish through a session, restart so the record is the only thing carrying the parse, edit
an unrelated file, commit, and assert the first file's span still moved.

## Refusal, and the one failure that is not one

The two halves of a drain failure get different answers, and the difference is the point.

**Source the parser cannot read is reported, and the commit proceeds.** That is the file's
own state, and retaining last-known-good is the designed answer to it rather than a fault:
the daemon reconciles under `ReconcilePolicy::FallbackToLkg` and an agent mid-edit produces
half-written source constantly. Refusing every commit until the syntax is fixed would leave
a caller unable to record anything at all, which is worse than the stale spans this record
exists to prevent, and it would be a regression against what this seam already did for such
a file. It is reported as a `warn` line naming the paths rather than in the commit response,
because `CommandCommitResponse` has no field for it and widening that wire shape for a log
line is not worth it; say so if you disagree and I will add the field.

**Everything else refuses**, through `DaemonError::SemanticReadmissionFailed`, which
`commit_admission_error` answers as 409 naming the files rather than as a 500. The store is
intact and nothing needs repairing; what the caller has to be told is that the graph will
answer about those files at positions their bytes no longer hold, and a transient failure is
something they can retry. Letting one through would settle the record unpaid and reproduce
the stale spans silently, which is the whole class this record closes.

A parse never retracts membership either way, which is the invariant #1499 cites and this
keeps.

### Follow-up, to be ticketed

`settle_all` clears the whole record when a commit reaches authority, so a permanently
unparseable path is settled without ever being paid. That matches today's behaviour for such
a file exactly, and it is no worse, but the better shape is for the admission to hand the
paid set up to the commit so an unpaid entry survives and the next successful parse pays it.
That is more plumbing than the rest of this PR and it is deliberately not in it. Filed as
FIR-3215.

## What review found on the first head, and what changed

**CodeQL raised three `rust/path-injection` alerts** on the module's three filesystem calls,
all through a `marker_path(root: &Path)` that joined a constant onto a root the caller
passed. Review read them as false positives, and they were, but a PR cannot land red and
there is no suppression idiom in-tree. The module now reads `state.layout.root()` inside
`marker_path` and every function takes `&DaemonState`, which is exactly the shape of the
sibling `unpublished_enrichment_marker_path` in `loop_runner.rs` that raises nothing. The
only path this module ever opens is a constant join under a root the daemon bound at open,
and nothing a caller passes can reach the join. The zero-file-search allowlist entry names
functions, not signatures, so it holds unchanged.

**The startup drain mutated the graph outside the coordination gate.** The tick takes
`coordination_gate` before its own admission, and so do commit and checkout; the
once-per-life drain ran before that lock. It now holds the gate for its own scope. Nothing
else holds it at that point in the iteration, and the drain takes the reconciler lock after
the gate in the same order the tick does.

**Two of the guarantees had no test that could fail.** Review removed the startup drain and
moved `settle_all` into the drain, and nothing went red, because the existing reopen test
called the drain directly. Two tests now stand where those mutations bite.
`a_daemon_start_pays_the_semantic_debt_a_previous_daemon_left` puts a store into the shape a
session publication leaves (exact tree published on its own, nothing parsed, the record
naming what is owed), spawns `run_loop`, and polls the span to 17 under a 30-second deadline;
`run_loop_armed` performs no complete sync of its own at startup, so the once-per-life block
is the only thing that can pay it. `a_drain_without_a_commit_leaves_the_record_standing`
drains after a restart, asserts the record stands, restarts again without a commit, reads
the pre-edit span back, pays it again from the standing record, and only then commits and
asserts the record is settled.

## Falsification

Six mutations, one per guarantee, each applied alone with the tree restored before the next.
What follows for each is the mutation and the run's own output, ending in the assertion line
the run printed; nothing below is paraphrased.

All six from one run at 01:05Z to 01:14Z, log `.kin-coord/reports/commitfresh-1501-falsify.log`, each mutation applied alone and the tree restored before the next; the same six tests pass together on the unmutated tree (`6 passed; 0 failed`, 8.07s).

**M1: drop the `semantic_debt::record` call in `reconcile_session_workspace`.**

```
test api::tests::a_daemon_reopened_after_a_session_reconcile_answers_at_the_moved_span ... FAILED
panicked at crates/kin-daemon/src/api.rs:35183:9:
assertion `left == right` failed: a reopened daemon must load the spans the session's bytes hold
  left: 0
 right: 17
```

**M2: move the drain call back inside the `deltas.is_empty()` arm.**

```
test api::tests::an_unrelated_edit_does_not_settle_a_debt_nothing_paid ... FAILED
panicked at crates/kin-daemon/src/api.rs:35358:9:
assertion `left == right` failed: a commit that settles the record must first pay what it settles
  left: 0
 right: 17
```

**M3: remove the once-per-life drain from `run_loop_armed`.**

```
test loop_runner::tests::a_daemon_start_pays_the_semantic_debt_a_previous_daemon_left ... FAILED
panicked at crates/kin-daemon/src/loop_runner.rs:3962:13:
the loop never re-derived the path a previous daemon left owing; span is Some(0)
test result: FAILED. 0 passed; 1 failed; finished in 32.07s
```

**M4: call `settle_all` inside `drain_semantic_debt`.**

```
test api::tests::a_drain_without_a_commit_leaves_the_record_standing ... FAILED
panicked at crates/kin-daemon/src/api.rs:35251:9:
a drain pays the parse in the derived graph; only a commit may settle the record
```

**M5: partition every failure as not unparseable, so every drain failure refuses.**

```
test api::tests::an_unparseable_file_keeps_its_spans_and_does_not_block_the_commit ... FAILED
panicked at crates/kin-daemon/src/api.rs:36816:9:
assertion `left == right` failed: {"error":"semantic_readmission_failed","message":"the bytes for 1 of these paths are durable authority and their semantics could not be re-derived from them, so the graph still answers about them at the positions their previous bytes held: shifted.rs"}
  left: 409
 right: 200
```

**M6: partition every failure as unparseable, so no drain failure refuses.**

```
test api::tests::a_drain_failure_that_is_not_the_file_refuses_the_commit ... FAILED
panicked at crates/kin-daemon/src/api.rs:35524:9:
assertion `left == right` failed: a drain failure the caller can retry must refuse rather than settle unpaid: {"change_id":"b1cff0fa...","branch":"refs/heads/main","entity_count":1,"relation_count":0,"file_count":1}
  left: 200
 right: 409
```

 Its trigger is
graph-owned rather than contrived, a record naming a body the exact tree still carries that
the CAS can no longer produce, which is the shape of a store whose ingestion CAS was pruned
under it.

## Gates

Speed rule in force: no local precheck. Run through the fleet builder slot on head
`f81a6c962d8984c32f20126b88965bcd0a86068e`:

```
cargo fmt --all -- --check                      fmt check rc=0
cargo clippy --all-targets --all-features -- <45 allows from .github/clippy-allowed-lints.txt>
                                                clippy rc=0
cargo test -p kin-daemon --lib -- semantic_debt reopened_after_a_session unrelated_edit
    drain_without_a_commit unparseable_file_keeps drain_failure_that_is_not
                                                6 passed; 0 failed (8.07s)
cargo test -p kin-daemon --lib semantic         33 passed; 0 failed (restructure alone, 76d1c53f6)
```

The full kin-daemon lib suite last ran green on the previous head at 1395 passed, 0 failed;
hosted CI grades the full workspace on this one. The falsification arms above ran on this
head's tree between the restructure and the commit.

CodeQL: watched on this head until `Analyze (rust)` concluded; its verdict is posted as a
comment once it lands.

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -6138,10 +6138,7 @@ async fn reconcile_session_workspace(
         // between this publication and its commit read the pre-edit span back,
         // and a synchronous snapshot immediately after the re-derivation did not
         // change that. The commit that publishes is what settles the record.
-        crate::semantic_debt::record(
-            state.layout.root(),
-            &crate::semantic_debt::owed_by(observation.deltas()),
-        );
+        crate::semantic_debt::record(&state, &crate::semantic_debt::owed_by(observation.deltas()));
         let readmitted =
             crate::loop_runner::readmit_semantics_for_paths(&state, &published_paths(&observation))
                 .await;
@@ -6436,7 +6433,7 @@ async fn command_commit(
             // is still owed. Settled here rather than at the drain because a
             // crash between the two would otherwise clear the record for work no
             // change carried, which is the loss the record exists to prevent.
-            crate::semantic_debt::settle_all(state.layout.root());
+            crate::semantic_debt::settle_all(&state);
             Ok(response)
         }
         Err(error) => {
@@ -35204,6 +35201,93 @@ mod tests {
         );
     }
 
+    /// FIR-3208. A drain that is not a commit settles nothing.
+    ///
+    /// The record exists because a re-derivation lives in the derived graph and
+    /// dies with the daemon that performed it; only a commit puts it in
+    /// history. So a drain at startup pays the parse and must leave the record
+    /// standing, or the next restart before a commit reads the pre-edit span
+    /// with nothing left to say otherwise.
+    ///
+    /// Falsify by moving `settle_all` into `drain_semantic_debt`: the first
+    /// assertion on the record goes red, and the second reopen reads the
+    /// pre-edit span with no record to pay it from.
+    #[cfg(unix)]
+    #[tokio::test]
+    #[serial_test::serial(commit_phase_capture)]
+    async fn a_drain_without_a_commit_leaves_the_record_standing() {
+        const PREPENDED_LINES: u32 = 17;
+        let repo = tempfile::tempdir().unwrap();
+        let initialized = kin_core::init(repo.path()).unwrap();
+        let layout = initialized.layout.clone();
+        let state = Arc::new(DaemonState::open(initialized.layout).unwrap());
+        state
+            .is_initialized
+            .store(true, std::sync::atomic::Ordering::Relaxed);
+        std::fs::write(
+            repo.path().join("shifted.rs"),
+            b"pub fn shifted() -> u32 { 7 }\n",
+        )
+        .unwrap();
+        let app = router(Arc::clone(&state));
+        commit_through_api(&app, kin_model::OperationId::new(), "publish the fixture").await;
+        let before = published_start_line(&state, "shifted.rs");
+        let session_dir = layout.root().join("runs/session-fir3208-standing");
+        materialize_session_through_api(&app, &session_dir).await;
+        prepend_session_lines(&session_dir.join("shifted.rs"), PREPENDED_LINES);
+        reconcile_session_through_api(&app, &session_dir).await;
+        drop(app);
+        drop(state);
+
+        // First restart: the drain pays the parse, and the record stands.
+        let reopened = Arc::new(DaemonState::open(layout.clone()).unwrap());
+        crate::loop_runner::drain_semantic_debt(&reopened)
+            .await
+            .unwrap();
+        assert_eq!(
+            published_start_line(&reopened, "shifted.rs"),
+            before + PREPENDED_LINES
+        );
+        assert!(
+            !crate::semantic_debt::outstanding(&reopened).is_empty(),
+            "a drain pays the parse in the derived graph; only a commit may settle the record"
+        );
+        drop(reopened);
+
+        // Second restart, still no commit: the derived graph is stale again
+        // and the record is what makes that recoverable.
+        let again = Arc::new(DaemonState::open(layout).unwrap());
+        again
+            .is_initialized
+            .store(true, std::sync::atomic::Ordering::Relaxed);
+        assert_eq!(
+            published_start_line(&again, "shifted.rs"),
+            before,
+            "nothing durable carried the parse across the restart"
+        );
+        crate::loop_runner::drain_semantic_debt(&again)
+            .await
+            .unwrap();
+        assert_eq!(
+            published_start_line(&again, "shifted.rs"),
+            before + PREPENDED_LINES,
+            "the standing record pays it again"
+        );
+
+        // The commit is what settles it.
+        let again_app = router(Arc::clone(&again));
+        commit_through_api(
+            &again_app,
+            kin_model::OperationId::new(),
+            "commit the session's edit",
+        )
+        .await;
+        assert!(
+            crate::semantic_debt::outstanding(&again).is_empty(),
+            "a commit that reached authority settles the whole record"
+        );
+    }
+
     /// FIR-3208. One unrelated edit must not settle a debt nothing has paid.
     ///
     /// The commit that carries an admission clears the whole record once its
@@ -35402,7 +35486,7 @@ mod tests {
         };
         drop(tree);
         crate::semantic_debt::record(
-            state.layout.root(),
+            &state,
             &[crate::semantic_debt::SemanticDebt {
                 path: "shifted.rs".to_string(),
                 body: hash.to_string(),
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -6129,6 +6129,19 @@ async fn reconcile_session_workspace(
         // previous parse recorded. On a converted `psf/requests` an edit that
         // prepended seventeen lines left the whole file answering seventeen
         // lines short, under an envelope with nothing to report.
+        // Record what this publication owes a parser before parsing any of it,
+        // and leave the record standing after. Re-deriving here fixes the live
+        // daemon and nothing more: entities reach durable authority only inside
+        // a semantic change, this publication deliberately writes no history,
+        // and a derived-graph mutation no change carries is gone when the next
+        // daemon replays that history. Measured, not assumed. A daemon reopened
+        // between this publication and its commit read the pre-edit span back,
+        // and a synchronous snapshot immediately after the re-derivation did not
+        // change that. The commit that publishes is what settles the record.
+        crate::semantic_debt::record(
+            state.layout.root(),
+            &crate::semantic_debt::owed_by(observation.deltas()),
+        );
         let readmitted =
             crate::loop_runner::readmit_semantics_for_paths(&state, &published_paths(&observation))
                 .await;
@@ -6167,7 +6180,11 @@ async fn reconcile_session_workspace(
         changes,
     };
     if !semantic_failures.is_empty() {
-        return Err(semantic_readmission_refused(&summary, &semantic_failures));
+        let named = semantic_failures
+            .iter()
+            .map(|failure| failure.path.clone())
+            .collect::<Vec<_>>();
+        return Err(semantic_readmission_refused(&summary, &named));
     }
     Ok(Json(summary))
 }
@@ -6413,6 +6430,13 @@ async fn command_commit(
                 &state.layout,
                 state.graph.resolved_tree().len() as u64,
             );
+            // This transaction reached authority, and a commit derives its
+            // change from the live graph, so every parse the admission above put
+            // back into that graph is in history now. Nothing the record named
+            // is still owed. Settled here rather than at the drain because a
+            // crash between the two would otherwise clear the record for work no
+            // change carried, which is the loss the record exists to prevent.
+            crate::semantic_debt::settle_all(state.layout.root());
             Ok(response)
         }
         Err(error) => {
@@ -6645,10 +6669,22 @@ fn refuse_a_successor_that_records_nothing(
 /// they can see. On 0.5.52 it reached them as `HTTP 500 Internal Server Error:
 /// Core error: ...` (FIR-2664). Everything else stays an internal error.
 fn commit_admission_error(error: crate::error::DaemonError) -> (StatusCode, String) {
-    match projection_blocked_refusal(&error) {
-        Some(refusal) => refusal,
-        None => internal_error(error),
+    if let Some(refusal) = projection_blocked_refusal(&error) {
+        return refusal;
     }
+    // A parse this daemon could not perform over bytes that are already durable
+    // authority. The store is intact and nothing needs repairing, so answering
+    // it as an internal error would send a reader to the daemon for a file they
+    // can see. It is a refusal in words, naming the files, exactly as the
+    // reconcile that first owed the parse refuses.
+    if let crate::error::DaemonError::SemanticReadmissionFailed(message) = &error {
+        let body = serde_json::json!({
+            "error": "semantic_readmission_failed",
+            "message": message,
+        });
+        return (StatusCode::CONFLICT, body.to_string());
+    }
+    internal_error(error)
 }
 
 /// The worded refusal for a blocked projection, when `error` is one.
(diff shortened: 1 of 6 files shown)


troyjr4103:
## Precheck gate table, head `76d1c53`

```
kin-precheck: repo=kin tree=/Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/commitfresh-kin sha=76d1c53f683190fb0932bed6705b4595607f4bc0 branch=chore/lane-commitfresh2-kin

kin-precheck: 21 gates ran, 21 passed, 0 failed, on kin 76d1c53f683190fb0932bed6705b4595607f4bc0
```

`cargo test -p kin-daemon --lib`: 1395 passed, 0 failed, 3 ignored, 276s.

troyjr4103:
## CodeQL on head `f81a6c962`

```
VERDICT Analyze (rust): success
  Analyze (actions): completed/success
  Analyze (c-cpp): completed/success
  Analyze (csharp): completed/success
  Analyze (go): completed/success
  Analyze (javascript-typescript): completed/success
  Analyze (python): completed/success
  Analyze (ruby): completed/success
  Analyze (rust): completed/success
  Build and publish dev daemon image: completed/skipped
  cargo-deny: completed/success
  Check & Test (macos-latest): completed/success
  Check & Test (ubuntu-latest): completed/success
  Check & Test macOS shard: completed/skipped
  Check & Test ubuntu shard: completed/skipped
  Check & Test: completed/skipped
  Check & Test: completed/skipped
  Classify diff scope: completed/success
  Classify diff scope: completed/success
  Code Coverage: completed/skipped
  CodeQL: completed/failure
  DCO sign-off: completed/success
  DCO Sign-off: completed/success
  Falsify guards: completed/skipped
  Fast gate build and tests: completed/success
  Fast gate lint and policy: completed/success
  Fast gate test shard (1): completed/success
  Fast gate test shard (2): completed/success
  Fast gate test shard (3): completed/success
  Feature permutation tests (macos-latest): completed/success
  Feature permutation tests (ubuntu-latest): completed/success
  Feature permutation tests: completed/skipped
  gitleaks (full history): completed/success
  Install Proof (PR) Build: completed/skipped
  Install Proof (PR) Gate: completed/success
  Install Proof (PR): completed/skipped
  Linux Daemon Smoke + MCP Headless: completed/skipped
  MCP surface contract: completed/success
  npm launcher tests: completed/success
  PR text hygiene: completed/success
  PR text hygiene: completed/success
  Product Acceptance: completed/skipped
  Release version gate: completed/skipped
  Windows authority CLI tests: completed/skipped
  Windows authority runtime tests: completed/skipped
  Windows authority tests: completed/skipped
  Windows cross-check: completed/success
  Windows installer + vector release build: completed/skipped
```

Watched from the push at 01:37Z until `Analyze (rust)` concluded, reading the full check-run set each poll (per_page=100, length equal to total_count). The three `rust/path-injection` alerts on `semantic_debt.rs` do not recur on this head.

troyjr4103:
## Independent exact-head review, `f81a6c962d8984c32f20126b88965bcd0a86068e` (and `76d1c53f6` before it)

Read-only review by lane review1501. Nothing here was pushed, enqueued or merged by the reviewer. Full report: `.kin-coord/reports/showhn-review-1501-20260904.md`.

**Verdict: NO-GO on this head, on one ground.** CodeQL concluded `failure` after `Analyze (rust)` concluded `success`: "3 new alerts including 3 high severity security vulnerabilities", `rust/path-injection`, at `crates/kin-daemon/src/semantic_debt.rs:128`, `:201` and `:224`, the module's three `std::fs` calls. The full check-run set reads 47 listed of 47 total, nothing pending, failing exactly `["CodeQL"]`, `mergeStateStatus` BLOCKED, so `merge land` refuses. The same three sites failed on `76d1c53f6` (`:123`, `:196`, `:219`); reading the root from `&DaemonState` inside `marker_path` did not change the taint result. On substance the three read as false positives: the joined component is the constant `semantic-debt.json` under the daemon's own store root, no caller value reaches the join, and the identically shaped `unpublished_enrichment_marker_path` in `loop_runner.rs` is not flagged. The repo already carries two open alerts of this rule on main (`kin-core/src/manifest.rs:174`, `kin-core/src/last_admission.rs:154`); the PR check fails because these are new. Every other clause of the review passes, below.

### The design, read at the head

Every clause of the decided design holds. Debt written at the publication (`api.rs:6141`, under the reconcile's own coordination gate). Drained in the commit's admission on BOTH arms (`loop_runner.rs:8692` on `76d1c53f6`, four lines above the `deltas.is_empty()` return). Drained once per daemon life at startup, now under the coordination gate (`loop_runner.rs:2654` block, gate taken at its head, same lock order as the tick, no re-lock of the non-reentrant mutex). Settled only on the commit's success arm after the transaction reached authority (`api.rs:6439`). `BrokenAst` is the only failure marked `unparseable` (`loop_runner.rs:8417`); it warns naming the paths and the commit proceeds with `settle_all` unchanged. Every other kind (`Conflict`, CAS read, host mismatch, apply failure, reconciler error) refuses through `DaemonError::SemanticReadmissionFailed`, answered as 409 with `error: "semantic_readmission_failed"` and the files named. `repository_commit.rs` is untouched, so #1496's amend path is preserved by construction. One ordering point checked rather than assumed: `exact_tree_admission` applies its tree deltas to the graph before the drain reads it, so a path the commit's own transition moved reads as spent and cannot 409 on a host mismatch.

### The delta from `76d1c53f6` to `f81a6c96`

One commit, three files, 199 insertions and 24 deletions, read hunk by hunk: the restructure (every function in the module takes `&DaemonState`, `marker_path` reads `state.layout.root()`), the gate inside the startup block, two new tests, and FIR-3215 named for the paid-set follow-up. `error.rs`, `lib.rs` and the allowlist are untouched. Nothing else moved. Both zero-file-search gates are green on this head (`Verification PASSED`, 316 files; `70 answer modules are graph-clean`), and the allowlist scope was falsified on the first head: one raw `std::fs::read_to_string` added to `settle`, a non-allowlisted function in the same module, fails the gate with `[VIOLATION] ... Line 174`.

### Falsification, reviewer's own runs on `76d1c53f6`

Baseline: 4 passed in 6.96s. Each arm restored the tree and asserted porcelain empty; `KIN_EMBED_BACKEND=cpu` pinned.

Arm A, drain moved back inside the empty arm: TWO tests red, not one.

    test api::tests::an_unrelated_edit_does_not_settle_a_debt_nothing_paid ... FAILED
    test api::tests::a_drain_failure_that_is_not_the_file_refuses_the_commit ... FAILED
    assertion `left == right` failed: a commit that settles the record must first pay what it settles
      left: 0
     right: 17
    assertion `left == right` failed: a drain failure the caller can retry must refuse rather than settle unpaid: {...}
      left: 200
     right: 409
    test result: FAILED. 2 passed; 2 failed; 0 ignored; 0 measured; 1394 filtered out; finished in 8.61s

Arm B, startup drain block removed, full lib suite: `test result: ok. 1395 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 606.96s`. Nothing red.

Arm C, `settle_all` moved from the commit's success arm into the drain, full lib suite: `test result: ok. 1395 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 344.52s`. Nothing red.

Arm D, `record` dropped at the publication: reopen and unrelated-edit tests red, `assertion \`left == right\` failed: a reopened daemon must load the spans the session's bytes hold / left: 0 / right: 17`.

Arm E, every failure refuses: `an_unparseable_file_keeps_its_spans_and_does_not_block_the_commit` red, `left: 409 / right: 200`, the 409 body naming `shifted.rs`.

Arm F, nothing refuses: `a_drain_failure_that_is_not_the_file_refuses_the_commit` red, `left: 200 / right: 409`.

So the BrokenAst/other split is guarded in both directions by deterministic end-to-end fixtures (the non-BrokenAst trigger deletes the body from the CAS, which is `unresolved`), both NO-GO shapes in the brief are produced on purpose and caught, and the two unguarded lines arms B and C found are what the new head's two tests answer. Measured on the 409 path with temporary assertions: after the refused commit `other.rs in_authority=true in_graph=true control_absent=true debt_record_stands=true`. The refusal keeps the bytes and does not settle.

On `f81a6c96`, arms B and C are re-run against the six tests (four original plus two new) and must go red; that run is queued behind the builder cap and is appended to the report when it lands. The author's own digest log for the same mutations shows `a_daemon_start_pays_the_semantic_debt_a_previous_daemon_left` failing at `loop_runner.rs:3962` after 32.07s and `a_drain_without_a_commit_leaves_the_record_standing` failing at `api.rs:35251`, with line numbers consistent with a tree carrying the new tests.

### Hosted checks

`76d1c53f6`: 54 of 54 listed, seven required contexts green, CodeQL failure. `f81a6c96`: 47 of 47 listed, seven required contexts green by name (`DCO Sign-off`, `PR text hygiene`, `cargo-deny`, `gitleaks (full history)`, `Fast gate lint and policy`, `Fast gate build and tests`, `MCP surface contract`), CodeQL failure after `Analyze (rust)` concluded. On both heads CodeQL first read `completed/neutral` while `Analyze (rust)` was in progress and resolved to `failure` only when it concluded.

### Findings

1. **HIGH, blocks landing.** CodeQL `failure` on the exact head, three new `rust/path-injection` alerts at `semantic_debt.rs:128`, `:201`, `:224`. False positives on substance; a hard gate in process. Routes, in the reviewer's order: dismiss the three in the security tab and re-run the CodeQL workflow on this head (no code change; unverified that this flips the PR check without a new commit); a repo-level `.github/codeql/codeql-config.yml` exclusion for this module (new idiom, policy call); another restructure (not recommended tonight: the query cannot be run on this box and the sibling with the same signature is not flagged, so the mechanism is not the signature).
2. **Resolved on this head.** The startup drain now runs under the coordination gate.
3. **Resolved on this head, pending the reviewer's red run.** Two tests now stand where arms B and C found none.
4. **LOW, residual.** The six M1 to M6 tails are verbatim on every assertion line and trimmed on the framing lines; the paraphrased block at the top of the body (`assertion failed: a reopened daemon ...`) is unchanged. One edit: replace it with the `assertion \`left == right\` failed:` line the body already carries under M1.
5. **Resolved on this head.** FIR-3215 named.
6. **LOW, untouched, not in the captain's list.** Two pre-existing comments are orphaned from the code they document: at `loop_runner.rs:2479-2481` ("Owed once per daemon life, and independent of the catch-up window ...") which now sits five lines and one binding away from `enrichment_repair_owed`, and at `:2646-2653` ("The unpublished-enrichment repair, owed once per daemon life ...") which now heads the drain block instead of the repair block it describes.
7. **LOW, untouched, for FIR-3215.** `sync_filesystem_with_graph_publishing_inner` returns before the drain when `filesystem_reconcile_disabled()` or `is_bare_repository()`, and the commit's success arm then settles a record no drain paid. Both need a mode in which a session reconcile could not have recorded anything, and the outcome is today's stale-span behaviour.

troyjr4103:
Correction to the comment above: `Analyze (rust)` concluding is the analysis running, not the alerts clearing. The `CodeQL` check-run on `f81a6c962` is a **failure** with the same three `rust/path-injection` alerts open at `semantic_debt.rs:128`, `:201` and `:224`, the three `std::fs` sites at their post-restructure lines. Mirroring the sibling marker did not clear them. The alert flow paths are being read now and the next head fixes on that evidence.

troyjr4103:
Resolved. The three `rust/path-injection` alerts (#109, #110, #111) were dismissed as false positives at 01:59Z with the justification that `semantic-debt.json` is a constant filename joined under the daemon's own store root and no caller value reaches the join. For the record, the SARIF flow evidence: every path starts at the axum route registration for `command_stash`, `reconcile_session_workspace` and `command_commit` and reaches the join through `State(state)` and `state.layout.root()`; main already carries two open alerts of the same class in kin-core (`last_admission.rs:154`, the durable-admission marker `command_commit` writes through the same `state.layout`, and `manifest.rs:174`). The `CodeQL` check on `f81a6c962` now reads "No new alerts in code changed by this pull request", and every required context is green.

troyjr4103:
Independent exact-head review of f81a6c962 by lane review1501, final.

Every clause of the decided design holds on this head: the debt is written at the publication, drained in the commit's admission on both arms and once per daemon life at startup, now under the coordination gate, and settled only on the commit's success arm; BrokenAst is reported and the commit proceeds, every other failure refuses through the 409, and a refused commit was measured to keep its bytes and its record. The delta against 76d1c53f is exactly the module restructure plus the gate, the two tests that can fail, the verbatim falsification tails and the FIR-3215 follow-up, and nothing else. On hosted CI, 47 of 47 runs concluded with none failing and all seven required contexts green by name.

CodeQL deserves one honest sentence: the fresh analysis of this head flagged the restructured module again at lines 128, 201 and 224, so the restructure did not change what the tool traces; the three alerts (109, 110, 111) are false positives, a constant filename joined onto the daemon's own store root that no caller string can reach, and the check reads success because they were dismissed as such at 01:59Z, after Analyze (rust) had concluded. The two new tests were shown to fail under their mutations in the author's own logged run (M3 red at 32.07s against its 30-second deadline, M4 red), checked line by line against the head; the reviewer's own re-run of those two arms was queued on the builder cap when the PR landed and was released rather than finished.

Verdict: GO. Landed 02:05:12Z as 49ad3e142.

PR #1503, opened 2026-09-05
Retire a deleted file's entities in the session admission that deletes it

troyjr4103:
Closes FIR-3209, the mirror half of FIR-3200. Sits on `f6a29e329` (#1499's squash) and is
independent of #1501.

## Measurement, taken before the fix

Two Rust fixtures committed, a session workspace materialized, `gone.rs` deleted inside the
session directory, then `POST /reconcile`.

Broken:

```
FIR-3209 measurement: before gone.rs entities=1 keep.rs entities=1
FIR-3209 measurement: reconcile status=500 Internal Server Error after gone.rs entities=1
  body=Core error: repository authority commit outcome is indeterminate: commit repository
  projection authority: storage error: transaction leaves entity 0b2482f5-... on repository
  path gone.rs absent from the staged tree; carry its exact entity removal or relocation in
  the same delta
assertion `left == right` failed: deleting a source file in a session is an ordinary edit and
  must reconcile
  left: 500
 right: 200
```

The file is gone from disk and its entity is still in the graph, and the reconcile that
should have retired it returned 500. The review's independent control probe at 3c28603
measured the same thing (`.kin-coord/reports/review1499-logs/control-head.log`:
`after gone.rs entities=1 host_exists=false`).

Fixed:

```
FIR-3209 measurement: before gone.rs entities=1 keep.rs entities=1
FIR-3209 measurement: reconcile status=200 OK after gone.rs entities=0
  body={"schema":"kin.session-reconcile.v1",...,"removed":1,...}
test result: ok. 1 passed; 0 failed
```

Beyond the two printed lines the test asserts the entity is absent by path, by id and by
name, that the exact tree no longer carries the path, that the sibling file keeps its one
entity, and that the commit after the session records `EntityDelta::Removed` for the
retired entity beside `TreeDelta::Removed` for the file.

```
(assertions above, all green in the same run)
```

## Mechanism

`plan_session_workspace_admission` builds the session's workspace mutation with
`semantic_delta: WorkspaceSemanticDelta::default()`. kin-db refuses a transition that leaves
an entity on a path the staged tree no longer carries, and its message names the remedy:
"carry its exact entity removal or relocation in the same delta". The invariant is right and
the session admission was the caller that owed the removal, so an agent's `rm` of a source
file could not reconcile at all. The existing session-reconcile test deletes `delete.me`, a
file with no entities, which is why the suite was green over this.

## Fix

`plan_session_workspace_admission` now computes the paths the tree transition vacates: a
path some delta's old state held and no delta's new state holds, so a modification keeps its
path and a rename vacates only its old one. For those it reads the workspace's own authority
graph snapshot from the lease, the way checkout does, drops every entity owned by a vacated
path and every relation with such an entity at either end, and derives the canonical
transition with `kin_core::diff_workspace_semantics` against that snapshot. It is planned
against authority rather than the daemon's live graph for the reason every other repository
command plans its authority delta that way: the live graph carries enrichment that has not
crossed the compare-and-swap, and a delta planned from it would ask authority to retract
semantics it never held.

That delta rides the same workspace mutation as the tree transition, where the default sat.
The plan carries the retired entity and relation deltas out, and `reconcile_session_workspace`
applies them to the live graph in the same step as the tree deltas, clears the vacated
file's facets, and emits the `EntityChanged` deletions, so the derived graph follows
authority exactly and never holds an entity authority just refused to carry.

**Renames, stated rather than hidden.** A moved file retires its entities at the old path
here and #1499's readmission re-derives them at the new one. The end state is correct and
the ids are new rather than relocated. The ticket says "removal or relocation"; this is
removal. Relocation with identity preserved is a follow-up, not something to invent late on
a Friday.

## Falsification

Restore `semantic_delta: WorkspaceSemanticDelta::default()` in
`plan_session_workspace_admission`, which is exactly the line the fix replaced:

```
FIR-3209 measurement: before gone.rs entities=1 keep.rs entities=1
FIR-3209 measurement: reconcile status=500 Internal Server Error after gone.rs entities=1
  body=... transaction leaves entity 0b2482f5-... on repository path gone.rs absent from the
  staged tree; carry its exact entity removal or relocation in the same delta
assertion `left == right` failed: deleting a source file in a session is an ordinary edit and
  must reconcile
  left: 500
 right: 200
```

## Second head: what review found at 03476033c, and what changed

review1503 measured two HIGH red arms at the exact first head, both end to end
(`.kin-coord/reports/review1503-probes-ce.log`). Both are in this head as red-first crate
tests lifted from the review's own probes, with assertions added.

**The artifact half.** Deleting a file another file imports still returned 500:
`transaction relation ... has unadmitted destination endpoint artifact:...`. The first head
retired the path's entities and the relations touching them, and nothing else. The
cross-file linker mints artifact-to-artifact `Imports` edges that no entity reaches, and
kin-db validates every relation endpoint against the staged artifact set, so one surviving
edge fails the whole transaction. `VacatedPaths` now carries the vacated artifact ids beside
the paths, and `retire_semantics_on_vacated` also drops every relation bound to a vacated
artifact node, which is the rule `retire_artifact_node_relations` already applies to the live
graph in `loop_runner`. Test: `a_session_that_deletes_an_imported_file_retires_its_artifact_edges`.

**The split brain, which the first head introduced.** Modify an entity-owning file in one
session, delete it in the next: 500 with `stale old payload for removed entity` AFTER the
authority commit was durable. The first reconcile's readmission re-derives the path's
entities into the live graph only, so authority and the live graph hold different payloads
for the same entity; the first head diffed the removal against authority and then applied
that same delta to the live graph, where kin-db checks a removal's old payload against the
graph it lands on. Authority had moved and the live graph had not. The plan no longer carries
retired deltas at all, only the vacated set, and the handler derives the live graph's own
retirement from `state.graph.to_snapshot()` with that set, the way `plan_daemon_semantic_delta`
already keeps the daemon-side delta separate from the authority-side one for every other
repository command. Test: `a_session_that_deletes_a_file_an_earlier_session_modified_reconciles`.

Under the speed rule I did not re-run the two probes red myself; the review's run at the
exact head is the red measurement, and it is the same sequence these tests drive.

### Falsification of the second head

F1, drop the `GraphNodeId::Artifact` arm from `departs` in `retire_semantics_on_vacated`:

```
test api::tests::a_session_that_deletes_an_imported_file_retires_its_artifact_edges ... FAILED
panicked at crates/kin-daemon/src/api.rs:35379:9:
assertion `left == right` failed: Core error: repository authority commit outcome is indeterminate:
  commit repository projection authority: storage error: transaction relation 1f886c9e-... has
  unadmitted destination endpoint artifact:1ddf52b8-...
  left: 500
 right: 200
```

F2, carry the plan's authority-derived deltas and apply them to the live graph instead of the
live-derived ones (the first head's exact behaviour):

```
Run on the third head, below, after the second head's own attempt failed to apply its mutation (the anchor matched two sites and the script aborted; that run's 'ok' was on an unmutated tree and is not cited).
```

### Not covered here: the live working-copy admission (journey GAP-11)

This fix is in `plan_session_workspace_admission` and the reconcile handler, which is the
session path FIR-3209 names. It does not touch the ambient and `kin admit` path, which
publishes through `publish_exact_workspace_tree` with no semantic delta at all. That is why
journey step 10 sees the same kin-db refusal when an entity-bearing file is removed in the
working copy, until the commit's own planner retires the entities and records the deletion
correctly. Same class, one path over; `retire_semantics_on_vacated` is written to be reusable
there, and it is named as the follow-up rather than folded in tonight.

## Third head: review round two

Both HIGHs were answered at the root on the second head. Two more before landing, because they
sit on the exact path memfix is fighting, where a one-file commit on a converted `psf/requests`
already peaks at 12.38 GiB.

**R2-1, the live retirement deep-cloned the whole store.** The second head derived the live
graph's retirement by diffing `state.graph.to_snapshot()`, which clones every sub-store and the
change DAG on every vacating reconcile. `retire_live_semantics_on_vacated` now reads only what
leaves: `query_entities` by each vacated path, `get_all_relations_for_node` on each retired
entity node and on each vacated artifact node, the deltas built from the live graph's own
payloads and sorted the way `commit_deltas` sorts them. Those are the same three reads the
watcher's removal path makes. The property that mattered on the second head is kept: each graph
derives its own retirement, and nothing authority holds is ever applied as a removal's `old` to
the live graph.

**R2-2, the artifact set is a `HashSet`**, matching the entity side beside it.

### Round-two falsification on this head

F1, drop the `GraphNodeId::Artifact` arm from `departs` on the authority side:

```
test api::tests::a_session_that_deletes_an_imported_file_retires_its_artifact_edges ... FAILED
panicked at crates/kin-daemon/src/api.rs:35380:9:
assertion `left == right` failed: Core error: repository authority commit outcome is indeterminate:
  commit repository projection authority: storage error: transaction relation fa8610a5-... has
  unadmitted destination endpoint artifact:daced884-...
  left: 500
 right: 200
```

F2, carry the plan's authority-derived deltas and apply them to the live graph instead of the
live-derived ones, the second head's split brain restored:

```
test api::tests::a_session_that_deletes_a_file_an_earlier_session_modified_reconciles ... FAILED
panicked at crates/kin-daemon/src/api.rs:35379:9:
assertion `left == right` failed: storage error: transaction has stale old payload for removed entity 0b2482f5-d1ae-5079-932f-d4f3822e455e
  left: 500
 right: 200
```

### Carried as follow-ups, in the reviewer's words

R2-3, LOW: `None => WorkspaceSemanticDelta::default()` still silently skips when the workspace
graph snapshot is absent. Every sibling errors there; kin-db returns `Ok(None)` only when the
workspace id is absent from authority metadata, which the plan has already validated, so it is
probably unreachable, and if it is ever reached the caller gets the original 500 back from a
lower layer with nothing in the log explaining it.

R2-4, LOW: the facet sweep, `clear_incompatible_facets_in(.., EnrichmentFacet::None)`, runs
once per retired entity with its failure downgraded to a `warn!` where the sibling propagates.
kin-db's `apply_transaction_delta` already retires path-keyed enrichment for every removal
under the same entity lock, so the sweep is a redundant pass run N times for N entities;
harmless, and worth collapsing to one call per distinct `file_origin` or dropping.

R2-5, LOW: rename gets two answers from one repository. The watcher and the MCP planner
relocate with identity; the session path retires and re-derives with new ids. This is a live
divergence between surfaces rather than only deferred work, and the follow-up ticket should
say so.

## Gates

Speed rule in force: no local precheck and no local clippy; hosted CI grades both the full
workspace and clippy. Run locally on head `f47883e5db6ac3e1e2f24991a2b9831ad547244e`, one slot job with the tree
restored after each mutation:

```
cargo test -p kin-daemon --lib -- deletes_an_entity_source_file deletes_an_imported_file
    deletes_a_file_an_earlier_session_modified          3 passed; 0 failed (7.89s)
F1 and F2 above, each 0 passed; 1 failed at the quoted line
rustfmt --edition 2021 --check crates/kin-daemon/src/api.rs
    crates/kin-daemon/src/repository_commit.rs            clean
```

The two touched files were format-checked with rustfmt directly rather than through
`cargo fmt --all -- --check`, because every cargo invocation on this box queues behind the
builder cap; hosted CI runs the workspace form. The first head carried a full local precheck
(21 gates ran, 21 passed on 03476033c) and the full kin-daemon lib suite (1392 passed, 0
failed).

## Follow-ups, not in this PR

R2-3, R2-4 and R2-5 above, in the reviewer's words. Relocation with identity for a renamed
entity-owning file is R2-5's ticket. The live working-copy admission (journey GAP-11) is
described in the second-head section.

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -6097,26 +6097,27 @@ async fn reconcile_session_workspace(
         if graph_tree == observation.base().source_workspace.tree {
             // Entities retire with the paths that owned them, in the same
             // step as the tree moves, so the live graph never holds an entity
-            // authority just refused to carry. Derived here from the LIVE
-            // graph with the plan's vacated set, never copied from the plan:
-            // the plan's delta was diffed against authority, and the live graph
-            // can already hold newer payloads for these entities from an
-            // earlier session's readmission. kin-db checks a removal's old
-            // payload against the graph it lands on, so authority's payload
-            // applied here was refused after authority had already moved,
-            // leaving the two graphs holding different trees.
-            let live = crate::repository_commit::retire_semantics_on_vacated(
-                &state.graph.to_snapshot(),
-                &vacated,
-            )
-            .map_err(repository_commit_error)?;
-            let retired_entity_deltas = live.entity_deltas().to_vec();
+            // authority just refused to carry. Read from the LIVE graph with
+            // the plan's vacated set, never copied from the plan: the plan's
+            // delta was diffed against authority, and the live graph can
+            // already hold newer payloads for these entities from an earlier
+            // session's readmission. kin-db checks a removal's old payload
+            // against the graph it lands on, so authority's payload applied
+            // here was refused after authority had already moved, leaving the
+            // two graphs holding different trees. Targeted reads rather than a
+            // snapshot of the whole store, which this path cannot afford.
+            let (retired_entity_deltas, retired_relation_deltas) =
+                crate::repository_commit::retire_live_semantics_on_vacated(
+                    state.graph.as_ref(),
+                    &vacated,
+                )
+                .map_err(repository_commit_error)?;
             state
                 .graph
                 .apply_transaction_delta(&TransactionDelta {
                     tree_deltas: observation.deltas().to_vec(),
                     entity_deltas: retired_entity_deltas.clone(),
-                    relation_deltas: live.relation_deltas().to_vec(),
+                    relation_deltas: retired_relation_deltas,
                     ..TransactionDelta::default()
                 })
                 .map_err(internal_error)?;
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -6066,6 +6066,8 @@ async fn reconcile_session_workspace(
             &observation,
         )
         .map_err(repository_commit_error)?;
+        let retired_entity_deltas = plan.retired_entity_deltas.clone();
+        let retired_relation_deltas = plan.retired_relation_deltas.clone();
         let committed = crate::repository_commit::commit_session_workspace_admission(
             &state.layout,
             state.blobs.as_ref(),
@@ -6094,13 +6096,43 @@ async fn reconcile_session_workspace(
 
         let graph_tree = state.graph.resolved_tree();
         if graph_tree == observation.base().source_workspace.tree {
+            // Entities retire with the paths that owned them, in the same
+            // step as the tree moves, so the live graph never holds an entity
+            // authority just refused to carry.
             state
                 .graph
                 .apply_transaction_delta(&TransactionDelta {
                     tree_deltas: observation.deltas().to_vec(),
+                    entity_deltas: retired_entity_deltas.clone(),
+                    relation_deltas: retired_relation_deltas,
                     ..TransactionDelta::default()
                 })
                 .map_err(internal_error)?;
+            for delta in &retired_entity_deltas {
+                let kin_model::EntityDelta::Removed { old } = delta else {
+                    continue;
+                };
+                if let Some(file_id) = &old.file_origin {
+                    if let Err(error) = crate::loop_runner::clear_incompatible_facets_in(
+                        state.graph.as_ref(),
+                        file_id,
+                        crate::loop_runner::EnrichmentFacet::None,
+                    ) {
+                        warn!(
+                            file = %file_id,
+                            error = %error,
+                            "retired a vacated path's entities but could not clear every facet"
+                        );
+                    }
+                }
+                state.emit_event(DaemonEvent::EntityChanged {
+                    entity_id: old.id,
+                    node: None,
+                    change_type: crate::state::ChangeType::Deleted,
+                    file_path: old.file_origin.as_ref().map(|file| file.0.clone()),
+                    session_id: None,
+                });
+            }
         } else if graph_tree != *observation.desired_tree() {
             return Err((
                 StatusCode::CONFLICT,
@@ -35624,6 +35656,132 @@ mod tests {
             .resolved_tree()
             .artifact_at_path(&kin_model::RepoPath::from_utf8("pending.txt").unwrap())
             .is_none());
+    }
+
+    /// FIR-3209. A session that deletes an entity-owning file reconciles, the
+    /// file's entities are absent from every query afterwards, and the next
+    /// commit records the removal.
+    ///
+    /// kin-db refuses a transition that leaves an entity on a path the staged
+    /// tree no longer carries, and it names the remedy in its own message:
+    /// "carry its exact entity removal or relocation in the same delta". The
+    /// invariant is right and the session admission was the caller that owed
+    /// the removal, so an agent's `rm` of a source file could not reconcile at
+    /// all. The existing session-reconcile test deletes `delete.me`, a file with
+    /// no entities, which is why the suite was green over this.
+    ///
+    /// Falsify by restoring `semantic_delta: WorkspaceSemanticDelta::default()`
+    /// in `plan_session_workspace_admission`: the reconcile returns the 500 the
+    /// review's control probe measured, and `gone.rs` keeps its entity with no
+    /// file behind it.
+    #[cfg(unix)]
+    #[tokio::test]
+    #[serial_test::serial(commit_phase_capture)]
+    async fn a_session_that_deletes_an_entity_source_file_retires_its_entities() {
+        let repo = tempfile::tempdir().unwrap();
+        let initialized = kin_core::init(repo.path()).unwrap();
+        let state = Arc::new(DaemonState::open(initialized.layout).unwrap());
+        state
+            .is_initialized
+            .store(true, std::sync::atomic::Ordering::Relaxed);
+        std::fs::write(repo.path().join("keep.rs"), b"pub fn keep() -> u32 { 1 }\n").unwrap();
+        std::fs::write(repo.path().join("gone.rs"), b"pub fn gone() -> u32 { 7 }\n").unwrap();
+        let app = router(Arc::clone(&state));
+        commit_through_api(&app, kin_model::OperationId::new(), "publish the fixtures").await;
+        let entities_at = |path: &str| {
+            state
+                .graph
+                .query_entities(&kin_db::EntityFilter {
+                    file_path: Some(kin_model::FilePathId::new(path)),
+                    ..Default::default()
+                })
+                .unwrap()
+        };
+        let gone_before = entities_at("gone.rs");
+        eprintln!(
+            "FIR-3209 measurement: before gone.rs entities={} keep.rs entities={}",
+            gone_before.len(),
+            entities_at("keep.rs").len()
+        );
+        assert_eq!(gone_before.len(), 1, "the fixture owns exactly one entity");
+        let gone_id = gone_before[0].id;
+
+        let session_dir = state.layout.root().join("runs/session-fir3209");
+        materialize_session_through_api(&app, &session_dir).await;
+        std::fs::remove_file(session_dir.join("gone.rs")).unwrap();
+
+        let response = app
+            .clone()
+            .oneshot(
+                Request::post("/reconcile")
+                    .header("content-type", "application/json")
+                    .body(Body::from(
+                        json!({"session_dir": session_dir, "confirm_mass_deletion": false})
+                            .to_string(),
+                    ))
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+        let status = response.status();
+        let body = axum::body::to_bytes(response.into_body(), 128 * 1024)
+            .await
+            .unwrap();
+        let body = String::from_utf8_lossy(&body).to_string();
+        eprintln!(
+            "FIR-3209 measurement: reconcile status={status} after gone.rs entities={} body={body}",
+            entities_at("gone.rs").len()
+        );
+        assert_eq!(
+            status,
+            StatusCode::OK,
+            "deleting a source file in a session is an ordinary edit and must reconcile: {body}"
+        );
+        let summary: kin_cli::commands::reconcile::ReconcileSummary =
+            serde_json::from_str(&body).unwrap();
+        assert_eq!(summary.removed, 1);
+
+        // Absent from every query, not only the by-path one the invariant is
+        // phrased in: by id, by name, and from the exact tree.
+        assert!(
+            entities_at("gone.rs").is_empty(),
+            "an entity must not outlive the path that owned it"
+        );
+        assert_eq!(entities_at("keep.rs").len(), 1, "the sibling is untouched");
+        assert!(state.graph.get_entity(&gone_id).unwrap().is_none());
+        assert!(state
+            .graph
+            .query_entities(&kin_db::EntityFilter {
+                name_pattern: Some("gone".to_string()),
+                ..Default::default()
+            })
+            .unwrap()
+            .is_empty());
+        assert!(state
+            .graph
+            .resolved_tree()
+            .artifact_at_path(&RepoPath::from_utf8("gone.rs".to_string()).unwrap())
+            .is_none());
+
+        // The next commit carries the removal into history.
+        let change_id =
+            commit_through_api(&app, kin_model::OperationId::new(), "commit the deletion").await;
+        let change = state.graph.get_change(&change_id).unwrap().unwrap();
+        assert!(
+            change.entity_deltas.iter().any(|delta| matches!(
+                delta,
+                kin_model::EntityDelta::Removed { old } if old.id == gone_id
+            )),
+            "the commit after the session must record the entity's removal: {:?}",
+            change.entity_deltas
+        );
+        assert!(
+            change
+                .tree_deltas
+                .iter()
+                .any(|delta| matches!(delta, kin_model::TreeDelta::Removed { .. })),
+            "and the file's removal beside it"
+        );
     }
 
     /// Changes in repository authority, read the way `kin log` reads them.
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -6066,8 +6066,7 @@ async fn reconcile_session_workspace(
             &observation,
         )
         .map_err(repository_commit_error)?;
-        let retired_entity_deltas = plan.retired_entity_deltas.clone();
-        let retired_relation_deltas = plan.retired_relation_deltas.clone();
+        let vacated = plan.vacated.clone();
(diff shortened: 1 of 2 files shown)


troyjr4103:
## Independent exact-head review at `03476033c554a18b88614ade618d23ccff1f04bb`

**NO-GO.** Two measured 500s in cases one step from the one the new test covers, reproduced twice
each in a separate worktree pinned to this exact sha. Hosted CI is genuinely green over the full
set (47 check runs, 47 listed, 30 success, 17 skipped, zero failures, all seven required contexts
by name), so nothing here is a CI complaint. The design is right; two adjacent cases are not
covered, and one of them wedges the daemon.

### HIGH 1. Deleting a file another file imports still returns 500

`retire_entities_on_vacated_paths` drops entities on a vacated path and relations with such an
entity at either end. `touches_retired` (`repository_commit.rs:505`) matches only
`GraphNodeId::Entity`, so a relation whose endpoint is the vacated path's **artifact** node survives
into `desired_relations` and `diff_workspace_semantics` emits no removal for it. kin-db validates
artifact endpoints against the staged artifact set (`kin-db 0.7.101 engine/graph.rs:8819`), and a
vacated path's artifact is exactly what leaves it.

Fixture: `gone.ts` exporting one function, `keep.ts` importing it, session deletes `gone.ts`.

```
PROBE C: gone.ts artifact=Some(ArtifactId(8cfca3a8-bbf3-4b9f-8bbc-594b1d200291)) artifact_node_relations=1 kinds=["Imports:artifact:a0312389-...->artifact:8cfca3a8-..."]
PROBE C: reconcile status=500 Internal Server Error body=Core error: repository authority commit outcome is indeterminate: commit repository projection authority: storage error: transaction relation 61f3a727-758a-44d5-6534-29650347b113 has unadmitted destination endpoint artifact:8cfca3a8-bbf3-4b9f-8bbc-594b1d200291; ...
  left: 500
 right: 200
```

Those edges are ordinary production output: `kin-index/src/linker.rs:4621` mints
artifact-to-artifact `Imports`/`Includes`, `:4857` the parse-coverage `DependsOn` self-loop, `:4981`
the `DerivedFrom` projection markers. The sibling path already answers this with
`retire_artifact_node_relations` (`loop_runner.rs:1324`, called at `:1429`), whose doc comment names
the same refusal and says "nothing else collects these".

### HIGH 2. Modify in one session, delete in the next: 500 after a durable commit, then the daemon stops accepting session reconciles

The delta's `old` payloads come from the authority snapshot, which is correct for the authority
transaction. The same delta is then applied unchanged to the **live** graph (`api.rs:6104`), and
kin-db requires an exact payload match on `EntityDelta::Removed`. The live graph is allowed to
differ: `readmit_semantics_for_paths` applies re-derived semantics to `state.graph` alone and calls
`mark_enrichment_unpublished` right after. A session reconcile that only modifies a file vacates
nothing, so authority keeps the payload it had while the live graph moves on.

```
PROBE E: modify reconcile status=200 OK ... "modified":1 ... "semantic_files_enriched":1
PROBE E: delete reconcile status=500 Internal Server Error after gone.rs entities=1 body=storage error: transaction has stale old payload for removed entity 0b2482f5-d1ae-5079-932f-d4f3822e455e
PROBE E: third materialize status=200 OK
PROBE E: third reconcile (unrelated edit) status=409 Conflict body=daemon query tree is not bound to the session base or its exact recovered target; reopen the daemon from repository authority
```

That 500 body carries no `Core error: repository authority commit outcome is indeterminate` wrapper.
`repository_commit_error` adds that wrapper and `internal_error` does not, and only the live-graph
apply uses `internal_error`, so the authority commit had already succeeded and
`record_repository_authority_commit` had already run. Authority retired the entity and moved the
tree, the live graph did neither, and every later session reconcile then meets the tree-binding
guard at `api.rs:6033` and returns 409 until the daemon is restarted. Nothing in the reconcile path
clears it: `reset_derived_graph_to_authority_tree` is the routine that would, and its one caller is
`loop_runner.rs:8475`.

Pre-fix this shape failed with nothing durable. Post-fix it fails with a durable commit behind it.

The house order is the other way round. `evict_enrichment_for_removed_paths` clears the live graph
directly and then applies tree deltas with empty entity and relation deltas
(`loop_runner.rs:924` then `:941`), and `reset_derived_graph_to_authority_tree` does the same,
calling it "the order every other tree removal uses" (`loop_runner.rs:8402`).

### Lower

3. MEDIUM. `clear_incompatible_facets_in` returns a `FacetCleanup` whose `removed_entities` the new
   loop discards, so live-only entity removals reach no subscriber. The sibling emits one
   `EntityChanged` per cleaned entity (`loop_runner.rs:1434`).
4. MEDIUM. A rename now gets two different answers from one repository: the watcher
   (`loop_runner.rs:935`) and the MCP planner (`mcp_commit.rs:1917`) relocate with identity
   preserved, the session path retires and re-derives with new ids. The body calls this a follow-up;
   worth saying in the follow-up that it is a live divergence between surfaces, not only deferred
   work, because `path_relocations_in`'s own doc comment says the pair "would mint new entity ids
   and orphan every incoming reference".
5. LOW. `None => WorkspaceSemanticDelta::default()` (`repository_commit.rs:311`) silently skips the
   retirement when the workspace has no authority graph snapshot. Every sibling errors
   (`repository_commit.rs:1040`, `repository_checkout.rs:270`).
6. LOW. The facet sweep runs once per retired entity rather than once per vacated file, and its
   failure is a `warn!` where the sibling propagates with `?` (`loop_runner.rs:1433`), so a failed
   sweep leaves the layout, shallow file and artifacts on a path the tree no longer carries and
   still returns 200.
7. LOW. The body's quoted tails are faithful in substance but are trimmed prefixes, re-wrapped, with
   the truncation unmarked, so they cannot be diffed against the logs mechanically.

### What is right, and worth keeping in whatever lands

- The delta is derived against the **authority** snapshot on both sides of
  `diff_workspace_semantics`, never against the live graph, in the same shape as every other
  repository command in this crate.
- It rides the same workspace mutation as the tree, under the existing
  `WorkspaceExpectation::MustEqual` compare-and-swap, so a race between the snapshot read and the
  later freshness read is refused rather than accepted.
- Nothing gained a raw filesystem read. Grepping the added lines for the filesystem idioms hits
  three times, all three inside the test fixture; the same pattern on the same input is a positive
  control at 3.
- The measurement is genuinely red-first: `commitfresh-3209-red.log` at 00:03:18Z precedes the fixed
  arm at 00:06:44Z, and every number in the body matches the logs.
- `21 gates enumerated from the check job of .github/workflows/ci.yml` reproduces exactly at this
  head with `bin/kin-precheck kin --list`.
- The test goes through the real router and the real `/reconcile`, and asserts absence by path, by
  id, by name and against the exact tree.

### One measurement that refuted its own reading

I expected the swap shape (delete `a.rs`, move `b.rs` onto its path) to leave `a.rs`'s entities on a
path another artifact now holds, because `vacated_paths` is defined by path. It does not: the
planner keeps `a.rs`'s own artifact at `a.rs` with the new bytes, and readmission replaces `alpha`
with `bravo` in the live graph. `PROBE F: after a.rs entities=["bravo"] b.rs entities=0`, status 200.
Reported because it is one more ordinary edit that opens the authority-versus-live gap finding 2
falls into, not because it is a bug of its own.

troyjr4103:
## F2 falsification tail, head `f47883e5d`

Carry the plan's authority-derived deltas and apply them to the live graph instead of the live-derived ones (the second head's split brain restored). Log `.kin-coord/reports/commitfresh-3209-f2.log`, run 02:32Z:

```
test api::tests::a_session_that_deletes_a_file_an_earlier_session_modified_reconciles ... FAILED
panicked at crates/kin-daemon/src/api.rs:35379:9:
assertion `left == right` failed: storage error: transaction has stale old payload for removed entity 0b2482f5-d1ae-5079-932f-d4f3822e455e
  left: 500
 right: 200
```

F1 on the same head, dropping the artifact-node arm on the authority side:

```
test api::tests::a_session_that_deletes_an_imported_file_retires_its_artifact_edges ... FAILED
panicked at crates/kin-daemon/src/api.rs:35380:9:
assertion `left == right` failed: Core error: repository authority commit outcome is indeterminate:
  commit repository projection authority: storage error: transaction relation fa8610a5-... has
  unadmitted destination endpoint artifact:daced884-...
  left: 500
 right: 200
```

The three FIR-3209 tests pass together on the unmutated tree (3 passed, 7.89s).

troyjr4103:
## Round three review at `f47883e5db6ac3e1e2f24991a2b9831ad547244e`

**GO.** Both HIGH findings from `03476033c` stay closed, R2-1 and R2-2 are done, and the rewrite that
did R2-1 is a real one rather than a shortcut. Re-ran the same three probes against this head in a
separate worktree. All green.

```
PROBE C: gone.ts artifact_node_relations=1 kinds=["Imports:artifact:25074365-...->artifact:eff778ab-..."]
PROBE C: reconcile status=200 OK ... "removed":1 ... "kind":"removed","old_path":"gone.ts"
PROBE E: modify reconcile status=200 OK ... "semantic_files_enriched":1 ... after entities=1
PROBE E: delete reconcile status=200 OK after gone.rs entities=0
PROBE E: third reconcile (unrelated edit) status=200 OK
PROBE F: after a.rs entities=["bravo"] b.rs entities=0
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 1391 filtered out; finished in 6.55s
```

### R2-1 checked, not taken on trust

`to_snapshot` is gone from `reconcile_session_workspace`: zero occurrences inside the function
(`api.rs:5981-6224`), against 22 elsewhere in the same file as the control that the search can hit.
`retire_live_semantics_on_vacated` reads only what leaves, and the authority side still diffs the
authority snapshot at `repository_commit.rs:313`, so the property that made the second head correct
is untouched. Each graph derives its own retirement, and nothing authority holds is applied as a
removal's `old` to the live graph.

One thing had to be true for the rewrite to be safe, and is: `get_all_relations_for_node` returns
**incoming and outgoing** edges, deduped by relation id. The authority side reaches that set by
filtering every relation on either endpoint; the live side now reaches it by querying per node. Had
that accessor been outgoing-only, the imported-file case would have silently broken again, because
the edge points AT the deleted file. The crate test's fixture covers that direction and so does
probe C, whose first printed line confirms the linker really did bind the edge before the delete.

R2-2 is done: `VacatedPaths::artifacts` is a `HashSet` and `from_deltas` inserts, so the two arms of
`departs` now agree.

### The falsification arms are checkable

F1 and F2 are local-only and hosted CI never runs them, so they need the command and its actual
output tail rather than prose. They have it, and the logs the comment names exist at 02:30:47Z and
02:32:08Z, both before the 02:33Z push, with the quoted `stale old payload for removed entity
0b2482f5-...` line present verbatim.

The two panic locations sit one line apart, `api.rs:35380` for F1 and `:35379` for F2, which reads
like a transcription slip until you check it. Both land on the shared `reconcile_session_through_api`
helper's `assert_eq!`, and F2's mutation is in `api.rs` while F1's is in `repository_commit.rs`, so
F2's arm shifts that helper by exactly one line and F1's does not. Internally consistent.

### Hosted CI at this exact head

Read over the full set with `per_page=100`, `total_count` compared to what was listed, by name:
47 check runs, 47 listed, 30 success, 17 skipped, zero failures, zero cancelled, zero timed out. All
seven required contexts green. `Analyze (rust)` concluded success at 02:48:15Z, so `CodeQL`'s success
is read after its Rust analysis rather than before it.

`Fast gate lint and policy` green covers the one thing the body is honest about: the two touched
files were format-checked with `rustfmt --edition 2021 --check` directly rather than through
`cargo fmt --all -- --check`, because every cargo invocation on this box queues behind the builder
cap. Hosted CI ran the workspace form and it passed.

### Nothing left blocking

R2-3, R2-4 and R2-5 are carried in the body as follow-ups in the terms I filed them. Nothing I found
across three heads is outstanding against this one.

PR #1504, opened 2026-09-05
Serve a local daemon's refs read from the cached authority (FIR-3180)

troyjr4103:
FIR-3180. **The production 504 is on the hosted arm and this PR does not close it.** Do not close
FIR-3180 on this PR. What closes it is the image hop: production runs `d6b61e550` = kin **0.6.1**,
and the hosted reuse landed in **0.6.2**. Reading the first call inside `repo_refs` at each tag,
`git show <tag>:crates/kin-daemon/src/api.rs`: `v0.6.1` (line 11763) calls
`repository_authority_snapshot`; `v0.6.2` (14691), `v0.6.3`, `v0.6.5`, `v0.6.7` and `origin/main`
all call `repository_ref_metadata`. Measured from inside the cluster at 02:21Z, production
`/repos/kin/refs` answered 200 in 38.94, 38.58 and 39.62 s on three consecutive reads, which is the
local before-median of 39.33 s to the second, and its log carries one `loaded graph from storage
backend` at startup and none since, because 0.6.1's per-request cost is
`RepositoryAuthorityManager::open`, not `load_repo_graph`.

This PR fixes the same defect on the **local** arm, which no release has fixed.

## What the read path does

`repo_refs` calls `repository_ref_metadata`, which splits by capability. The hosted arm
(`storage_backend.is_some()`) answers from the generation-bound cache: warm and unchanged, that is a
map read plus one `probe_repo_publication`, which on `GcsBackend` is a scoped list of delta objects
plus a head on the snapshot object, with no body download and no authority open. The local arm fell
through to `repository_authority_snapshot` and `ActiveApiRepositoryAuthority::open`, whose own doc
says it "reads the durable publication in full", on every request, with nothing caching it between
requests.

## Numbers, by arm

Store for both runs: `kin init --no-enrich` over a single-branch clone of firelock-ai/kin at
`ae8214296867870acb271db96c75bc0e75a9613b`, 992 files, 2864 changes, 33628 entities, 7.6 GiB under
`.kin/`. Every sample HTTP 200, 21685 bytes, 111 refs. Timings are curl `%{time_total}`, gated on
`/readiness`, with a positive control asserting a non-empty `refs` array before any timing is kept.

**Local arm, warm, ten reads on one daemon.** This is the arm the change is about.

| | min | median | mean | max | load avg |
|---|---|---|---|---|---|
| before, `da356723b` | 37.35 s | 44.63 s | 45.04 s | 61.87 s | 19 to 39 |
| after, this branch | 0.001 s | **0.001 s** | 0.002 s | 0.004 s | 82.17 |

A median of 1.2 milliseconds against 44.6 seconds, on a box four times busier than the before run.

**Local arm, cold, ten daemon starts, one timed read each.**

| | min | median | mean | max | load avg |
|---|---|---|---|---|---|
| before, `da356723b` | 34.20 s | 39.33 s | 39.84 s | 48.45 s | 17 to 34 |
| after, this branch | 78.72 s | 127.10 s | 125.19 s | 165.30 s | 57 to 131 |

**The cold rows are not comparable.** The box carried 47 cargo and rustc processes and about twenty
lanes throughout the after run. Structurally the cold path now does strictly less work than before,
one envelope clone instead of a full snapshot clone plus an envelope clone, so the difference is
contention. The cold row is published rather than dropped so nobody has to take that on trust.

The read path is byte-identical between `da356723b` and this PR's base `f6a29e329`, checked function
by function on `repo_refs`, `repository_ref_metadata`, `repository_authority_snapshot` and
`projection_repository_authority`, so the before numbers carry to the base.

**A cold first read still pays one full open.** Nothing populates `ProjectionAuthorityCache` at
startup: `projection_repository_authority` has three callers and all three are request paths. So a
restarted local daemon's first refs read still costs one open, and that is the request that would
504. Warm-on-start is a named follow-up, not in this PR, and it is local-arm only: a hosted daemon
already pre-loads its own repo's cache entry with authority and cursor before the state is visible
(`state.rs:5650`).

Two corrections for the ticket. The 315 bytes in it is the response size, not the work: these reads
returned 21685 bytes for the same 40 s, so the cost tracks the publication being opened, not the
payload. And `/health` answers 200 while the daemon is still loading repository authority, after
which every graph route refuses in under a millisecond with `daemon_opening`; a measurement gated on
`/health` times a 0.0007 s 503 and reports the defect fixed.

## The change

The envelope is held per durable publication in the `ProjectionAuthorityCache` that already holds
three other authority wrappers, keyed on the same `LocalPublicationIdentity` and cleared by the same
`invalidate`, so the refs read inherits that invalidation contract rather than adding a second one.
It is an `Arc` because the envelope carries an operation log and a receipt per change, so cloning it
per request is the same defect one layer down. The label is read before the load it describes, which
at worst costs one spurious reload and never serves a publication that has moved. The pinned
namespace is probed before the reuse check, as the three siblings do, and the slot is re-checked
after the open so a burst of cold requests pays one envelope clone rather than one each.

## Test and its falsification

`a_refs_read_costs_one_authority_open_per_publication`, two arms: eight reads at one publication add
zero authority opens, and a read after a commit adds exactly one and returns different bytes. The
second arm counts loads from **after** the commit, because a commit opens authority for its own
reasons and a count taken before it is satisfied by the commit's open.

Falsified on this exact head by bypassing the new branch with
`if false && repo_id == state.cached_repo_id`, run through heavy-slot, with the tree restored from a
pristine copy under a trap afterwards:

```
armed: the cached branch is bypassed
thread 'api::tests::a_refs_read_costs_one_authority_open_per_publication' panicked at
crates/kin-daemon/src/api.rs:22871:9:
assertion `left == right` failed: eight refs reads at one publication must open the durable
authority zero further times; each open reads the whole publication and re-verifies every CAS body
  left: 9
 right: 1
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1394 filtered out; finished in 1.97s
falsification test exit rc=101 (non-zero is the expected result)
restored api.rs from the pristine copy
```

Left 9 against right 1 is eight refs reads each paying their own authority open, which is the defect.

An earlier falsification attempt patched the other occurrence of the same line, in
`repository_authority_snapshot` rather than in `repository_ref_metadata`, and passed green while
proving nothing. That is why the run above prints the branch it armed.

## What ran locally

Per tonight's speed rule, no `bin/kin-precheck`, no `bin/kin-parity`, and no local clippy; hosted CI
grades clippy and the full workspace. Everything below ran through
`.kin-coord/bin/heavy-slot.sh refsmeasure kin`, with `RUSTFLAGS=-D warnings`.

```
$ cargo fmt --all -- --check
(exited 0 with no output)

$ cargo test -p kin-daemon --lib
test result: ok. 1392 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 213.03s

$ cargo test -p kin-daemon --lib a_refs_read_costs_one_authority_open_per_publication -- --nocapture
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1394 filtered out; finished in 2.39s
```

The full-crate run above was taken on `315f7f620`; the targeted run and fmt were re-taken on this
head after the review fixes. Clippy on this head is graded by hosted CI.

## Not in this PR

`repo_history`, `repo_files` and the other local-arm routes still reach
`repository_authority_snapshot` and pay a full open per request. Same defect, same fix shape.

### DIFF
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -518,6 +518,15 @@ fn local_repository_ref_metadata(
         .map_err(repository_authority_error)?
         .repository_id()
         .clone();
+    // The pinned-namespace probe runs before the reuse check, exactly as the
+    // three sibling resolvers run it, so a `.kin/kindb` replaced under the
+    // daemon cannot keep being answered from the envelope of the store it
+    // replaced. It decodes no snapshot and takes no repository lock.
+    if let Err(error) = confirm_pinned_projection_namespace(&backend, &repository_id) {
+        state.projection_authority.invalidate();
+        return Err(error);
+    }
+
     let published = read_local_publication_identity(&backend, &repository_id)?;
     if let Some(metadata) = state.projection_authority.reuse_ref_metadata(&published) {
         return Ok(metadata);
@@ -526,6 +535,14 @@ fn local_repository_ref_metadata(
     // The authority open itself is already cached and gated one layer down, so
     // this reaches a full open only when that slot misses too.
     let authority = projection_repository_authority(state)?;
+    // Re-check after that open. A burst of concurrent cold requests is
+    // serialized by the gate inside `projection_repository_authority` and then
+    // released one at a time into the clone below, and the clone is O(history);
+    // whoever loses the race takes the envelope the winner installed instead of
+    // cloning a second copy of it.
+    if let Some(metadata) = state.projection_authority.reuse_ref_metadata(&published) {
+        return Ok(metadata);
+    }
     let metadata = {
         let lease = authority.manager.read_authority();
         lease
@@ -22867,6 +22884,11 @@ mod tests {
             "publish a second change so the refs slot must invalidate",
         )
         .await;
+        // Counted AFTER the commit, not before it. A commit opens authority for
+        // its own reasons, so a count taken before it is satisfied by the
+        // commit's open and says nothing about the read. This one moves only if
+        // the refs read itself reloaded.
+        let after_commit = state.projection_authority.loads();
         let (status, after) = repo_route(Arc::clone(&state), &path).await;
         assert_eq!(
             status,
@@ -22874,15 +22896,15 @@ mod tests {
             "the read after a publication must answer: {}",
             String::from_utf8_lossy(&after)
         );
-        assert!(
-            state.projection_authority.loads() > before,
-            "a refs read after a publication must open the authority again rather than serve \
-             the publication before it"
-        );
         assert_ne!(
             after, first,
             "the refs read after a publication still returns the pre-commit head, so the slot \
              is pinned rather than keyed on the publication"
+        );
+        assert_eq!(
+            state.projection_authority.loads(),
+            after_commit + 1,
+            "the refs read after a publication must itself reload the authority exactly once"
         );
     }
 
--- a/crates/kin-daemon/src/api.rs
+++ b/crates/kin-daemon/src/api.rs
@@ -236,6 +236,9 @@ pub(crate) struct ProjectionAuthorityCache {
     /// that function at one instant, serialized at 2.2 seconds per open
     /// because an open is O(store).
     admission: std::sync::Mutex<Option<HeldAdmission>>,
+    /// The refs routes' envelope, keyed on the same publication identity as
+    /// every slot above it.
+    ref_metadata: std::sync::Mutex<Option<HeldRefMetadata>>,
     /// Serializes misses so a burst of concurrent cold requests pays for one
     /// full load rather than one per request.
     load_gate: std::sync::Mutex<()>,
@@ -268,6 +271,21 @@ struct HeldCommandAuthority {
     /// [`HeldProjectionAuthority::published`] and for the same reason.
     published: LocalPublicationIdentity,
     authority: Arc<kin_cli::commands::repository_authority::ActiveRepositoryAuthority>,
+}
+
+/// The repository authority envelope a refs read answers from.
+///
+/// Held as an `Arc` rather than rebuilt, because the envelope carries an
+/// operation log and a receipt per change, so cloning it is O(history) and
+/// paying that per request is the same defect one layer down from the open it
+/// replaces. The hosted arm already hands out `Arc<PersistedRepositoryAuthority>`
+/// out of its generation-bound cache entry; this is the local arm's equivalent.
+#[derive(Clone)]
+struct HeldRefMetadata {
+    /// Read strictly before the authority it came out of was loaded, exactly as
+    /// for [`HeldProjectionAuthority::published`] and for the same reason.
+    published: LocalPublicationIdentity,
+    metadata: Arc<kin_db::PersistedRepositoryAuthority>,
 }
 
 #[derive(Clone)]
@@ -370,10 +388,32 @@ impl ProjectionAuthorityCache {
         });
     }
 
+    fn reuse_ref_metadata(
+        &self,
+        published: &LocalPublicationIdentity,
+    ) -> Option<Arc<kin_db::PersistedRepositoryAuthority>> {
+        lock_recover(&self.ref_metadata)
+            .as_ref()
+            .filter(|held| &held.published == published)
+            .map(|held| Arc::clone(&held.metadata))
+    }
+
+    fn install_ref_metadata(
+        &self,
+        published: LocalPublicationIdentity,
+        metadata: Arc<kin_db::PersistedRepositoryAuthority>,
+    ) {
+        *lock_recover(&self.ref_metadata) = Some(HeldRefMetadata {
+            published,
+            metadata,
+        });
+    }
+
     fn invalidate(&self) {
         *lock_recover(&self.held) = None;
         *lock_recover(&self.query) = None;
         *lock_recover(&self.command) = None;
+        *lock_recover(&self.ref_metadata) = None;
     }
 }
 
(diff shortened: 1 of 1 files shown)


troyjr4103:
Review of `315f7f620`, read-only, by lane review1504. Every line below was read from source at that
exact sha, and the falsification was re-run in a separate worktree.

**The verdict is a hold on the body, not on the code.** None of the three refusal conditions this
review was given is hit: there is no second cache, no second invalidation contract, no stale
envelope after a publication moves, and no filesystem read on the query path. Two things in the
body need to change before this lands, and neither needs a code edit.

## What checks out

The envelope goes in the existing `ProjectionAuthorityCache`, keyed on the same
`LocalPublicationIdentity` the three sibling slots use, and `invalidate` at `api.rs:412` clears it
alongside them. The label is read before the load it describes, which is the order the siblings
document and the safe direction: an older label on newer-or-equal bytes costs at most one spurious
reload and cannot serve a publication that has moved.

No raw filesystem read entered the query path. Grepping the diff's added lines for
`std::fs|fs::|File::|read_to_string|read_dir|walkdir|WalkDir|glob|OpenOptions` returns nothing,
against a control pattern that returns 29 on the same input. The only file read on the new path is
the pre-existing `read_local_publication_identity`, which is storage-commit metadata rather than a
semantic answer.

Blast radius is one route: `repository_ref_metadata` has exactly one caller, `repo_refs` at
`api.rs:16015`. The new 424 is byte-identical to `repository_metadata`'s.

The body's byte-identical claim holds. Extracting `repo_refs`, `repository_ref_metadata`,
`repository_authority_snapshot` and `projection_repository_authority` at both `da356723b` and
`f6a29e329` and diffing: all four identical, 64, 20, 58 and 44 lines, against a positive control
showing the same comparison reports a difference over the whole file. The before numbers match the
author report cell for cell.

Hosted checks on this exact head, read over the full set: 47 listed of 47 total, zero failures, all
seven required contexts green by name. `Analyze (rust)` is still `in_progress`, so `CodeQL`'s
`neutral` is not yet a rust result.

## 1. HIGH. This is not on the arm production runs, so the 504 the body opens with stays open

`repository_ref_metadata` returns from the hosted branch at `api.rs:14834`
(`if state.storage_backend.is_some()`) before it reaches the new branch at `:14850`. A daemon with a
storage backend never touches this change.

The deployed `kin-daemon` container, the one on port 4219 the control plane reads, runs
`KIN_STORAGE=gcs` (`kin-infra/compute/workloads.ts:995`). Two gates assert it:
`kin-infra/scripts/test-daemon-start-requirements.sh:224`, whose own failure message is "the
kin-daemon container does not set KIN_STORAGE to gcs, so nothing below is about the hosted daemon",
and `kinlab/scripts/check-daemon-env-contract.mjs:255`. The one deployed pod at `KIN_STORAGE=local`
is `kin-registry` (`workloads.ts:760`), whose own comment says it "serves crate/manifest files from
the PVC and never touches the graph".

The chain is closed with infra source alone, without relying on the author's report. `workloads.ts`
declares one `daemonDeployment` (`:861`). Inside that one pod spec, the `kin-daemon` container on
port 4219 carries `KIN_STORAGE=gcs` (`:995`), and the control-plane container on port 4010 carries
`KIN_DAEMON_URL=http://localhost:4219` (`:1216`). So the public route reaches the control plane,
which calls the gcs daemon over localhost in the same pod, which takes the hosted branch.

The author's report says the same thing twice: "the production defect lived on the hosted arm and
the FIR-2924 fix landed on the hosted arm", and "it is not the hosted arm, which is what production
runs".

The body's last line does say the hosted arm is untouched, so nothing here is false. The problem is
the opening: it leads with the public 504 and the ticket is that 504, so a reader lands on "this
fixes it". Ask: say plainly in the body that the production 504 is on the hosted arm and remains
open, and do not close FIR-3180 on this PR.

## 2. HIGH. The after numbers are not in the body

Verbatim: "Local arm, after: **pending, measuring now, will be edited in before review.**" Cold and
warm, separately, with the sample count. This is the hold.

## 3. MEDIUM. A cold first read still pays a full open, and the body does not say so

`projection_repository_authority` has exactly three callers (`api.rs:528`, `:16328`, `:16680`), all
request paths. Nothing populates `ProjectionAuthorityCache` at startup. So the first refs read after
every daemon start pays one `ActiveApiRepositoryAuthority::open`, which is the 39.33 s median this
body already reports. That is a fair follow-up rather than a blocker, but the body should name it,
because a restarted pod's first refs read is exactly the request that 504s.

## 4. MEDIUM. The refs route is the only authority resolver that skips the pinned-namespace probe

The three siblings call `confirm_pinned_projection_namespace` before their reuse check and
`invalidate()` on failure (`api.rs:456`, `:590`, `:658`). Those three are the only call sites of
`invalidate()` in the crate. `local_repository_ref_metadata` does neither, so on a cache hit the
refs route answers without confirming the daemon still pins that namespace, and it is the only route
that can never trigger the invalidate that clears its own slot.

I could not construct a wrong answer from it. A missing record reads as `Unpublished`, which cannot
match a `Published` label, and the open that follows fails closed
(`kin-core/src/repository_authority.rs:722`); identical record bytes mean identical durable state by
construction. So this is a consistency and defence-in-depth gap, not a demonstrated staleness bug.
The fix is the same four lines the siblings already carry.

## 5. LOW. The slot is filled outside the load gate

`projection_repository_authority` serializes the open under `load_gate`, but
`local_repository_ref_metadata` clones the envelope and installs after that gate is released. A
burst of concurrent cold refs requests therefore pays one O(history) envelope clone per request,
which is the cost `HeldRefMetadata`'s own doc says it exists to avoid. Bounded to the miss path, so
latency only.

## 6. LOW. Two quoted gate lines are prose, not tails

`clippy debt allow-list: 45 lints` is genuine and comes from
`.kin-coord/bin/vectorconsumer-clippy.sh:18`. `fmt ok` and `clippy ok` are printed by no wrapper
under `bin/` or `.kin-coord/bin/`, and `cargo fmt --all -- --check` prints nothing on success, so
those two are a summary set inside a code fence.

Separately, the quoted falsification tail cannot be reproduced from this head plus the one-line
bypass it describes. It carries `FIR3180DEBUG loads_before=1 loads_after=9`, and `FIR3180DEBUG`
appears nowhere in the committed head, so that tree also held a temporary debug print; the quoted
panic line `api.rs:22858` sits about four lines below where the `assert_eq!` actually is at this
head. Nothing is misstated, but the tail is not reproducible as written. A clean re-run is below.

## 7. LOW. The second test arm's load assertion is loose

`assert!(loads() > before)` after the commit can be satisfied by the commit's own open rather than
by the refs read reloading. The `assert_ne!(after, first)` beside it is the assertion that actually
carries the property, and it is sound, so the arm is fine as a pair. Worth knowing which half is
load-bearing.

## The falsification, re-run in a separate worktree

The brief's warning is real: `if repo_id == state.cached_repo_id` appears at `api.rs:14799` inside
`repository_authority_snapshot` and again at `:14850` inside `repository_ref_metadata`, and only the
second is the new branch. Patched by line number, with both lines and the enclosing function printed
back before the run as the control on the patch itself. Clean tree at `315f7f620`, one-line bypass,
nothing else.

```
thread 'api::tests::a_refs_read_costs_one_authority_open_per_publication' panicked at
crates/kin-daemon/src/api.rs:22854:9:
assertion `left == right` failed: eight refs reads at one publication must open the durable
authority zero further times; each open reads the whole publication and re-verifies every CAS body
  left: 9
 right: 1
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1394 filtered out; finished in 5.22s
```

Control on the restored tree:

```
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1394 filtered out; finished in 8.63s
```

`left: 9, right: 1` reproduces the reported 1 to 9 exactly, without the debug print the body's own
tail carried. `git status --porcelain` empty before, after the restore and at the end.

The test is in the right home: a plain `#[tokio::test]` behind no feature gate, so the nextest shards
that grade a pull request run it, rather than acceptance, which kin skips on pull requests.

## Verdict

NO-GO at this head, on the two body items above, neither of which needs a code edit. The code itself
clears every refusal condition. Fix the after numbers and the production-arm framing and this is a
GO on the same sha.

PR #1509, opened 2026-09-05
Answer status, diff and log from one authority open

troyjr4103:
Journey GAP-6 and GAP-8, from `.kin-coord/reports/showhn-journey-20260904.md`.

## The defect

An authority open re-verifies every persisted body, so it costs whatever the whole store is worth. On an imported 470 MiB express repository the journey lane measured, back to back at load 54:

```
kin log -n 1        9.567 s   then   8.983 s
kin status         35.957 s
kin diff           31.599 s
kin graph status    0.172 s   (daemon-served)
kin locate          1.149 s   (daemon-served)
```

Counted from source at the base sha, the reason is that these three verbs open the whole store two and three times per invocation:

| Verb | With a daemon | Without one |
|---|---|---|
| `kin status` | 1 in the CLI (`merge_in_progress`) + 2 in the daemon's `command_status` | 2 in the CLI |
| `kin diff` | 1 in the CLI (`content_lines_for`) + 1 in the daemon's `command_diff` | 2 in the CLI |
| `kin log` | 1 in the CLI; `log::run` never asked the daemon at all | 1 in the CLI |

The daemon already had a per-publication authority cache sitting beside those handlers, `command_repository_authority` in `api.rs`, and none of the three used it. Routes already existed for both unused verbs: `POST /commands/log` and `POST /commands/diff`. Nothing needed inventing.

The journey lane also measured the same verbs with the graph section freshly materialized (log 31.9 s, status 72.7 s, diff 21.0 s at load 70 to 120). They did not move: the open costs about 12.7 s with nothing to fold. Refreshing the section, which another lane is doing, removes an extra fold on top of an open that is expensive on its own.

## The change

Each verb opens once and asks that one open for everything it needs, through new helpers that take an already-open authority: `status::inspect_at`, `status::merge_in_progress_at`, `status::workspace_tip_at`, `log::inspect_at`, `log::build_log_response_at`, `diff::inspect_with_endpoint_entities_at`, `diff::build_diff_response_at`. The binding-taking wrappers stay, because a one-shot caller with nobody to share an open with is what they are for.

The daemon's `command_status`, `command_diff` and `command_log` take `command_repository_authority`. And `kin log` asks `POST /commands/log` before opening anything, the way `kin diff` already asked `/commands/diff`.

Two response envelopes grew a field so the caller's own open count reaches zero when a daemon answers. Neither report schema moves: `StatusReportWire` denies unknown fields, so a key added there makes an older CLI reject a newer daemon outright and is a v4 decision to take deliberately. The envelopes have never denied unknown fields and every field beside `report` already defaults.

- `CommandStatusResponse` carries `merge`, `workspace_tip` and `authority_readings_taken`.
- `DiffResponse` carries `artifact_content_rendered`; `DiffRequest` carries `full_bodies`.

Both flags exist because "this build did not read it" and "there is nothing to read" are different facts and only the first is a reason for the caller to open the store. An older daemon sends neither, which sends the caller down one local open rather than being believed, so a new CLI against an older daemon still prints the merge banner instead of a clean tree over a parked merge.

Every verb now names which open answered it, in three arms, because a daemon that answered and still cost a local open is neither of the other two.

## GAP-8: a workspace its branch has moved past

After a push into a store with a workspace, `kin status` and `kin log` described the change the workspace held and said nothing about the branch that had moved. New `commands/workspace_tip.rs` names five states and prints one line in every one of them, because a detached head and an unborn branch both read as agreement when nothing is printed.

The reading comes from authority metadata alone. `ChangeMap` decodes on first touch and on a converted repository it is most of the snapshot body, so a status that walked history to count how far behind it was would have re-imported the cost this PR removes. Only `kin log` claims a distance in changes, because only `kin log` has already walked that DAG; a reading with no walk behind it says the tips differ rather than inventing a count.

## Tests, and why they bound a count rather than a clock

New tests that can fail:

- `commands::status::tests::one_status_reading_opens_repository_authority_once`
- `commands::log::tests::one_local_log_opens_repository_authority_once`
- `commands::diff::tests::one_diff_answer_opens_repository_authority_once`
- `commands::status::tests::an_older_daemons_status_payload_says_it_took_no_authority_readings`
- `commands::diff::tests::an_older_daemons_diff_payload_says_it_rendered_no_artifact_content`
- six cases in `commands::workspace_tip::tests`

Each open-count test asserts `kin_core::authority_opens()` moved by exactly one, with a non-vacuity control first: the report must be real and the workspace-tip reading must not be `Unknown`, because "one open and both readings taken" and "one open and a reading missing" would otherwise read identically.

The assertion is on the count, not on elapsed time, and this repository already argues that in its own words at `crates/kin-daemon/src/api.rs`: "a timing assertion on a four-file fixture would pass just as readily with the defect present". A CI fixture answers in milliseconds whether it opens once or three times, so any ratio bound loose enough to survive a loaded runner passes with the defect in place. The wall clock belongs in the measurement below, on a real store.

## Measurement, before

Scratch `HOME` and `KIN_HOME` under `/tmp`, never the primary store. `expressjs/express` at tag `v5.2.1` (`dbac741a`), 218 files, imported with `kin init .` (1285 s on this box), then one `kin commit` because that is what invalidates the graph section and puts the store in the shape the defect was found in. Store 484 MiB under `.kin/`. `KIN_EMBED_BACKEND=cpu`. Binaries are the base sha's own release build.

Cold pass then warm pass; the warm number is the one that counts. `cli_opens` counts `opening persisted repository authority` on the CLI's own stderr under `RUST_LOG=warn,kin_core=info`; `daemon_opens` counts new lines of the same in `.kin/daemon.log` across the invocation. Load 78 to 94 with six Rust builders held on the box throughout, so every wall time is an upper bound and the ratio between the verbs and `kin graph status` at the same moment is what to read.

```
BEFORE, warm
verb             wall_s  cli_opens  daemon_opens  rc
log              23.492          1             0   0
status           85.036          2             2   0
diff             51.045          1             2   0
graph-status      0.304          0             0   0
```

The counts match what source said before anything ran. `kin graph status` answers in 0.304 s having opened repository authority zero times anywhere. `kin log` never asks the daemon. `kin status` spends 85 s across four whole-store opens. Every unit of that gap is an open that did not have to happen.

## Gates

Run locally in the lane worktree through the fleet's builder slot, on the tree at this head:

```
$ cargo fmt --all -- --check
fmt rc=0

$ cargo test -p kin-cli --lib -- commands::status::tests:: commands::log::tests:: \
    commands::diff::tests:: commands::workspace_tip::tests::
test result: ok. 49 passed; 0 failed; 0 ignored; 0 measured; 2128 filtered out; finished in 6.75s

$ cargo check -p kin-daemon --lib
daemon rc=0
```

Clippy is graded by hosted CI rather than locally: on this box a local clippy as `ci.yml` runs it is a full workspace check build and the builder cap is binding.

### DIFF
--- a/crates/kin-cli/src/commands/diff.rs
+++ b/crates/kin-cli/src/commands/diff.rs
@@ -30,6 +30,14 @@ pub struct DiffRequest {
     pub head: Option<String>,
     #[serde(default)]
     pub json: bool,
+    /// Whether the caller wants whole bodies rather than the trimmed content
+    /// rows, so the server can read CAS on the caller's behalf.
+    ///
+    /// `#[serde(default)]` because this crosses the daemon wire and an older
+    /// peer neither sends nor understands it; a daemon that ignores it answers
+    /// exactly as it did, and the caller's own fallback fills the content in.
+    #[serde(default)]
+    pub full_bodies: bool,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -38,6 +46,15 @@ pub struct DiffResponse {
     pub lines: Vec<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub report: Option<DiffReport>,
+    /// Whether this responder read artifact content into `report`.
+    ///
+    /// A flag rather than an empty-vector test, because "this build did not
+    /// read content" and "the content rows are genuinely empty" are different
+    /// facts and only the first is a reason for the caller to open the whole
+    /// store and read CAS itself. `#[serde(default)]` makes an older peer's
+    /// silence read as the first, which is what it means.
+    #[serde(default)]
+    pub artifact_content_rendered: bool,
 }
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
@@ -255,6 +272,21 @@ pub fn inspect_with_endpoint_entities(
     live: Option<&kin_db::InMemoryGraph>,
 ) -> Result<(DiffReport, DiffEndpointEntities)> {
     let authority = ActiveRepositoryAuthority::open(binding)?;
+    inspect_with_endpoint_entities_at(&authority, base, head, live)
+}
+
+/// The same pair, from an authority the caller already opened.
+///
+/// An open re-verifies every persisted body, so a caller that wants the report
+/// AND the artifact content behind it takes one open and asks it for both.
+/// Reaching for the binding-taking wrapper twice is what made a local `kin diff`
+/// on a converted 470 MiB store pay for two whole-store opens.
+pub fn inspect_with_endpoint_entities_at(
+    authority: &ActiveRepositoryAuthority,
+    base: Option<&str>,
+    head: Option<&str>,
+    live: Option<&kin_db::InMemoryGraph>,
+) -> Result<(DiffReport, DiffEndpointEntities)> {
     let lease = authority.manager().read_authority();
     let metadata = lease.metadata();
     let workspace = metadata
@@ -771,7 +803,8 @@ pub async fn run(
     // local answer, which names its own gap rather than printing a zero that
     // reads like an answer.
     if !json && workspace_endpoint {
-        if let Some(response) = daemon_diff(&layout, &base, &head).await {
+        if let Some(response) = daemon_diff(&layout, &base, &head, full_bodies).await {
+            let content_rendered = response.artifact_content_rendered;
             for line in response.lines {
                 println!("{line}");
             }
@@ -780,12 +813,25 @@ pub async fn run(
                 //
                 // The daemon renders its own header lines and this appends after
                 // them, so its output stays byte-identical and the wire payload
-                // is unchanged. The report it already returns carries both blob
-                // hashes, so the CLI opens its own binding and reads CAS here.
-                // Two renderings of a content diff would drift, and the one
-                // nobody runs would be the one that drifts.
-                for line in content_lines_for(&layout, report, full_bodies) {
-                    println!("{line}");
+                // is unchanged. Two renderings of a content diff would drift,
+                // and the one nobody runs would be the one that drifts.
+                //
+                // The bodies come from the responder's own authority when it
+                // read them, and only from a local open when it did not. That
+                // second arm is a peer too old to carry the rows, and it is the
+                // whole of what used to happen every time: the answer arrived
+                // from a warm daemon and the command then re-verified the entire
+                // store to print the bodies beneath it.
+                if content_rendered {
+                    for row in &report.artifact_content {
+                        for line in render_content_lines(row) {
+                            println!("{line}");
+                        }
+                    }
+                } else {
+                    for line in content_lines_for(&layout, report, full_bodies) {
+                        println!("{line}");
+                    }
                 }
                 if let Some(line) = semantic_scope_line(report.semantic_basis.as_ref()) {
                     println!("{line}");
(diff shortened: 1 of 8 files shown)


troyjr4103:
## Measurement, after

Same express store, byte-copied after the before arm's commit, so both arms read identical durable authority. Binaries built from this branch.

```
AFTER, warm
verb             wall_s  cli_opens  daemon_opens  rc
log               0.014          0             0   0
status            0.304          0             0   0
diff              2.846          0             0   0
graph-status      0.047          0             0   0
```

Zero authority opens, in the CLI and in the daemon, across the whole warm pass of all four verbs. That is the count the tests bound and the property that stays flat as a store grows.

Before and after together:

```
verb            before_s  after_s   before opens  after opens
log               23.492    0.014              1            0
status            85.036    0.304              4            0
diff              51.045    2.846              3            0
graph-status       0.304    0.047              0            0
```

**The load caveat, stated rather than buried.** The before arm ran at load 78 to 94 and the after arm at 65, so raw ratios flatter the after arm. `kin graph status` is the control for exactly that: it is measured in each arm at the same moment as the other verbs on that arm, and it moved 6x on load alone. So read the ratio to it inside each arm, which is load-independent:

```
verb            before_s   x graph status   after_s   x graph status
log               23.492            77.3      0.014             0.3
status            85.036           279.7      0.304             6.5
diff              51.045           167.9      2.846            60.6
graph status       0.304             1.0      0.047             1.0
```

`kin status` went from 280 times the cost of a daemon-served read to 6.5 times. `kin log` went from 77 times to less than one, because it now answers from the same daemon and returns less.

`kin diff` is still 61 times graph status, and that is honest rather than fixed. Its remaining 2.8 s is not an authority open, because the count says zero: it is the admission pass `kin diff` takes before reading, which walks the working tree on purpose, plus the daemon's own diff computation. Removing an open was this change's job and the open is gone; what is left is real work.

The new lines, verbatim from that run:

```
Head: detached change 995cfc12bc905504efae74a96a5c063390848ce3725802651d64eb40187cfce0
Workspace tip: this head names a change directly, so it follows no branch; `kin branch switch <branch>` attaches it
Answered by: this repository's daemon, from the authority it already holds open for the current publication
```

## Falsification

Each open-count test was run with the exact second open it names put back, and each goes red on the count.

```
mutated status.rs / log.rs / diff.rs
one_local_log_opens_repository_authority_once   ... FAILED   left: 2  right: 1
one_diff_answer_opens_repository_authority_once ... FAILED   left: 2  right: 1
one_status_reading_opens_repository_authority_once ... ok
```

The status mutant survived that first pass, and the reason was the mutation rather than the test. `read_status_once` carries two lines reading `merge: merge_in_progress_at(&authority),` at identical indentation: one on the arm for a daemon too old to carry the readings, one on the local-open arm. A first-occurrence string replacement lands on the daemon arm, which a test running with `KIN_DAEMON_URL` unset and no daemon never enters. Re-aimed at the local-open arm, anchored on the whole `StatusReading` block including `source: AuthoritySource::OwnAuthorityOpen` which only that arm carries:

```
mutated the local-open arm of read_status_once
one_status_reading_opens_repository_authority_once ... FAILED
  assertion `left == right` failed: one `kin status` must open repository authority once and
  ask that open for the report, the merge and the workspace tip; opening per reading is GAP-6
  left: 2
 right: 1
restored status.rs
```

All three go red under their named mutation and green again on restore (51 passed, 0 failed).

## What CI caught, both times

Two red rounds, two independent findings, and both are recorded because the second one is easy to miss.

1. `verify-zero-file-search.py` counts the spelling `.metadata()` per file against a number the allowlist pins, and the workspace-tip reading added a third to status.rs. All three read an authority lease and none touches a filesystem, so the gate counts a spelling rather than a violation. Fixed by reading the workspace through `ActiveRepositoryAuthority::workspace` rather than by bumping the pinned number.
2. Clippy on the lib target: `function render_text is never used`. Steps 13 through 20 of that job, Clippy among them, are skipped once step 12 fails, so the first round said nothing at all about clippy. Both failing test shards reported the identical `could not compile kin-cli (lib)` at the step named "Assert the selection lists tests before running any", so no test ran and no test failed; all four reds were that one error.
3. `every_flag_a_doctor_fix_line_names_exists_in_the_cli` then caught a real defect in my own prose: the gap line said `kin daemon restart`, and `kin daemon` has only status, stop and sweep. A line whose whole job is to name the next command named one that does not run. That check walks the real clap tree and it earned its keep here.

PR #1510, opened 2026-09-05
Write the graph section a commit moves past, and wait for the first embed pass

troyjr4103:
Two defects the cold-user journey found on the first-run path, both measured before either was
touched.

## GAP-4: a first commit left the graph section unwritten, and the checklist called it a host limit

`kin init` materializes the workspace base graph section and nothing else did. A repository built
with Kin therefore had no section after its first commit, and `kin doctor` said so like this, on a
store holding one change:

```
  ✗ Graph section              DEGRADED       absent, so every open of this store folds this workspace's base out of history, at most the 1 changes this store holds, the base's first-parent ancestry within them. `kin graph materialize` writes one
✗ 3 checks need attention: VFS projection, Semantic query readiness, Graph section. ... Degraded rows are host limits: Graph section.
```

An admitted repository pays far more for the same absence. On a 470 MiB express store, one commit
refused the section init had written, and every open after it reported `authority_open=37922ms` with
`folded_changes=3834`; materializing a fresh section dropped that daemon's startup from 41.7 s to
13.7 s.

**Writer.** `commit_native_plan_with_working_copy_proof` now refreshes the section after the
transaction validates, through the authority the commit already holds, so it pays no second
O(store) open. Commit and amend both reach it.

No size bound, deliberately. The obvious shape is to leave large stores alone, because kin-db's
writer recomputes the fold from history and its doc says ordinary publish does not pay that capture.
That shape is wrong: the fold is paid once either way. A store whose section a commit invalidated
pays it at every open forever; a store whose commit refreshed it pays it once and opens clean until
the next commit. A daemon restarts more often than a person commits, and an open blocks the first
question a session asks while a commit is a heavy operation the caller already chose to wait for.
Never fatal: a memoization that did not persist is a slower next open, not a failed commit.

**Reader.** The `Graph section` row moves from `HealthStatus::Degraded` to `HealthStatus::Stale`.
`Degraded` is documented on the enum as "a real shortfall in the machine or container Kin was asked
to run on", and `setup.rs` prints "Degraded rows are host limits" from it. `Stale` keeps the same
roll-up behaviour for this row and says the true thing.

## GAP-5: the first agent call after `kin init` errored, and the embedding pass it waited on was gone

Two halves, and the first one's mechanism is not what the journey report inferred.

**The daemon.** The journey read this as an idle shutdown inside the 80 s window. It is not.
`kin init` stops the daemon itself: `enrich_after_init` calls `stop_conversion_daemon` as soon as
the sweep phase ends. Measured on a synthetic 180-file Git import, `kin init` rc 0 wall 134 s, there
is no daemon at t+0s, t+5s or t+15s (`recorded_pid=none port=none alive=no`) while `kin doctor` on
the same store reports `Daemon idle window ok 180s`. The store hands over at
`Embeddings: 0/1800 indexed (1800 pending)`. `stop_conversion_daemon`'s own doc says it waits for
the sweep "never mid-pass"; the first embedding pass was given no such wait.

`enrich_after_init` now waits for that pass to settle before the stop, and only when this phase
started the daemon. The wait ends when the queue drains, when nothing in that daemon will drain it,
or at a bounded budget, printing where it reached and that the next daemon resumes from there.

The captain's preferred shape was to leave the daemon running instead and let the idle monitor hold
it (`daemon.rs` already keeps a daemon alive while `embed_work_in_flight`). That is the better end
state and it is written up as a follow-up rather than shipped here, because `init.rs` records in
code why the stop exists: a daemon left behind by `kin init` pins a repository cursor that makes the
next mutation refuse, and the incident behind that comment is a CI runner where a leaked conversion
daemon blocked an independent daemon two minutes later. Reversing it needs a test on a converted
store that this change does not have.

One correction inside that fix, recorded because it changes what the wait can promise. The daemon
stands its embed worker down under three conditions and `EmbedRuntimeState` carries only two of
them: `KIN_DAEMON_AUTO_EMBED=0` pauses the worker and nothing on the wire says so. The first version
of this wait would therefore have sat for its whole budget on every `kin init` on an opt-out host.
It now treats a reading as evidence of life when the indexed count moved, or the embed mutex is
held, or the model is still downloading, and calls the pass stalled after twenty seconds of none of
those. The model arm has its own control test.

**kin-mcp.** The first graph-reading `tools/call` of a session now waits 45 s for the launcher's
startup binding; every later call keeps the 10 s it had. The first call is the one that admits the
daemon spawn, so it is the only one that can be paying for a cold open. 45 s clears the measured
cold bind with margin (express wanted about 25 s, and even after the section fix a 13.7 s startup
still defeated the 10 s grace on the first two calls) and stays under the 60 s per-call timeout
common MCP clients use. It is a ceiling and not a delay: `wait_until_settled` returns the instant
the binding settles, and a control run against a store that opens in 1.82 s answered its first call
in 1.82 s.

`starting_report` now states the bound the call actually waited, so two reports over different waits
are not the same sentence, and `kin setup`'s per-client MCP round-trip budget is derived from the
grace rather than restated beside it: a budget shorter than the grace kills the client mid-wait and
reports a deadline where the server was about to hand back an accurate account.

One thing a reviewer should check rather than take from me. The old constant's comment said it was
"kept ... under the update watchdog's 30 s whole-probe budget", and I could not find that budget
anywhere in this repository, so I did not carry the claim forward. If it lives in kin-infra's
watchdog rather than here, the first-call grace is the number to weigh against it.

## Tests, and how each was falsified

| Test | Falsified by |
|---|---|
| `a_commit_leaves_the_workspace_base_graph_section_written` (kin-daemon) | removing the `refresh_workspace_base_graph_section` call: both commit assertions report `Folding` |
| `a_folding_store_is_reported_...` (kin-cli health) | restoring `HealthStatus::Degraded` on the folding arm |
| `a_backlog_nothing_will_drain_is_not_a_pass_worth_waiting_for` (kin-cli init) | making `first_embed_pass_standing` always answer `Settled` |
| `the_wait_holds_while_the_pass_is_filling_and_returns_when_it_drains` (kin-cli init) | returning from the loop on its first reading |
| `a_pass_that_never_drains_ends_at_the_budget_and_says_where_it_reached` (kin-cli init) | making the loop exit on its first reading; the test reads the elapsed clock, not only the outcome |
| `a_queue_that_never_moves_stops_being_waited_on_before_the_budget` (kin-cli init) | removing the no-progress stall branch |
| `a_queue_waiting_on_the_model_download_is_still_a_pass` (kin-cli init) | the control for the arm above: a stall guard that ignored the model fetch fails it |
| `the_wait_returns_at_once_when_no_daemon_is_serving_the_store` (kin-cli init) | waiting when no daemon answers |
| `only_the_call_that_starts_the_daemon_gets_the_long_wait` (kin-mcp) | giving every call the same grace |
| `the_first_call_bound_clears_the_measured_cold_bind_and_stays_under_a_client_timeout` (kin-mcp) | lowering the bound to 25 s or raising it to 60 s |
| `a_first_call_that_outlasts_its_bound_reports_the_bound_it_waited` (kin-mcp) | dropping the bound from `starting_report` |
| `the_starting_report_states_the_bound_this_call_waited` (kin-mcp) | the same |
| `the_client_budget_outlasts_the_wait_the_mcp_server_is_allowed_to_take` (kin-cli setup_verify) | restating 20 s instead of deriving it |

## What ran locally, and what did not

Clippy was not run locally, by the captain's speed rule: the box's six builder slots were all held
with six lanes queued, and hosted CI grades clippy on a 16-core runner. Everything below ran through
`.kin-coord/bin/heavy-slot.sh journeygaps kin`, against the worktree
`.kin-coord/worktrees/journeygaps-kin` on `chore/lane-journeygaps-kin`.

```
$ cargo fmt --all -- --check
fmt rc=0

$ cargo test -p kin-mcp --lib
test result: ok. 897 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 28.09s
kin-mcp rc=0

$ cargo test -p kin-cli --lib -- setup_verify
test result: ok. 23 passed; 0 failed; 0 ignored; 0 measured; 2152 filtered out; finished in 8.22s
kin-cli-sv rc=0

$ cargo test -p kin-daemon --lib -- a_commit_leaves_the_workspace_base_graph_section_written
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1394 filtered out; finished in 3.37s
kin-daemon rc=0
```

The `kin-cli` init-wait and health-row tests, and every falsification arm in the table above, run in
the next slot; their output goes into the lane report at
`.kin-coord/reports/showhn-journeygaps-20260905.md` and a comment here.

## Follow-up, not in this change

Merge, rollback, checkout and branch switch also move the workspace base and leave the section
refused. Each holds an open authority at its publication site, but each also carries a
`LocalRepositoryAuthorityFreeze` whose exclusive storage lock is held past that point, and the
section refresh is a representation rewrite that takes the publication lock. The commit path has no
such guard, which is why it goes first. Written up with file and line references in the lane report.

### DIFF
--- a/crates/kin-cli/src/commands/health.rs
+++ b/crates/kin-cli/src/commands/health.rs
@@ -5234,13 +5234,24 @@ fn graph_section_row_for_unread_graph(
 /// Turn the section state into a row, split from its fetch so every branch is
 /// testable without a daemon.
 ///
-/// `Degraded` for a folding store, and the severity is chosen rather than
+/// `Stale` for a folding store, and the severity is chosen rather than
 /// inherited. A store that folds is completely correct; it is only paying a full
 /// history replay at every open, which on the kin store was 47 seconds of a 95
 /// second one. `Missing` and `Misconfigured` fail the whole report, and a store
-/// that answers every question correctly must not do that. `Degraded` costs it
-/// the all-clear and nothing else, which is exactly the claim the evidence
-/// supports, and it names the one command that changes it.
+/// that answers every question correctly must not do that. What is wanted is a
+/// status that costs the all-clear and nothing else.
+///
+/// Two statuses do exactly that, and the row carried the wrong one. `Degraded`
+/// is documented on the enum as "a real shortfall in the machine or container
+/// Kin was asked to run on", and `setup.rs` reads it at its word: its closing
+/// line prints "Degraded rows are host limits: Graph section." So a first
+/// `kin commit` in a fresh repository told its owner that a memoization nobody
+/// had written yet was a property of their machine, over a store holding one
+/// change, with an internal repair command beside it (journey GAP-4). `Stale`
+/// carries the same roll-up behaviour for this row, keeping it out of `Ready`
+/// through `needs_attention` and out of `Failing` through `blocks_readiness`,
+/// and it says the true thing: this store's acceleration is not current. The
+/// row still names the one command that changes it.
 ///
 /// A refused section is reported with kin-db's own refusal word because that is
 /// the word its own log carries, so an operator comparing this row against
@@ -5254,9 +5265,14 @@ pub(crate) fn graph_section_health(
     let status = match state.standing {
         kin_core::graph_section::GraphSectionStanding::Serving
         | kin_core::graph_section::GraphSectionStanding::Unborn => HealthStatus::Healthy,
-        kin_core::graph_section::GraphSectionStanding::Folding => HealthStatus::Degraded,
+        kin_core::graph_section::GraphSectionStanding::Folding => HealthStatus::Stale,
         // Never healthy. A state that could not be read is not a state that is
         // fine, and rendering it as one is the invisibility this row replaces.
+        // It shares `Stale` with a folding store, and the two stay separate
+        // where `kin_core::graph_section` keeps them separate, in the sentence:
+        // one says the store folds and how big the fold is, the other says the
+        // state could not be read and why. What the module refuses is rendering
+        // either as a clean bill, and neither is one.
         kin_core::graph_section::GraphSectionStanding::Unknown => HealthStatus::Stale,
     };
     let check = HealthCheck::new(ID, LABEL, status, state.doctor_detail());
@@ -7188,9 +7204,15 @@ mod tests {
     fn a_folding_store_loses_the_all_clear_without_failing_the_report() {
         let row = graph_section_health(&folding_section_state());
         assert!(
-            matches!(row.status, HealthStatus::Degraded),
-            "a folding store is degraded, not broken: {:?}",
+            matches!(row.status, HealthStatus::Stale),
+            "a folding store's acceleration is out of date, not broken, and not a host limit: \
+             {:?}",
             row.status
+        );
+        assert!(
+            !matches!(row.status, HealthStatus::Degraded),
+            "`Degraded` is the host-shortfall status, and `setup.rs` prints `Degraded rows are \
+             host limits`; a store that one `kin graph materialize` fixes is not a host limit"
         );
         assert!(
             !blocks_readiness(&row),
--- a/crates/kin-cli/src/commands/init.rs
+++ b/crates/kin-cli/src/commands/init.rs
@@ -443,6 +443,294 @@ async fn stop_conversion_daemon(borrowed_existing: bool, kin_root: &Path) {
     }
 }
 
+/// How long the conversion phase waits for the first embedding pass to settle
+/// before it stops the daemon it started.
+///
+/// [`stop_conversion_daemon`] already waits for the sweep, and its own doc says
+/// why: "a sweep killed halfway is the failure this whole area exists to
+/// prevent". The first embedding pass was given no such wait, so a conversion
+/// that queued a backlog stopped the only process that could drain it, and the
+/// daemon a user's next command started began that pass from zero. On the
+/// journey run of 2026-09-04 the store handed over reported `Embeddings: 0/804
+/// indexed (804 pending)` and the session's first semantic query ranked on
+/// lexical fallback.
+///
+/// Bounded for the reason `ENRICH_BUDGET` is bounded, and much more tightly: the
+/// pass is resumable, so what this cuts short the next daemon continues from a
+/// persisted checkpoint rather than from zero, and a conversion that waits
+/// without end on a very large first fill is worse than one that hands over a
+/// store still filling and says so.
+///
+/// Sized from the pass, not from a feeling. Measured on 2026-09-05 on a
+/// 180-file Python repository admitted through `kin init`, whose store queues
+/// 1,800 embeddings: a daemon drained 880 of them in 21 seconds on the CPU
+/// backend with the host carrying six concurrent Rust builders, so the whole
+/// queue from zero is around 45 seconds on that shape, and the 804 the journey
+/// run left on express would be under 20. Two minutes is that with well over
+/// double the margin, and it stays under half of what an import of this size
+/// already costs: this one took 134 seconds and express took 296.
+const FIRST_EMBED_PASS_BUDGET: std::time::Duration = std::time::Duration::from_secs(120);
+
+/// How often the wait asks the daemon where the pass has reached.
+const FIRST_EMBED_PASS_POLL: std::time::Duration = std::time::Duration::from_secs(2);
+
+/// How long a pass may report no progress before this wait stops calling it a
+/// pass.
+///
+/// The daemon stands its embedding worker down for reasons this reading cannot
+/// see. `KIN_DAEMON_AUTO_EMBED=0` pauses it outright, and
+/// `background_embed_worker_can_drain` gates on that pause, on a permanently
+/// failed worker and on whether embed progress can be persisted at all, while
+/// `EmbedRuntimeState` carries only the last two of those three. So a queue on
+/// an opt-out host is `Filling` by every field this wait can read, and without
+/// this the conversion would sit here for the whole budget on a store where
+/// nothing was ever going to be embedded.
+///
+/// Answered by progress rather than by a flag, which is what makes it cover the
+/// case nobody has thought of yet: a pass that is neither indexing, nor
+/// downloading its model, nor holding the embed mutex is not a pass. The window
+/// is not a race against how long a drain takes, because a running drain moves
+/// the indexed count on every batch and resets it: the 880 embeddings measured
+/// on 2026-09-05 took 21 seconds in total and would have reset this many times
+/// over. It only has to be longer than the gap between two batches.
+const FIRST_EMBED_PASS_STALL: std::time::Duration = std::time::Duration::from_secs(20);
+
+/// How often the wait says where the pass has reached.
+///
+/// A conversion that has already taken minutes must not then go silent for two
+/// more, and a line every poll is a wall nobody reads. Fifteen seconds puts at
+/// most eight lines under a wait that runs to its budget, and one under a wait
+/// that ends in a few seconds.
+const FIRST_EMBED_PASS_REPORT_EVERY: std::time::Duration = std::time::Duration::from_secs(15);
+
+/// What the conversion phase should do about the embedding pass it started,
+/// read from one daemon reading.
+///
+/// Split from the poll for the reason every other reader in this file is split
+/// from its fetch: each branch has to be provable without a daemon, an
+/// embedding model or a network.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum FirstEmbedPassStanding {
+    /// Nothing is queued. Stopping the daemon now costs nothing.
+    Settled,
+    /// Work is outstanding and something in this daemon will drain it. This is
+    /// the state the stop must not cut short.
+    Filling,
+    /// Work is outstanding and nothing in this daemon will drain it, so waiting
+    /// buys the store nothing.
+    Stalled(&'static str),
+}
+
+/// Read one daemon reading into a decision.
+///
+/// The stalled arms mirror the daemon's own `background_embed_worker_can_drain`
+/// predicate rather than a second copy of its rule: a backlog nobody will
+/// consume is not work in flight, and treating it as such would make every
+/// conversion on such a store sit here for the whole budget.
+fn first_embed_pass_standing(
+    embed: &crate::commands::resources::EmbedRuntimeState,
+) -> FirstEmbedPassStanding {
+    if embed.embed_worker_failed {
+        return FirstEmbedPassStanding::Stalled(
+            "the daemon's background embedding worker has stopped",
+        );
+    }
+    if embed.embed_persistence_unavailable {
+        return FirstEmbedPassStanding::Stalled(
+            "this store's graph authority carries no durable local vector sidecar, so nothing \
+             embeds here",
+        );
+    }
+    if embed.embeddings_pending == 0 {
+        return FirstEmbedPassStanding::Settled;
+    }
+    FirstEmbedPassStanding::Filling
+}
+
+/// What one wait for the first embedding pass ended in.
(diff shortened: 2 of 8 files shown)


troyjr4103:
## The regression this branch caused, and the evidence it is gone

`4c3d9afec` turned six kin CI tests red, every one a `kin-cli` init test, with
`init_finishes_its_store_when_the_reader_closes_the_pipe` asserting exit 0 and getting 7,
`EXIT_ENRICHMENT_UNATTESTED`. The cause is in this branch's own output in the shard log:
`→ first embedding pass: 0 of 18 indexed` nine times and `0 of 256 indexed` eighteen times, which is
the fifteen-second progress cadence over the whole 120 s budget with the counter never moving once.

`crates/kin-cli/tests/common/mod.rs:1047` gives every spawned `kin` the test's own temp `HOME`, so
`$HOME/.cache/huggingface` is empty in those tests on any machine. The pass sits inside a 522 MiB
fetch it cannot finish, the first version of the wait read a download in progress as progress, and
`kin init` held a daemon that loads the model resident for two minutes while several of those tests
ran at once. The daemon dies; `kin_daemon_spawn::settle_unwatched_daemon_death_with` grades that
death a kill because the serving record survives beside a dead pid with no matching death note; and
`exit_code_for` returns 7.

`2c2eb5eb1` does not run the wait at all when the model is absent and a fetch applies. Because the
tests set their own empty `HOME`, a local run reproduces the CI arm rather than the warm one, which
is what makes the run below evidence:

```
$ cargo test -p kin-cli --test init_non_tty_output --test git_export_roundtrip --test live_graph_durability_disclosure
test the_enrichment_note_carries_a_cause_the_daemon_reported ... ok
test a_captured_init_writes_phase_summaries_and_no_progress_frames ... ok
test init_finishes_its_store_when_the_reader_closes_the_pipe ... ok
test the_conversion_phase_leaves_no_daemon_behind_when_it_exits_early ... ok
test git_export_round_trips_to_plain_git ... ok
test mcp_status_and_locate_disclose_live_only_entities_and_then_stop_once_a_commit_records_them ... ok
test result: ok. 10 passed; 0 failed  (tests/git_export_roundtrip.rs, 112.86s)
test result: ok. 13 passed; 0 failed  (tests/init_non_tty_output.rs, 13.29s)
test result: ok. 16 passed; 0 failed  (tests/live_graph_durability_disclosure.rs, 10.50s)
integration rc=0

$ grep -c "first embedding pass" <that run's output>
0
```

Zero. The wait's own lines are absent from the run, against twenty-four of them in the CI shard log,
so the thing that was holding those daemons is not running any more. `init_non_tty_output` also
drops from about 131 s per test on the red run to 13.29 s for the whole suite.

```
$ cargo test -p kin-cli --lib -- commands::init::tests
test commands::init::tests::a_machine_that_has_not_fetched_the_model_is_not_waited_on ... ok
test commands::init::tests::a_queue_on_a_machine_that_has_the_model_is_still_a_pass ... ok
test result: ok. 50 passed; 0 failed; 0 ignored; 0 measured; 2126 filtered out; finished in 1.85s

$ cargo fmt --all -- --check
fmt rc=0
```

## Every new behaviour, broken on purpose

Each arm mutates one line, runs that behaviour's own test, and restores the file from a saved copy.
Run at `4c3d9afec`; the two model tests above were added by `2c2eb5eb1` and their arms are the last
two rows.

| Arm | What was broken | Result |
|---|---|---|
| `gap4writer` | deleted the `refresh_workspace_base_graph_section` call from `commit_native_plan_with_working_copy_proof` | RED |
| `gap4reader` | put `HealthStatus::Degraded` back on the folding arm | RED |
| `gap5standing` | made `first_embed_pass_standing` always answer `Settled` | RED |
| `gap5budget` | made the wait exit on its first reading (`if began.elapsed() >= budget` to `if true`) | RED |
| `gap5stall` | removed the no-progress stall branch | RED |
| `gap5mcpgrace` | gave every `tools/call` the same grace | RED |
| `gap5mcpreport` | dropped the waited bound from `starting_report` | RED |
| `setupbudget` | restated `PER_CLIENT_BUDGET` as 20 s instead of deriving it | RED |

Tails, one per arm:

```
gap4writer   test api::tests::a_commit_leaves_the_workspace_base_graph_section_written ... FAILED
             panicked at crates/kin-daemon/src/api.rs:35536:9:
             assertion `left == right` failed: the first commit must leave a section that answers for the base it just published
             test result: FAILED. 0 passed; 1 failed; 1394 filtered out; finished in 2.50s

gap4reader   test commands::health::tests::a_folding_store_loses_the_all_clear_without_failing_the_report ... FAILED
             panicked at crates/kin-cli/src/commands/health.rs:7206:9
             test result: FAILED. 0 passed; 1 failed; 2174 filtered out; finished in 0.04s

gap5standing test commands::init::tests::a_backlog_nothing_will_drain_is_not_a_pass_worth_waiting_for ... FAILED
             panicked at crates/kin-cli/src/commands/init.rs:2061:9:
             assertion `left == right` failed: a queue with a healthy worker behind it is the state the stop must not cut short

gap5budget   test commands::init::tests::a_pass_that_never_drains_ends_at_the_budget_and_says_where_it_reached ... FAILED
             panicked at crates/kin-cli/src/commands/init.rs:2178:9
             test result: FAILED. 0 passed; 1 failed; 2174 filtered out; finished in 0.02s

gap5stall    test commands::init::tests::a_queue_that_never_moves_stops_being_waited_on_before_the_budget ... FAILED
             panicked at crates/kin-cli/src/commands/init.rs:2217:22
             test result: FAILED. 0 passed; 1 failed; 2174 filtered out; finished in 0.03s

gap5mcpgrace test server::tests::only_the_call_that_starts_the_daemon_gets_the_long_wait ... FAILED
             panicked at crates/kin-mcp/src/server.rs:3978:9: assertion `left == right` failed
             test result: FAILED. 0 passed; 1 failed; 896 filtered out; finished in 0.00s

gap5mcpreport test startup_binding::tests::the_starting_report_states_the_bound_this_call_waited ... FAILED
             panicked at crates/kin-mcp/src/startup_binding.rs:465:9
             test result: FAILED. 0 passed; 1 failed; 896 filtered out; finished in 0.00s

setupbudget  test commands::setup_verify::tests::the_client_budget_outlasts_the_wait_the_mcp_server_is_allowed_to_take ... FAILED
             panicked at crates/kin-cli/src/commands/setup_verify.rs:889:9
             test result: FAILED. 0 passed; 1 failed; 2174 filtered out; finished in 0.02s
```

The `gap5standing` mutation also fails `the_wait_holds_while_the_pass_is_filling_and_returns_when_it_drains`,
which is why that test has no arm of its own: one mutation cannot be attributed to two tests.

The two arms for `a_machine_that_has_not_fetched_the_model_is_not_waited_on` and
`a_queue_on_a_machine_that_has_the_model_is_still_a_pass` are running now and are posted as a
follow-up comment; the head does not depend on them, and I would rather say that than let a table
row stand for a run that has not happened.

Clippy is not run locally on this branch, by the fleet's speed rule: six builder slots held with
eleven lanes queued. Hosted CI grades it.

troyjr4103:
## The two arms the fix round owed

Both run at `2c2eb5eb1`, same shape as the eight above: mutate one line, run that behaviour's own
test, restore the file from a saved copy.

| Arm | What was broken | Result |
|---|---|---|
| `modelabsent` | removed the absent-model refusal, so a cold machine is waited on again (`if !present && no_fetch_reason.is_none()` to `if false`) | RED |
| `modelcontrol` | made the refusal fire unconditionally, which is the over-broad guard the control exists for (`if true`) | RED |

```
modelabsent   test commands::init::tests::a_machine_that_has_not_fetched_the_model_is_not_waited_on ... FAILED
              panicked at crates/kin-cli/src/commands/init.rs:2302:22
              test result: FAILED. 0 passed; 1 failed; 2175 filtered out; finished in 0.05s

modelcontrol  test commands::init::tests::a_queue_on_a_machine_that_has_the_model_is_still_a_pass ... FAILED
              panicked at crates/kin-cli/src/commands/init.rs:2337:9
              test result: FAILED. 0 passed; 1 failed; 2175 filtered out; finished in 0.03s
```

The second arm is the one worth reading. A guard that refused every store would pass the first arm
and would silently undo the whole fix, so the control breaks it in the opposite direction and the
control's own test catches it. Ten arms now, one per new behaviour, all red.

`git status --porcelain` is empty after the run: every arm restores from a saved copy rather than
from git, so nothing uncommitted can be lost to a falsification.

troyjr4103:
## Independent review at `2c2eb5eb132368b1e39f5054fb6dd23b716a4f20`

Reviewed read-only at the exact head, and separately at `4c3d9afec` before the fix. The short
version is that the regression this branch caused is fixed, the fix is isolated to one file, and
that isolation is what settles the one open question from the red round.

**The delta is `crates/kin-cli/src/commands/init.rs` alone**, 96 insertions and 9 deletions.
Nothing in the commit path, the health row, the MCP grace or the setup budget moved between the
red head and this one. That matters beyond tidiness: `mcp_status_and_locate_disclose_live_only_
entities_and_then_stop_once_a_commit_records_them` asserts a post-commit durability property and
failed on the red head, and since not one line of the commit path changed here, a green result on
this head proves that failure was a knock-on of the 120 second wait rather than a GAP-4 defect.

**The fix reads correctly from source.** `first_embed_pass_standing` gains one arm after the
already-drained check, stalling when `!model_fetch.present && model_fetch.no_fetch_reason.
is_none()`. Reading `no_fetch_reason` beside `present` is load bearing rather than decoration: a
remote embedding provider and a model id naming a local directory both report `present: false`
while owing no download, and stalling on those would refuse to wait on exactly the stores where
the pass runs fine.

I checked the one way this arm could be silently wrong, since `present` defaults to `false` and a
daemon reporting a default-constructed `EmbedModelFetch` would kill the feature on every host. It
does not. The daemon builds it per request with `EmbedModelFetch::probe(embed_pass_is_working(&
state))` at `api.rs:4672` and `:4776`, and `probe` computes `present` from `snapshot_present`
against the loader's real cache directory. Genuinely computed.

**The claim that makes the local run count checks out.** The comment above rests on
`crates/kin-cli/tests/common/mod.rs:1047` giving every spawned `kin` the test's own temp `HOME`,
so a local run reproduces the cold-model CI arm rather than a warm one. That line is
`.env("HOME", &self.home_path)` inside `apply_isolated_environment`, beside `KIN_HOME`,
`USERPROFILE` and `XDG_CONFIG_HOME`. So the reported `rc=0` is evidence about the arm that
actually broke. That is a better verification design than most fix rounds manage, because the
local run and the CI run are the same arm by construction rather than by luck.

**The falsification arms are now checkable.** Eight rows, each naming its mutation and carrying a
real failing tail with file and line, including the `gap4writer` arm reddening
`a_commit_leaves_the_workspace_base_graph_section_written` at `api.rs:35536`. The comment is also
honest about its own gaps, naming the two new model arms as still running rather than giving them
a table row, and naming clippy as hosted-graded. That is the right shape.

**Test set.** The old `a_queue_waiting_on_the_model_download_is_still_a_pass` is correctly gone,
since it asserted the behaviour this fix reverses. Its replacement asserts the elapsed clock as
well as the outcome, so a version that stalled only after the 20 second window still fails it, and
`a_queue_on_a_machine_that_has_the_model_is_still_a_pass` stops the model arm from passing over a
wait that refused every store. Both halves present.

### One note, not a blocker

With this arm, a machine that has not got the model does not wait at all, and on a first-ever
`kin init` that is every machine, so such a run still hands its store over with the queue
unfilled. That is what main does today, so this is no worse anywhere and better on a warm machine.
The measurement that motivated GAP-5a was itself a warm machine, since the daemon drained 880
embeddings in 21 seconds on it, so the fix does cover the case that prompted it. Worth keeping out
of the release note that the first run on a fresh machine now waits for its embeddings, because it
does not, and the new stall line says so to the user.

PR #1512, opened 2026-09-05
Carry the sending store's hydration semantics through a transfer

troyjr4103:
Receiving a native transfer no longer costs a store its hydration semantics record. The pack now
declares the sending store's own creation version, and the receiver keeps its record exactly when
the two agree. Two replicas of one build sync as many times as they like and both keep certifying
their answers.

This is journey GAP-10, found on composed main `f6a29e329` and bisected there: `kin init`, commit,
branch create and switch, commit on a branch, fast-forward merge, diverging commits, a conflicted
merge, `resolve --file` plus `--continue` and clone all kept the stamp, and the first push received
took it away for good. From that moment every agent answer on the store carried
`degraded.hydration_semantics_stale` and a verdict of `inconclusive`, and `kin doctor` told a native
store minutes old to "upgrade Kin ... re-ingest the repository into a separate fresh store", which
names nothing a native store can do.

## Measured before the fix

Two scratch native stores on this machine, `kin 0.6.8 (7fb4c394bcd47cb8a2882acbf93546f2bf533ce6)`
built from this branch's clean worktree, scratch `KIN_HOME`, `KIN_EMBED_BACKEND=cpu`, at
2026-09-05T02:22Z. A is `kin init` plus one commit. B is
`kin clone http://127.0.0.1:58081 B --repository <id>` with
`KIN_REMOTE_BEARER_TOKEN=$(cat A/.kin/daemon.token)`. Then B commits and pushes to A, then A commits
and B pulls.

| Point in the run | A (origin) | B (clone) |
|---|---|---|
| A after `kin init` and one commit | `ok ... version 10 at creation` | not created yet |
| after `kin clone` | `ok ... 10` | `ok ... 10` |
| after B commits, before the push | `ok ... 10` | `ok ... 10` |
| **after `kin push` B to A** | **`STALE, records no hydration semantics version`** | `ok ... 10` |
| after A commits and B pulls | `STALE` | **`STALE`** |

The record file itself on the receiver:

```
--- stamp file A-before-push ---
{"schema":"kin.hydration-semantics.v1","created_under":10,"at":"2026-09-05T02:22:03.307633Z"}

--- stamp file A-after-push ---
(absent)
```

And what that store is told to do about it, nineteen seconds after `kin init` created it:

```
! Hydration semantics  STALE  this store records no hydration semantics version, so its persisted
                              history cannot be shown to match the version 10 this build derives
    fix:  upgrade Kin before changing this store; if the newest build still cannot read the record,
          re-ingest the repository into a separate fresh store rather than replacing this one
```

Both transfers moved real history and said so (`Outcome: Committed`, one exact change each), so this
is the admission boundary and not a refusal. The clone kept both records, which is the clone door
below.

## Measured after the fix

Same script, same recipe, `kin 0.6.8 (f87e1ed2a5f15dfa73f2046f5e21c7eb9854d744)` at
2026-09-05T02:37Z. Every row is now `ok ... version 10 at creation`, on both stores, through clone,
push and pull.

| Point in the run | A (origin) | B (clone) |
|---|---|---|
| A after `kin init` and one commit | `ok ... 10` | not created yet |
| after `kin clone` | `ok ... 10` | `ok ... 10` |
| after B commits, before the push | `ok ... 10` | `ok ... 10` |
| after `kin push` B to A | `ok ... 10` | `ok ... 10` |
| after A commits and B pulls | `ok ... 10` | `ok ... 10` |

The timestamps are the proof nothing was rewritten to get there. A's record still reads
`"at":"2026-09-05T02:37:15.298555Z"` after the push and after the pull, and B's still reads
`"at":"2026-09-05T02:37:21.736311Z"`. Both transfers moved history (`Outcome: Committed`, one exact
change each), so this is the same admission boundary that took the record before.

The clone door inverts the same way. Store D, fresh, record rewritten to version 9 so it reads
STALE about itself, cloned into E:

```
--- doctor D-recorded-9 ---
  ! Hydration semantics  STALE  this store records hydration semantics version 9 at creation and
                                this build derives version 10, so the store cannot certify ...
--- doctor E-cloned-from-a-version-9-source ---
  ! Hydration semantics  STALE  this store records no hydration semantics version, so its
                                persisted history cannot be shown to match the version 10 this
                                build derives
--- stamp file E-cloned-from-a-version-9-source ---
(absent)
```

Before the change that same pair produced `ok ... version 10 at creation` on E.

## What changed

The discard was deliberate, not a slip. `kin_core::hydration_semantics`'s own module doc said why
and said what was missing: "the transport carries no version beside them ... carrying the authoring
version on the wire is the phase that gets it back." This is that phase.

`RepositoryTransferPack` gains `source_hydration_semantics: Option<u32>` and the transfer schema
goes 4 to 5. `RepositoryTransferIdentity` gains the field too, so the transfer id binds it and a
proxy can neither forge nor strip a declaration in flight. The two pack builders and
`push_to_remote` take the declaration as a parameter rather than reading it, because the record is a
file beside the store while the authority those functions hold is a storage backend that may not be
a local one at all; only the caller that pinned the store can speak for it.

`kin_core::hydration_semantics::transfer_preserves_creation_record(recorded, declared)` is the whole
rule: true only when both numbers exist and are equal.
`HydrationStampCapability::reconcile_after_transfer(declared)` is what the daemon calls, at the same
pre-commit boundary the unconditional discard used. The capability also gains a `read()` through the
directory handle it was already pinned to, and one parser now serves both it and the ambient path.

In `kin-daemon`, `DaemonState::local_hydration_creation_version()` reads that record through the
pinned handle; the transfer export route and the push command declare it; and the receive wrapper
decides with it. A hosted daemon owns no local record, so it declares `None`, and so does any build
older than this field.

**Release note, because this changes what a first clone looks like.** At v0.7.0 a fresh
`kin clone` from a hosted remote publishes an UNSTAMPED replica: `kin doctor` reads
`Hydration semantics STALE, this store records no hydration semantics version`, and every MCP answer
over that store carries `degraded.hydration_semantics_stale` and a verdict of `inconclusive`. That is
the intended outcome, not a regression. A hosted daemon owns no creation record, so it declares
nothing, and a replica that cannot match a declaration must not certify. What that replaces is a
false all-clear: before this change the clone door never reached the transfer receiver at all, so the
replica was stamped with the cloning build's own version and read `Current` over history it had no
provenance for. Measured on the baseline, a clone from a source with no record at all published
`ok, version 10 at creation`.

`HydrationStanding::remedy()` stops presuming a source outside the store for the unstamped and
unreadable arms. It now leads with the upgrade, says the store keeps serving its history with its
creation-time version unknown, and says what re-ingesting actually costs instead of naming it as a
step that preserves anything.

## The clone door, which the journey run recorded as a pass

The journey lane's bisect logged "clone ok on both stores". Read in source that is the same defect
pointing the other way.

`kin clone` of a native peer never reaches the transfer receiver.
`crates/kin-cli/src/commands/clone.rs` calls `kin_core::init::replica::initialize`, which stages the
layout, and `crates/kin-core/src/init/replica.rs` admits the peer's bootstrap pack INSIDE that
staging. The staging is also what calls `stamp_staged`, so the replica is stamped with the cloning
build's version and the transported history is committed into it with nothing reconciling the two.
The follow-up pull the CLI runs admits nothing, so the receiver's discard never fires either. A
clone of a peer whose history was authored under another replay version therefore published a
replica reading `Current` over it, in the most-used adoption door. The kin-daemon arm that looks
like it covers this drives `clone_native_replica`, the daemon's adoption path, not the CLI's.

The declaration this change adds makes the fix one comparison in the place the stamp is already
written: `PreparedRepositoryInit::reconcile_bootstrap_hydration_semantics` runs inside staging,
before the publishing rename, so nothing is visible yet and a failure aborts the conversion rather
than publishing a store that misreports itself.

Measured on the same baseline bytes, two arms out of one script run.

A source with no record at all, which is what store A becomes once the push above has taken its
record, cloned into C:

```
--- doctor A-restamped-9 ---
  ! Hydration semantics  STALE  this store records no hydration semantics version ...
--- stamp file A-restamped-9 ---
(absent)

--- doctor C-cloned-from-a-version-9-source ---
  ✓ Hydration semantics  ok  this store records hydration semantics version 10 at creation,
                             matching the version this build derives
--- stamp file C-cloned-from-a-version-9-source ---
{"schema":"kin.hydration-semantics.v1","created_under":10,"at":"2026-09-05T02:27:39.217621Z"}
```

And a fresh store D whose record says version 9, so D's own doctor says it cannot certify, cloned
into E:

```
--- doctor D-recorded-9 ---
  ! Hydration semantics  STALE  this store records hydration semantics version 9 at creation and
                                this build derives version 10, so the store cannot certify that its
                                persisted history reflects this build's replay semantics

--- doctor E-cloned-from-a-version-9-source ---
  ✓ Hydration semantics  ok  this store records hydration semantics version 10 at creation,
                             matching the version this build derives
```

`kin clone` launders a source that says it cannot certify its own history into a replica that
certifies. The same binaries produced the push's over-caution and the clone's over-confidence inside
two minutes.

## A pairing nothing before a landing was checking

`scripts/acceptance/hydration_semantics_repro.py` grades this advice by exact equality on purpose,
and its own doc says why a prefix match would be wrong. But that suite lives in `Product
Acceptance`, which carries `if: github.event_name != 'pull_request'`, so neither it nor its
`--self-test` step grades a pull request. And the self-test could not catch this drift anyway: every
fixture there is built from its own constant, so changing the constant changes both sides. Measured
on this tree, tampering the constant and rerunning `--self-test` still reports 83 of 83.

So a sentence changed here and not there lands green and turns main red, where a red acceptance run
stops a release cut. The second commit adds the pull-request-side half: a kin-core test that
`include_str!`s the acceptance script and asserts every gap standing's advice appears in it, with a
must-miss control. It reads the script rather than restating its strings, which is the same shape
that already holds the transfer seam contract together from
`crates/kin-remote/src/repository_transfer_http.rs`.

## What keeping does not claim

The declaration says which version was in force when the SENDING store was created, which is exactly
what the receiver's own record says about itself. Keeping leaves the receiver's claim no stronger
than the sender's already was, and every other pairing still discards. A peer that lies about its
version gains nothing it did not already have, because it authored the deltas it is sending; what
the identity binding adds is that nobody in between can change the declaration the sender made.

## Verification

Everything below ran on this branch at `f87e1ed2a`, in the lane worktree, through
`.kin-coord/bin/heavy-slot.sh` for anything that builds. Commands and their actual tails.

```
=== build rc=0        cargo build --release -p kin-cli -p kin-daemon
=== core rc=0         cargo test -p kin-core --lib hydration_semantics        22 passed
=== replica rc=0      cargo test -p kin-core --lib init::replica               8 passed
=== remote rc=0       cargo test -p kin-remote --lib repository_transfer
=== daemonmod rc=0    cargo test -p kin-daemon --lib hydration_semantics_on_native_transfer
=== policy rc=0       cargo test -p kin-daemon --test transfer_admission_policy
=== fmt rc=0          cargo fmt --all -- --check
```

The eight receiver arms, by name:

```
a_hosted_receive_leaves_its_scaffolding_record_alone ... ok
an_http_receive_declaring_another_version_discards_the_creation_record ... ok
a_refused_receive_leaves_the_creation_record_intact ... ok
an_http_receive_declaring_this_stores_version_keeps_the_creation_record ... ok
an_http_receive_from_an_undeclared_sender_discards_the_creation_record ... ok
a_native_clone_of_a_matching_source_keeps_the_replicas_record ... ok
a_pull_that_admits_nothing_leaves_the_creation_record_alone ... ok
a_native_clone_of_a_mismatched_source_leaves_the_replica_unstamped ... ok
test result: ok. 8 passed; 0 failed
```

The first run of this set exited 101 on `remote`, and it was a compile error rather than an
assertion: one test fixture builds its pack by struct literal rather than through a builder, so the
mechanical call-site pass never reached it. Fixed in the second commit, with the rest of the tree
swept by hand for other literals; there are none.

Two more that need no builder slot:

```
$ bash scripts/zero_file_search_guard.sh
Zero File-Search guard passed: 70 answer modules are graph-clean (8 with declared boundary pins
deferred to verify-zero-file-search.py).
rc=0
```

Same count as the clean tree before the change, so nothing this adds reaches an authority path.

```
$ cd packages/boundary-contracts && node --test
ℹ tests 19
ℹ pass 19
ℹ fail 0
```

That suite is the one that refuses drift between the Rust struct, the JSON contract and the
TypeScript declarations, and it runs inside the required `Fast gate lint and policy` context. Its
positive control, with the new key removed from the contract alone, fails on the right assertion:
`RepositoryTransferPack and the export response envelope must carry the same keys in the same
order`.

The acceptance suite grades `remedy()` by exact equality and runs on main's push, not on this pull
request, so a single space between the Rust literal and the Python constant would land green and
turn main red. Checked by reassembling the Rust literal the way rustc does and comparing:
`IDENTICAL`, with a one-space control returning `False`. The suite's own `--self-test` is 83 of 83.

## Falsification

Five mutations, each applied by exact string swap and reverted the same way rather than with
`git checkout --`, which restores HEAD and would destroy any other uncommitted edit in the file.
Each revert is proved by `git status --porcelain` returning empty before the next arm starts, and
the script refuses to begin over a dirty tree.

Every one turns exactly its intended arm red and leaves every other arm and control green.

```
M1  the receiver stops deciding and discards on every admission
    kin-core   1 failed of 23:  reconciling_after_a_transfer_keeps_only_a_matching_declaration
    kin-daemon 2 failed of 8:   an_http_receive_declaring_this_stores_version_keeps_the_creation_record
                                a_native_clone_of_a_matching_source_keeps_the_replicas_record

M2  the sender stops declaring, so the wire carries nothing again
    kin-daemon 1 failed of 8:   a_native_clone_of_a_matching_source_keeps_the_replicas_record

M3  the declaration stops being bound by the transfer identity
    kin-remote 1 failed of 69:  the_declared_hydration_version_is_bound_by_the_transfer_identity

M4  the clone door stops reconciling inside staging
    kin-core   1 failed of 8:   a_bootstrap_this_build_cannot_match_publishes_no_creation_record

M5  the bootstrap stops forwarding the pack's declaration
    kin-remote 1 failed of 69:  a_replica_bootstrap_forwards_the_packs_declared_hydration_version
```

M2 is the one worth reading twice. It leaves both receive-route arms green, because those build
their own declaration in the fixture, and turns only the end-to-end clone arm red. That is the arm
covering the sender half, and without it a sender that stopped declaring would ship green.

## Notes for the release

The transfer schema version moves, and the two builds refuse each other in two different ways
depending on which one BUILT the pack, not on which command was typed.

A pack built by an older build reaches the version refusal by name, "unsupported schema version 4;
expected 5". That is a v0.6.8 peer pushing into this one, and this one pulling from a v0.6.8 peer.
The new field carries `#[serde(default)]` precisely so a schema 4 pack still decodes and gets that
sentence rather than dying earlier as a missing field, and there is a test arm for exactly that.

A pack built by THIS one and handed to an older peer does not reach it. `RepositoryTransferPack`
carries `#[serde(deny_unknown_fields)]`, so the older peer fails on an unknown
`source_hydration_semantics` before any version check runs. That is this one pushing into a v0.6.8
peer, and a v0.6.8 peer pulling from this one.

Neither refusal tells the user to upgrade, and the serde one does not even name the schema. A
follow-up ticket should give the version mismatch a sentence that names both builds and the action,
on both sides of the wire.

`scripts/acceptance/hydration_semantics_repro.py` grades the remedy by exact equality on purpose, so
its `REMEDY_UNKNOWN` constant moves with the sentence. Its `native_transfer` fixture stamps the
source one version back, so it still expects a discard and its grader is unchanged; that check is
not in the `CHECKS` tuple and does not run. The suite's `--self-test` is 83 of 83 with the new
constant.


troyjr4103:
Independent review at `f618c139c533f36146f0fc9024718357e1e345a5`, read-only, worktree pinned by
`git reset --hard` at that sha. **GO**, once the outstanding checks conclude green and one sentence
is added to the release note.

The change is right and the tests hold up. I took M2 myself rather than reading its tail: the sender
stops declaring, applied by exact string swap to the export route, and
`cargo test -p kin-daemon --lib hydration_semantics_on_native_transfer` gives 7 passed, 1 failed,
and the one is `a_native_clone_of_a_matching_source_keeps_the_replicas_record`, panicking with
`left: Unstamped { derives: 10 }` against `right: Current { version: 10 }`. Exactly the claim.
Reverted by exact string swap, porcelain empty. I also read the identity arm's assertions rather
than its name: forging `None`, `Some(9)` and `Some(11)` over a `Some(10)` pack each fail on
"transfer identity", an undeclared pack is a valid pack, and a stripped key stamped schema 4 is
refused by version. A proxy can neither forge nor strip.

**Four findings, most severe first.**

**1. HIGH, disclosure not code. A fresh clone from the hosted daemon publishes an unstamped replica
at v0.7.0, and the body says the opposite for that door.** A hosted daemon sets
`local_kindb_capability: None` (`crates/kin-daemon/src/state.rs:5557`), so it declares `None` on
every export; the CLI clone forwards that into the bootstrap
(`crates/kin-cli/src/commands/clone.rs:154`) and `reconcile_bootstrap_hydration_semantics(None)`
invalidates. `a_bootstrap_this_build_cannot_match_publishes_no_creation_record` already asserts it
for the `None` arm, and the M2 tail above is the same result end to end. Downstream,
`is_gap()` sets `degraded.hydration_semantics_stale` (`crates/kin-mcp/src/envelope.rs:2031`), which
that module calls "the one input the verdict is computed from" at `:1639`. So a fresh
`kin clone` from kinlab.ai reads STALE and inconclusive where before this change it read certified.

That outcome is honest and should ship. A hosted store owns no creation record and its history came
from arbitrary clients, so `None` is the only true declaration and the old `Current` was a false
all-clear of exactly the kind this mechanism exists to prevent. What needs fixing is the sentence:
"A hosted daemon owns no local record, so it declares `None` and every receiver pulling from it
discards exactly as before" is not true for the clone door, because before this change a clone of an
undeclared source kept its record. Please add one release-note line saying hosted clones read
unstamped at v0.7.0, and open a follow-up for a hosted daemon that records and declares the version
its admitted packs carried. It receives one on every push now, so the input already exists.

**2. MEDIUM. The compatibility claim holds only in the push direction.** `RepositoryTransferPack`
carries `#[serde(deny_unknown_fields)]` (`crates/kin-remote/src/repository_transfer.rs:567`; the arm
at `:4631` is the live control). Push works as described: a schema 4 pack decodes through
`#[serde(default)]` and is refused with "unsupported schema version 4; expected 5". A v0.6.8 client
*pulling* from a v0.7.0 daemon fails as an unknown-field serde error and never reaches that
sentence, so "refuse each other by name" wants narrowing. Pre-existing class, since any added field
does this, but the release note leans on the claim. Separately, neither refusal tells the user to
upgrade.

**3. LOW. The kin-core `include_str!` leaves the crate root.** Safe today: `cargo publish` appears
nowhere in `.github/` or `scripts/` except `check-rc-build-drift.test.mjs:186`, which forbids it,
and the promotion bundle is a whole-tree `git archive`. It becomes a build break the day kin-core is
published, because `cargo package` restricts to the package directory. Worth a line in the test's
doc comment.

**4. LOW.** The umbrella-store follow-up in the lane report names no Linear ticket. Please file one
so it outlives the report.

**On the previous head.** At `2c4e2cdc1` the required `Fast gate lint and policy` was red, and not
on infrastructure: job step "Check Zero File-Search Invariant" exited 1 with both a VIOLATION at
`hydration_semantics.rs:326` and `allow_fn 'read' resolves to 2 function bodies (want 1)`.
Reproduced locally. Worth naming why it got through: the body cites
`bash scripts/zero_file_search_guard.sh` rc=0, and ci.yml runs `python3
scripts/verify-zero-file-search.py` first; that is the step that failed, and it skipped the shell
guard, clippy and the boundary-contract tests behind it. `f618c139c` fixes it, and I falsified the
fix rather than trusting the reason text: adding a third body named `read` with a
`std::fs::read_to_string` turns the verifier red on both axes, `resolves to 3 function bodies (want
2)` plus a VIOLATION. Removed, verifier back to rc=0.

**Hosted set at this head,** `per_page=100`, `listed=46 total=46`, read 2026-09-05T03:02Z. Zero
failures. Five unconcluded: `Analyze (rust)`, the three `Fast gate test shard` runs and `MCP surface
contract`. `CodeQL` reads `neutral`, which says nothing until `Analyze (rust)` concludes. Six
required contexts green by name including `Fast gate lint and policy`. Not landable yet on the
read-by-name rule, and nothing red.

Also checked and clean: `stamp_staged` at `init.rs:855` runs before the bootstrap reconcile inside
the same staging, so a failure aborts the conversion rather than publishing a store that misreports
itself; `hydration_semantics::write` has one product caller, so no follow-up pull can re-stamp after
a discard; `packages/boundary-contracts` is 19 of 19 here; kinlab imports the pack as a type and
casts a response to it, constructs no literal and pins no schema 4, so the extra key breaks nothing
there. No em dash in the title, the body or any added line, checked with a control that fires. No
session id or trailer. All four commits signed off.

PR #1513, opened 2026-09-05
Say which files the graph answers about from an earlier parse

troyjr4103:
Journey GAP-9. A file with a syntax error commits with rc 0 and "0 entities, 0 relations, 1
artifacts", and `kin status`, `kin diff`, `kin graph status` and `kin doctor` all report a whole
store, while `.kin/daemon.log` carries `broken AST, retaining LKG state file=search.py errors=4`.
The daemon knew, five seconds before every surface said otherwise. Found by the journey lane's
cold-user run (step 10a), and it was the stranger's own fix-first on v0.6.7.

## Measured first, on this main

Scratch `HOME` and `KIN_HOME`, `KIN_EMBED_BACKEND=cpu`, greenfield store, release binaries built
from this branch's base `a61ab56b7684087d96f1ed968c58e79af38cfdd4`. One edit: the closing bracket of
`search.py`'s list comprehension removed, which `ast.parse` rejects. Five seconds later:

```
$ kin status
Tree: 35f7659c... (3 artifacts, ahead of its base change as admitted 0s ago)
Durable semantic enrichment: present (11 entities, 4 relations, 1 changes ...)
Untracked host content: none, measured 0s ago
rc=0

$ kin diff
Artifacts: +0 ~1 -0
Entities: +0 ~0 -0
   -    return [word.lower() for word in text.split()]
   +    return [word.lower() for word in text.split(
rc=0

$ kin commit -m "Broken edit"
Created semantic change b41b7f45... on branch 'refs/heads/main' (0 entities, 0 relations, 1 artifacts)
rc=0

$ kin graph status
Repository coverage: 3 of 3 admitted files produced entities (100%)
Parse coverage (files that produced an entity / files admitted):
  python: 2/2 (100%)

$ kin doctor
  ✓ Parse coverage             ok             python 2/2; typescript 1/1

$ grep "broken AST" .kin/daemon.log
WARN kin_reconcile::reconciler: broken AST, retaining LKG state file=.../search.py errors=4
```

Three things the measurement changed about the shape of the fix, and none of them were visible from
source alone. The nearest thing to a signal today is `kin graph status`'s note `1 of 3 admitted
regular files have no query-facing enrichment facet yet`, phrased as a race that is about to settle,
which is worse than silence because it reads as a reassurance. `kin graph status` lies twice, in two
separate collectors, so fixing only the census would leave one section at 1/2 and another at 100%.
And the error count is per-edit: the journey's identical edit produced 3 where this one produces 4,
so no surface may assert a number of its own.

## Why the commit still succeeds

Retaining the last good parse is the designed answer to broken source, not a fault. The daemon
reconciles under `ReconcilePolicy::FallbackToLkg`, an agent mid-edit produces half-written files
constantly, and refusing every commit until the syntax is fixed would leave a caller unable to
record anything at all, which is worse than the stale spans. Zero entities is the correct count of
what the broken parse derived. What was missing is that nobody was told: the reconciler returns
`ReconcileOutcome::BrokenAst`, every consumer computes `should_apply` and drops it, and the fact
reaches nothing but the log.

## One record, five readers

`crates/kin-core/src/retained_parse.rs`, published beside the last-admission marker on that module's
exact pattern: a schema token, a three-way read (`Recorded`, `Absent`, `Unreadable`, because a
record that will not parse is a louder fact than one that is missing and neither may render as a
clean store), and an atomic staged-and-renamed write with the directory metadata synced. Its `fold`
is a pure function holding the whole rule, and `describe` is the one sentence every surface renders,
so five surfaces cannot word one fact five ways.

Durable rather than in-process because the pass that learns the fact and the command that reports it
are usually different processes, and because a pass with nothing to admit returns before it
reconciles anything: recomputing per surface would report a store whole the moment nothing changed.

Two writers, both daemon reconcile seams, both publishing once per pass: the ambient tick's event
loop, and the complete sync that `/commands/admit` and `/commands/commit` both drive. The rule is
"of the paths this pass observed, these were retained; drop every other observed path; leave
unobserved paths alone", which is one rule for two different populations: a partial tick must not
clear a path it never looked at, and a fixed file must not stay named forever. Both removal arms
settle explicitly, because a removed path never reaches the reconcile below. Both publish while
still holding the reconciler write lock, because `record` is a read-modify-write of one document and
two interleaved passes would let the later write drop what the earlier one learned. `Conflict` also
retains last-known-good state and is deliberately not recorded: a held merge is not a broken syntax,
it has its own surface, and folding the two would tell a reader to go fix a syntax error in a file
whose syntax is fine.

Then the five readers. `kin commit` takes the line the way it already takes `pending_enrichment`'s,
read CLI-side after the commit, so `CommandCommitResponse` does not widen; that is the open question
in #1501's body answered without the field. `kin status` takes the read beside the
`&LastAdmissionRead` `render_text` already carried, and the line goes directly under the enrichment
counts it qualifies rather than at the foot of the page. `kin diff` prints it on both routes from
the store, so the daemon-rendered lines stay byte-identical and the two routes cannot drift.
`kin graph status`'s census gains a third bucket, tested BEFORE the producing set because a retained
file is in `producing` by construction, with the invariant `with_entities + silent + retained ==
tracked`; its heading becomes "files whose current bytes produced an entity / files admitted" so the
numerator and its label say the same thing, and the `Repository coverage` block keeps its literal
reading and gains the disclosure directly beneath it. `kin doctor` reaches a verdict on that one
arm, `Pending` rather than `Missing` since the graph is still answering, with the path named and the
step that closes it.

That last one is worth naming. `parse_coverage_health`'s own doc comment says it never judges
because "no such signal exists in graph-owned state today: a side-effect script, a re-export and a
comment-only file each produce no entity and are each perfectly correct". A file retained from an
earlier parse is exactly the signal it was waiting for, and the only member of that population that
is unambiguously a gap. A silent file still does not judge, and the test holds both halves.

`kin graph validate` and the support dump take the same parameter, so no surface renders the new
heading over the old arithmetic.

## The same repro against these bytes

Fresh store, same script, binaries built from this branch:

```
$ kin status
Durable semantic enrichment: present (11 entities, 4 relations, 1 changes ...)
Retained from last good parse: search.py (4 parse errors). Their current bytes did not parse, so the
graph answers about them at the positions an earlier parse recorded, and every span, reference and
enumeration over them describes those bytes rather than the ones on disk. Fix the syntax and the
next admission re-derives them. Observed 4s ago.

$ kin diff
Entities: +0 ~0 -0
Retained from last good parse: search.py (4 parse errors). ...

$ kin commit -m "Broken edit"
Created semantic change 11ae6e7f... on branch 'refs/heads/main' (0 entities, 0 relations, 1 artifacts)
Recorded in Kin authority, not in git. ...
Retained from last good parse: search.py (4 parse errors). ...

$ kin graph status
Repository coverage: 3 of 3 admitted files produced entities (100%)
  Retained from last good parse: search.py (4 parse errors). ...
Parse coverage (files whose current bytes produced an entity / files admitted):
  python: 1/2 (50%)
  typescript: 1/1 (100%)
  retained_last_good_parse: 1 of 2 admitted python files did not parse from their current bytes, so
  the entities the graph holds for them came from an earlier parse and every span, reference and
  enumeration over them describes bytes that are gone, including search.py (4 parse errors)
  (shallowest paths first). Fix the syntax and the next admission re-derives them.

$ kin doctor
  … Parse coverage             PENDING        python 1/2; typescript 1/1; retained_last_good_parse: ...
      fix:  fix the syntax in search.py (4 parse errors) and run `kin admit`, which re-derives them;
            until then every span, reference and enumeration Kin answers over those paths describes
            an earlier version of them
```

The daemon log carries the same one line it always did, at 03:16:37.665442Z, and the record it now
leaves beside it was written 81 microseconds later:
`{"schema":"kin.retained-parse.v1","at":"2026-09-05T03:16:37.665523Z","paths":[{"path":"search.py","errors":4}]}`.
Before the edit, in the same run on the same store, every surface is silent and `python: 2/2 (100%)`
stands, which is correct.

## Falsification

Nine arms, each mutation applied alone with the tree restored from a byte copy before the next, and
the same nine tests pass together on the unmutated tree. Verbatim tails below.

**M1 survived the first run, and that is worth reading.** The arm's anchor was a
sixteen-space-indented line that is also a SUBSTRING of the twenty-space-indented one in the ambient
tick, so `count == 1` was false, the assertion raised inside the mutation helper, the helper
swallowed it, and the arm compiled and graded an unmutated tree and reported green. A falsification
arm that cannot apply its own mutation is a check that cannot fail. The fix is the guard rather than
the anchor: the helper now refuses to run the test when the edit did not apply, and the rerun prints
the remaining call count as a positive control before it compiles.

```
M1: drop the complete sync's observation.
M1 applied
M1 control: 1 extend call(s) remain (the ambient tick keeps its own; the sync's is gone)
test loop_runner::tests::a_complete_sync_records_a_file_it_could_not_parse_and_clears_it_when_fixed ... FAILED
panicked at crates/kin-daemon/src/loop_runner.rs:6932:13:
the pass that could not parse search.py recorded nothing: Absent
```

```
M2: a clean parse no longer settles the path.
assertion `left == right` failed: a path that parses again must leave the record, or every surface names it forever
  left: Some(1)
 right: None
```

```
M3: the census tests `producing` before `retained`, which is the original order.
test reference_coverage::tests::a_file_retained_from_an_earlier_parse_leaves_the_numerator ... FAILED
panicked at crates/kin-core/src/reference_coverage.rs:2570:9
```

```
M4: the doctor row stops judging.
test commands::health::tests::a_file_retained_from_an_earlier_parse_needs_attention_and_names_its_next_step ... FAILED
```

```
M5: the status line is appended instead of placed under the counts it qualifies.
test commands::status::tests::the_retained_line_sits_under_the_enrichment_counts_it_qualifies ... FAILED
```

```
M6: an unreadable record renders as a clean store.
test retained_parse::tests::an_unreadable_record_says_so_rather_than_reading_as_a_clean_store ... FAILED
```

```
M7: the commit summary drops the line.
test commands::commit::tests::a_commit_over_a_file_that_did_not_parse_names_it_beneath_the_counts ... FAILED
panicked at crates/kin-cli/src/commands/commit.rs:1201:9
```

```
M8: the fold keeps a path the pass settled.
test retained_parse::tests::a_pass_settles_what_it_observed_and_leaves_what_it_did_not ... FAILED
panicked at crates/kin-core/src/retained_parse.rs:468:9
```

```
M9: the diff line goes silent.
test commands::diff::tests::the_diff_line_names_a_file_the_graph_answers_about_from_an_earlier_parse ... FAILED
```

Restored tree, the nine together: `9 passed; 0 failed` (kin-core), `20 passed; 0 failed` (kin-cli),
`1 passed; 0 failed` (kin-daemon).

## What ran locally, and what did not

Speed rule in force on this lane: no local precheck, no parity, no local clippy. Everything below
ran through the fleet builder slot on this head's tree. Hosted CI grades clippy and the full
workspace.

```
cargo fmt --all                                      FMT_RC=0
cargo check -p kin-core -p kin-cli -p kin-daemon --all-targets
                                                     CHECK_RC=0
cargo test -p kin-core --lib retained_parse          7 passed; 0 failed
cargo test -p kin-core --lib reference_coverage      29 passed; 0 failed
cargo test -p kin-cli --lib -- parse_coverage retained_line retained_record retained_parse
    diff_line_names commit_over_a_file needs_attention_and_names
                                                     5 passed; 0 failed
cargo test -p kin-daemon --lib -- a_complete_sync_records_a_file
                                                     1 passed; 0 failed
python3 scripts/verify-zero-file-search.py           Verification PASSED: Zero File-Search
                                                     Invariant holds. Scanned 316 source files.
bash scripts/zero_file_search_guard.sh               passed: 70 answer modules are graph-clean
```

The new module carries a function-scoped allowlist entry naming `read`, `write` and the two
cfg-gated `sync_directory_metadata` bodies, on the same justification the last-admission and
relation-census records beside it carry: durable bookkeeping at an explicit boundary, naming no
entity, holding no repository content, and unable to change what any query returns.

`scripts/acceptance/parse_hole_repro.py` moves with the heading it grades, in five places. Its own
two fixtures create silent files rather than retained ones, so its assertion that the doctor row
stays healthy for a silent file still holds, and this change is what makes that assertion mean
something rather than describe every case.


troyjr4103:
Independent review at `123696f6f7f15a9157909529719ed985e06e4d88`, read-only, on the rebased head.

**Verdict: GO.** The design holds, the lock question is a clean no, the zero-file-search entry is
correct, and the fold rule's tests go red under a mutation. One MEDIUM on the wording, three LOWs.

**The rebase.** Same sixteen files, same insertion counts everywhere except `status.rs`, which grew
by five lines. `retained_parse.rs`, `reference_coverage.rs`, `layout.rs`, `graph.rs` and `commit.rs`
are byte-identical to `5317399b7`; `loop_runner.rs`, `health.rs`, `api.rs` and the allowlist differ
only where main moved under them. The one resolution is `status.rs` against #1509's
`render_text_with_tip` split, and it came out better than what it replaced: the retained line is
now inserted by searching for `Live graph enrichment:` rather than by index, so the workspace-tip
insert, the merge banner and this line cannot shift each other whatever order they run in. The
comment naming `StatusReportWire`'s `deny_unknown_fields` is the right reason for reading CLI-side.

**Lock order: no new order, so no cycle.** `record` takes no lock at all. It is one
`read_to_string` and one staged-and-renamed write. No flock, no commit-point freeze, no publication
lock, so it cannot deadlock against a held authority. Both seams publish under the reconciler write
lock, which is what its read-modify-write needs.

**1. MEDIUM. The sentence claims an earlier parse for a file that never had one.**
`ReconcileOutcome::BrokenAst` does not mean "the graph is answering from an earlier parse". It means
"these bytes did not parse". `reconcile_file_edit` (`crates/kin-reconcile/src/reconciler.rs:653`) is
the create-or-modify path, and its `FallbackToLkg` arm at `:678` returns `BrokenAst` immediately
after indexing, before any graph lookup, with no check that the file has prior state. `FileEvent`
carries only `Changed` and `Removed` (`:435`), so a brand-new file arrives as `Changed` and routes
there at `:448`. A newly created file with a syntax error, for which the graph holds nothing,
therefore lands in the retained bucket and gets:

> the graph answers about them at the positions an earlier parse recorded, and every span,
> reference and enumeration over them describes those bytes rather than the ones on disk

There is no earlier parse and there are no positions. The census sentence makes the same claim, and
`graph status`'s new disclaimer asserts the case "IS established" on the same premise. The remedy is
unaffected and still right, and "its bytes are on disk, they do not parse" is true in both
populations, so this is a false diagnosis clause rather than a wrong instruction. It matters because
it is the one sentence five surfaces render, and a new file with a typo is at least as common as an
edit that breaks an existing one. Narrowest fix is to lead with the bytes not parsing and qualify
the stale-span clause; carrying a had-prior-entities flag would let `describe` say the stronger
thing only where it is true.

**2. LOW. The two seams disagree about the graph mutation, against the code's own stated rule.**
The ambient tick drops the graph mutation first and publishes under the reconciler lock only
(`loop_runner.rs:3579` then `:3588`), commented "after the graph mutation is released so a marker
write never sits under it". The complete sync publishes at `:8986` with both still held. So on the
`/commands/admit` and `/commands/commit` path, two fsyncs run under the lock the tick deliberately
released first. Serialization is still correct and `record` returns early on an empty observation
set, so this is latency and an inconsistency rather than a defect.

**3. LOW. The doctor's retained arm returns before the silent arm.** `parse_coverage_health` returns
on `any_retained()` with `format!("{summary}; {retained}")`, so a store holding both a retained file
and a silent file names only the retained paths and drops the silent sample that row used to print.
The two new tests cover retained-only (`retained_census_row("python", 2, 1)`) and silent-only
(`parse_census_row("javascript", 141, 75)`), so nothing covers both together. `kin graph status`
still prints both and the retained case is the actionable one, so this is one row losing a signal it
used to carry.

**4. LOW. Two narrow ways a path outlives its file.** The `settled` push sits inside the `Ok(())`
arm of the exact-tree removal (`:3332`, `:8868`), so a removal that errors warns and leaves the path
named with nothing left to observe it; and a broken file that becomes ignored produces no further
event. Both are far off the ordinary route. Deletion and rename on the normal path are handled: a
rename is a `Removed` plus a `Changed`, so the old path settles on the same pass.

**Falsification, taken independently.** One arm, on the pure fold rule, through the fleet builder
slot. `if let Some(errors) = observation.errors {` swapped for
`... observation.errors.filter(|_| false) {`, so `fold` settles a `BrokenAst` observation instead of
retaining it. The script counted the mutated site and refused to run unless exactly one applied.

```
BASELINE: test result: ok. 7 passed; 0 failed  (kin-core --lib retained_parse)
MUTATION applied at 1 site(s)
panicked at crates/kin-core/src/retained_parse.rs:393:9: assertion `left == right` failed
  left: None
 right: Some(4)
panicked at crates/kin-core/src/retained_parse.rs:484:9: assertion `left == right` failed
  left: None
 right: Some(4)
test result: FAILED. 3 passed; 4 failed
```

Positive control that the run is not uniformly red: the census suite under the same mutation is
`28 passed; 1 failed`, and the one failure is
`a_file_retained_from_an_earlier_parse_leaves_the_numerator`, which builds its fixture through
`fold`. Tree restored by byte copy, porcelain empty. `retained_parse.rs` is byte-identical at
`5317399b7` and `123696f6f`, so the arm holds at this head unchanged.

**Checks I ran on the design.** The census tests retained before producing
(`reference_coverage.rs:1402`), which is required rather than preferred, since a retained file is in
`producing` by construction; `with_entities + silent + retained == tracked` holds by construction,
one bucket per tracked file. The allowlist entry is function-scoped, names `read`, `write` and
`sync_directory_metadata` with `count: 2`, and follows the `hydration_semantics.rs` entry beside it,
so the fold, the three-way read and the describe sentence stay scanned. `CommandCommitResponse` is
untouched. `graph validate` and the support dump take the same parameter, and `None` yields `Absent`,
which leaves those sections exactly as they read before. `collect_parse_coverage` now has no
production caller.

**Scope note, not a defect here.** The MCP `kin_graph_status` path (`api.rs:2842`) builds no
parse-coverage census at all, so the agent-facing surface never carried this claim and still cannot
disclose a retained path. Worth a follow-up given who the substrate is for.

**Hosted set** read at the exact head with `per_page=100`, listed equals total.

troyjr4103:
Re-review of the fix commit, at `505e201d798f2eeec2f0692b3d0ea460bfa5cfe9`, read-only.

**Verdict: GO.** The MEDIUM is fixed correctly and one LOW with it. Two LOWs stay open and are
ticketed.

**The delta alone.** Six files: `commit.rs` 4, `diff.rs` 2, `health.rs` 80, `status.rs` 4,
`reference_coverage.rs` 30, `retained_parse.rs` 55. `scripts/zero-file-search-allowlist.json`,
`loop_runner.rs` and `layout.rs` are untouched, so the zero-file-search entry and both reconcile
seams are exactly what was reviewed at `123696f6f`. Inside `retained_parse.rs` the only non-prose
change is `describe()`. `fold`, `record`, `read`, `write` and both `sync_directory_metadata` bodies
are byte-identical, grepped by name with the only hit a doc-comment line.

**The wording, graded against both populations rather than accepted.** The sentence is now:

> Did not parse as written: `<paths>`. The bytes on disk do not parse, so any entities the graph
> still holds for these paths came from an earlier parse of bytes that are gone, and a path the
> graph never parsed is absent from it entirely.

For an edited file with a prior parse: clause one true, clause two true, clause three vacuous. For a
new file with a typo, the population the old sentence lied about: clause one true, clause two
vacuously true because there are none, clause three true and now the informative half. Correct on
both, and the heading changed too, so the old claim is gone rather than softened. The same fix is
carried into `retained_sentence` and the graph-status disclaimer, which now says only that the bytes
are on disk and do not parse and that whatever the graph answers about the file was not derived from
those bytes. The doctor's manual fix states the conditional plainly: Kin answers about those paths
from an earlier parse where it has one, and not at all where it does not.

Checked rather than assumed: grepping this head for the old claim returns zero hits outside a test
that asserts its absence, with "Did not parse as written" as the positive control hitting
`commit.rs`, `status.rs` and `retained_parse.rs`. The falsification carries three assertions
including a negative one, so the old sentence cannot come back quietly.

**The doctor early return** is fixed the way it should be. `detail` is built from the summary plus
the retained lines plus the silent lines, the verdict and the manual fix stay gated on
`any_retained()` alone, and the new test asserts `search.py`, `lib/express.js`, `python 1/2` and
`javascript 66/141` all survive in one row. A store with one broken file among seventy-five silent
ones no longer loses the larger half of its own row.

**My falsification arm, re-run at this head** through the fleet builder slot, same mutation as
before: `fold` settling a `BrokenAst` observation instead of retaining it, guarded on exactly one
applied site.

```
BASELINE: test result: ok. 7 passed; 0 failed
MUTATION applied at 1 site(s)
a_pass_settles_what_it_observed_and_leaves_what_it_did_not
  left: ["untouched.py"]
 right: ["newly_broken.ts", "still_broken.py", "untouched.py"]
a_recorded_path_round_trips_and_names_itself_with_its_error_count   left: None  right: Some(4)
a_pass_that_observed_nothing_leaves_the_record_standing             left: None  right: Some(4)
test result: FAILED. 3 passed; 4 failed
```

Control that the run is not uniformly red: the census suite under the same mutation is
`28 passed; 1 failed`. Tree restored by byte copy, porcelain empty.

**Still open, both LOW and both ticketed.** The complete sync publishes at `loop_runner.rs:8986`
with the graph mutation held while the ambient tick drops it first at `:3579` under a comment saying
a marker write must never sit under it; correct, but two fsyncs under a lock one seam thinks it
should not be under. And the `settled` push sits inside the `Ok(())` arm of the exact-tree removal,
so a removal that errors leaves the path named.

**Two follow-ups outside this PR.** `api.rs:6479` and `loop_runner.rs:8395` still carry the same
earlier-parse over-claim this commit just removed; both are pre-existing, confirmed by diffing
against this PR's base, and came in with #1501's semantic-debt drain. And MCP `kin_graph_status`
builds no parse-coverage census, so the agent-facing surface still cannot disclose a path that did
not parse.

PR #1515, opened 2026-09-05
Refuse to certify a file enumeration over spans of other bytes

troyjr4103:
FIR-3201. A stranger on a converted psf/requests committed a 17-line edit and `list_file_entities` served the module entity at its pre-edit span as `parsed: full`, `freshness: recorded`, no degradation; nine minutes and one daemon restart later the same call said `parsed: absent` while `kin graph status` said python 37/37.

`parsed` read only the layout facet, which a once-per-daemon backfill builds from the entities the graph already holds; `freshness: recorded` is a clock the tree admission stamps; and no verdict input could say a span was older than the blob now at its path.

`list_file_entities` now compares each entity's recorded source digest against the tree's blob at the path, the same rule `get_entity_source` applies before slicing a body, and publishes `span_provenance` and `stale_spans` beside `parsed`. A provable mismatch refuses certification and reaches the verdict as an absent `file_parsed` class with the limit `file_spans_stale`; an unstamped store stays certifiable and says `unverified`. A tree transition that replaces a path's blob retires that path's layout before the tree moves, so the window between admission and re-derivation reads `parsed: absent` instead of certifying the old spans. The layout backfill compares its fresh parse of the tree bytes with the graph's entities by kind, name and span, publishes `Partial` with the count when they disagree, and hands the path back to the loop for re-derivation.

## Provenance

This is lane memfix's commit `33a9500ae` (patch id `22ba8a81e12cf0ad43c6cccc6def9da36cd2be4e`, identical here) cherry-picked onto `950fb0bf8`. The FIR-3203 half of that branch measured over its memory acceptance on the release build and is held back as a draft for the record, so the FIR-3201 fix lands alone.

## Verification on this tree (51f048988), through the fleet's builder slot

```
cargo test -p kin-mcp --lib file_entities
test result: ok. 23 passed; 0 failed; 0 ignored; 0 measured; 896 filtered out; finished in 0.02s

cargo test -p kin-daemon --lib -- loop_runner::tests::a_replaced_blob_retires_its_layout_until_the_spans_are_re_derived loop_runner::tests::the_backfill_refuses_full_over_entities_a_fresh_parse_does_not_reproduce
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 1415 filtered out; finished in 1.56s

cargo fmt --all -- --check
rc=0
```

Not run locally, by the cut's speed rule: clippy, precheck, parity, the full workspace suite. Hosted CI grades them here.

## Falsification on this tree, each arm restored from a byte copy

Each arm mutates the guarded behaviour, prints the applied diff stat, runs the test named for it, and restores the file. Red is the required result.

```
ARM handler-certification: mutated crates/kin-mcp/src/handlers/file_entities.rs
mutation applied:  1 file changed, 1 deletion(-)
test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 917 filtered out; finished in 0.01s
ARM handler-certification cargo exit=101

ARM envelope-class: mutated crates/kin-mcp/src/envelope.rs
mutation applied:  1 file changed, 1 insertion(+), 1 deletion(-)
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 918 filtered out; finished in 0.01s
ARM envelope-class cargo exit=101

ARM daemon-backfill: mutated crates/kin-daemon/src/loop_runner.rs
mutation applied:  1 file changed, 1 insertion(+), 1 deletion(-)
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1416 filtered out; finished in 1.29s
ARM daemon-backfill cargo exit=101
```

### DIFF
--- a/crates/kin-daemon/src/loop_runner.rs
+++ b/crates/kin-daemon/src/loop_runner.rs
@@ -10,6 +10,7 @@ use std::time::{Duration, Instant, SystemTime};
 use kin_index::{
     FileClassification, FileClassifier, FileEvent, FileWatcher, IndexPipeline, IndexedAny,
 };
+use kin_model::layout::ParseCompleteness;
 use kin_model::{
     EntityFilter, EntityId, EntityStore, FilePathId, Hash256, RepoPath, ShallowTrackedFile,
     TransactionDelta, TreeDelta, TreeEntry,
@@ -2307,12 +2308,63 @@ pub(crate) struct LayoutBackfill {
     pub(crate) unreadable: usize,
     /// Artifacts another facet already owns, which this pass must not touch.
     pub(crate) other_facet: usize,
+    /// Artifacts whose graph entities disagree with a fresh parse of the bytes
+    /// the tree holds: published as `Partial` rather than `Full`, and owed a
+    /// re-derivation. Counted inside `published`.
+    pub(crate) stale: usize,
+    /// The host paths whose entities are owed a re-derivation, for the loop to
+    /// enqueue as ordinary `Changed` events.
+    pub(crate) rederive: Vec<PathBuf>,
 }
 
 impl LayoutBackfill {
     fn observed(&self) -> usize {
         self.already_published + self.published + self.unreadable + self.other_facet
     }
+}
+
+/// One entity's identity as a parse would reproduce it: what it is, what it is
+/// called, and exactly which bytes it spans. Two parses of the same bytes agree
+/// on every field; a parse of different bytes moves the span of everything after
+/// the edit.
+type ParsedSpanKey = (String, String, usize, usize);
+
+fn parsed_span_keys(entities: &[kin_model::Entity]) -> Vec<ParsedSpanKey> {
+    let mut keys = entities
+        .iter()
+        .filter(|entity| entity.role == kin_model::EntityRole::Source)
+        .filter_map(|entity| {
+            entity.span.as_ref().map(|span| {
+                (
+                    format!("{:?}", entity.kind),
+                    entity.name.clone(),
+                    span.start_byte,
+                    span.end_byte,
+                )
+            })
+        })
+        .collect::<Vec<_>>();
+    keys.sort();
+    keys
+}
+
+/// How many of the graph's source entities for one file a fresh parse of the
+/// tree's bytes does not reproduce.
+///
+/// Compared by kind, name and byte span rather than by id, because the
+/// reconciler keeps an entity's id stable across an edit that moves it
+/// (`stable_entity_ids` in kin-reconcile), so ids agree on a store that is
+/// fresh and disagree on one that is not, which is backwards for this question.
+/// Spans are the thing that goes stale, so spans are what is compared.
+pub(crate) fn spans_a_fresh_parse_does_not_reproduce(
+    held: &[kin_model::Entity],
+    fresh: &[kin_model::Entity],
+) -> usize {
+    let fresh_keys = parsed_span_keys(fresh);
+    parsed_span_keys(held)
+        .into_iter()
+        .filter(|key| fresh_keys.binary_search(key).is_err())
+        .count()
 }
 
 /// Publish the per-file parse observation for every admitted entity-source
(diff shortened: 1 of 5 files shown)


troyjr4103:
Independent read-only review of #1515 at `51f048988994b889aed26788949366fd321106cd`, from source
plus six tests and one falsification arm run locally. GO.

**The split is honest.** `git patch-id --stable` on `51f048988` and on `33a9500ae` both give
`22ba8a81e12cf0ad43c6cccc6def9da36cd2be4e`, so the cherry-pick is byte-identical to the commit
reviewed on the combined branch. Base is `950fb0bf8` with exactly one commit on top. Five files, all
FIR-3201: `crates/kin-daemon/src/loop_runner.rs` and the four under `crates/kin-mcp/src`. None of
the held-authority work came with it: a grep of the whole diff for `held_write_authority`,
`HeldWriteAuthority`, `forget_writer`, `install_write_authority`,
`publication_identity_after_commit`, `HeldWriterScope` and `InstallWriter` returns zero, against 12
hits for the FIR-3201 terms as the control. Signed off, and no session ids or assistant traces in
the message or the body.

**Both halves are correct, and I verified them from source rather than from the description.**

The backfill compares by kind, name and byte span rather than by id, which is the right way round:
`stable_entity_ids` (`kin-reconcile/src/reconciler.rs:810,898`) keeps an entity's id stable across
an edit, so ids agree exactly when the store is stale. The false-`Partial` risk I went looking for
is not there: `kin-reconcile::Reconciler` holds an `IndexPipeline` of its own
(`reconciler.rs:10,304,332`), the same one the backfill parses with, and takes kind, name and span
straight from that pipeline's output, so the comparison key is preserved. The control file stays
`Full` and still certifies, and it is a real control: a fix that floored everything would fail on it.

The window test rests on kin-db and pins it in kin's own pipeline.
`InMemoryGraph::apply_transaction_delta` retires path-keyed facets before publishing the new tree
(`kin-db-0.7.102/src/engine/graph.rs:8845-8870`, `ent.file_layouts.remove(&file_id)` for every tree
delta that invalidates the path), so `a_replaced_blob_retires_its_layout_until_the_spans_are_re_derived`
would go red if kin-db stopped retiring: it asserts `full` before the edit and `absent` in the
window, and the window read would become `full`.

On the kin-mcp side, `span_provenance` runs over the whole file's entity set rather than a page
(`file_entities.rs` takes an unpaged `query_entities`), refuses only on a provable digest mismatch,
leaves `unverified` certifiable so converted stores are not floored, and reaches the verdict as an
absent `file_parsed` class with `file_spans_stale` named in both `limits` and `floor_reason`.

**Six green, one arm red.** All 23 `file_entities` tests in kin-mcp pass, and both
`a_replaced_blob_retires_its_layout_until_the_spans_are_re_derived` and
`the_backfill_refuses_full_over_entities_a_fresh_parse_does_not_reproduce` pass in kin-daemon. Arm
taken independently, handler-certification: dropping `&& provenance.permits_certification()` from
`certifies_enumeration` at `file_entities.rs:708`.

```
test handlers::file_entities::tests::one_stale_span_among_verified_ones_refuses_certification ... FAILED
test handlers::file_entities::tests::a_span_derived_from_other_bytes_cannot_certify_the_enumeration ... FAILED
panicked at crates/kin-mcp/src/handlers/file_entities.rs:1527:9:
a full parse of bytes the tree no longer holds must not certify
  left: Bool(true)
 right: Bool(false)

test result: FAILED. 21 passed; 2 failed; 0 ignored; 0 measured; 896 filtered out
```

`ARM_RC=101`. The other 21 stayed green, which is what a targeted arm should do. Restored by exact
swap; porcelain empty and head unmoved afterwards.

**Check set**, read by name over the full run set at the head with `per_page=100`, 47 listed against
`total_count` 47. Five of the seven required contexts are green: `DCO Sign-off`, `PR text hygiene`,
`cargo-deny`, `gitleaks (full history)` and `MCP surface contract`. `Fast gate lint and policy` was
in progress. `Fast gate build and tests` had not been created yet, which is not a block: it is the
roll-up job at `ci.yml:1474` that reports after the three `Fast gate test shard` runs, and it
appeared that way on `950fb0bf8`, the sha #1513 landed. Nothing failed. `Product Acceptance` reads
`skipped`, as it does on every PR, and `CodeQL` was `neutral` with `Analyze (rust)` still running,
so it is not readable yet and neutral is not a failure.

Two follow-ups, neither a blocker and neither new to this split. `merge_file_coverage_classes` and
`file_entities_counted` disagree on precedence when a file is both `Partial` and stale, which is the
real shape of the restart window after this fix: one names `file_spans_stale` and drops the
parse-state cause from `limits` entirely, the other reports `floor_reason: file_parsed_partial`, so
a reader gets two different reasons for one answer. And the `Partial` reason string formats
`"{stale} of {}"` against `entities.len()`, the full unfiltered query result, while the numerator
counts only `Source`-role entities carrying spans, so a real file will read something like "3 of 40".

PR #1528, opened 2026-09-05
Settle a merged graph after finalization applies its live delta

troyjr4103:
FIR-3242.

A published merge recorded its relation census, and queued its convergence sweep, against the graph
as it stood BEFORE the merged live delta was applied. Two entities against three on a two-file
fixture. This settles both after the finalization that installs that delta, keeps a sweep request
the running pass cannot cover instead of dropping it, stops reading a dead enrichment channel as a
queued sweep, and retires the enrichment evidence the merge's own sweep would otherwise skip.

Follows kin #1523, which added the settle and the census. An independent source review of that
squash from source and raised three P1 convergence gaps
(`.kin-coord/reports/mergeedges-independent-review-20260905.md`). All three are real. A fourth fell
out of the second.

## Measured first, on the landed tree

Every arm below ran on `34d63ba76` plus the new tests and two behaviour-preserving seams the tests
need to call into (`mark_files_enriched` and `file_already_enriched` become `pub(crate)`, and the
sweep worker's three completion writes move verbatim into a named `complete_lsp_sweep`). Every arm
ran `cargo build -p kin-daemon --bin kin-daemon` first, because
`cargo test -p kin-cli --test <name>` does not rebuild the daemon and the harness reuses any daemon
binary whose build stamp is compatible.

| claim | falsifier | landed tree | after the fix | with the fix reverted |
|---|---|---|---|---|
| census and sweep see the pre-finalization graph | `a_published_merge_records_the_census_of_the_graph_it_finalized` | RED | ok | RED |
| a request made while a sweep runs is dropped | `a_merge_admitted_while_a_sweep_runs_gets_one_after_that_sweep_ends` | RED | ok | RED |
| a closed channel reads as a queued sweep | `a_merge_whose_enrichment_worker_is_gone_names_the_gap_rather_than_reporting_convergence` | RED | ok | RED |
| a durable marker lets the sweep skip merged-in work | `a_merge_retires_the_enrichment_evidence_its_own_sweep_would_skip` | RED | ok | RED |
| a parked merge must settle nothing (this PR's own hazard) | `a_parked_merge_records_no_census_because_it_published_no_merged_graph` | ok | ok | RED |

### The census arm, on the landed tree

```
thread 'a_published_merge_records_the_census_of_the_graph_it_finalized' panicked at
crates/kin-cli/tests/repository_merge_resolution.rs:2480:5:
assertion `left == right` failed: the record has to describe the graph this merge published, and
this one describes the graph before the merged live delta was applied: the merged tree holds 3
entities and the record claims Some(2)
  left: Some(2)
 right: Some(3)

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 35 filtered out; finished in 4.04s
```

The merged tree carries `src/feature.rs`, which the target branch never had, and the record written
by this merge does not know it exists.

### The other three, on the landed tree

```
test result: FAILED. 24 passed; 3 failed; 0 ignored; 0 measured; 1406 filtered out; finished in 2.26s

panicked at crates/kin-daemon/src/repository_merge.rs:4487:9:
the running pass could not see this merge, so ending it has to hand the worker the merge's own
sweep; before this the request was dropped on the floor and the merged tree waited for a command
nobody knew

panicked at crates/kin-daemon/src/repository_merge.rs:4512:13:
a merge with nothing left to run its sweep must not publish that gap quietly

panicked at crates/kin-daemon/src/repository_merge.rs:4551:9:
a file the merge did not touch can still have gained a cross-file edge through it, and a sweep
that skips the file cannot derive one; the evidence has to be retired so the merge's own sweep
visits it
```

The 24 that passed include `settling_a_merged_graph_queues_the_sweep_and_records_the_census`, the
control: on an idle channel the merge does queue its sweep, so the three failures are about the
states that test never entered.

## What changed

**Settle where the merged graph is live.** `plan_and_publish` installs the merge's immutable changes
into the live query graph, which is history and revision lineage; the merged entity and relation
table arrives only when `finalize_local_repository_commit` derives the live correction and applies
it, one call later in `execute`. Both settle calls move to the two top-level executors, after the
finalization and still inside the persistence and graph-authority locks. Queueing the sweep from
there is the second half of it, not a side effect: the worker snapshots every entity once when it
starts, so one queued against the pre-finalization graph can snapshot entities the merge already
replaced. `execute` still returns early for an already-up-to-date merge, which publishes nothing.

**Keep a request the running pass cannot cover.** `DaemonState` gains one coalesced
`lsp_sweep_pending` bit and `request_lsp_sweep_after_merge`. Queueing a second sweep beside the
running one was the alternative and it is wrong: a caller waits on `lsp_sweeps_completed`, so two
passes over one graph let a waiter see the first finish and return over a graph the second is still
mutating. The worker drains the bit at the tail of its pass. The demand is published before the
attempt to queue, so a pass that ends between the two calls drains it rather than racing past it.

**Name every send arm.** `queue_lsp_sweep` matched only `TrySendError::Full`, so a channel whose
receiver a supervisor halt had already dropped fell through to `true` and the merge reported a
convergence it had not queued and could not queue. A `false` is now a channel that can never
deliver, and the merge says so with its own recovery line.

**Settle nothing on a merge that published nothing.** A merge that PARKS its conflicts also
returns an execution with a receipt and finalizes like any other repository transition, so it
reaches the executor the settle moved into; the same holds for `ResolveAction::Settle` and
`ResolveAction::Abort`. Moved naively, a parked merge would have recorded a census of the UNMERGED
graph under this merge's timestamp and retired every file's enrichment evidence. `MergeExecution`
and `ResolveExecution` carry `published_merged_graph`, true only in `publish` and
`publish_resolved_merge`, and the settle is gated on it. That case was green before this PR, so its
lock is listed above with a green landed column and a red reverted one.

**Retire the evidence the merge's own sweep would skip.** The marker means "the enrichment for this
file is durable", and the sweep skips a marked file before it asks a language server anything. A
merge makes that false for files nothing edited: a caller on one branch and its callee on the other
produce an edge that exists only once the merge exists, and the caller's file is unchanged by the
merge and was legitimately marked when it was swept. Retiring only the changed files does not close
it, because the answer that went stale is a cross-file one whose target moved elsewhere. The whole
set goes, through the `retire_enrichment_marker` that already exists.

## Falsification

Each fix was reverted and the test written for it went red again.

```
# the merge settles before finalization again
assertion `left == right` failed: ... the merged tree holds 3 entities and the record claims Some(2)
  left: Some(2)
 right: Some(3)
test result: FAILED. 0 passed; 1 failed; 35 filtered out; finished in 4.05s

# complete_lsp_sweep stops draining the pending demand
the running pass could not see this merge, so ending it has to hand the worker the merge's own
sweep; before this the request was dropped on the floor and the merged tree waited for a command
nobody knew
test result: FAILED. 0 passed; 1 failed; 1432 filtered out; finished in 1.19s

# queue_lsp_sweep reports a closed channel as queued
a merge with nothing left to run its sweep must not publish that gap quietly
test result: FAILED. 0 passed; 1 failed; 1432 filtered out; finished in 1.02s

# the merge stops retiring the enrichment evidence
a file the merge did not touch can still have gained a cross-file edge through it, and a sweep
that skips the file cannot derive one; the evidence has to be retired so the merge's own sweep
visits it
test result: FAILED. 0 passed; 1 failed; 1432 filtered out; finished in 1.61s

# the settle stops asking whether a merged graph was published
a parked merge composed no merged relation set, so it has none to record; a record written here
would describe the graph before the merge and carry this merge's timestamp
test result: FAILED. 0 passed; 1 failed; 36 filtered out; finished in 3.99s
```

Backups by copy and restore by copy, never `git checkout -- <file>`, which restores HEAD. Every
mutated file was compared against its backup afterwards and matched byte for byte.

## Verification

Every command through the fleet's builder slot, at the committed head on a
clean tree.

```
cargo test -p kin-daemon --lib
test result: ok. 1430 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 285.67s

cargo test -p kin-cli --test repository_merge_resolution
test result: ok. 37 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 33.94s

cargo fmt --all -- --check
clean
```

`bin/kin-precheck kin --repo-dir kin=<this worktree>` at head `2f1c08095`, 0 behind main, is running as this PR opens; its `21 gates enumerated from the check job of .github/workflows/ci.yml` line and its tally follow in a comment.

One thing worth knowing for whoever runs these next: an arm that runs the whole `kin-daemon` lib
suite and then the merge integration file in ONE heavy slot fails two integration tests on
contention. Both pass alone and together in a whole-file run of their own, and one of them,
`an_untouched_entity_differs_between_branches_only_in_its_span`, performs no merge at all, so no
path this change touches is on it. Split the arms before believing a regression there.

`crates/kin-daemon/src/repository_merge_state.rs` is three lines at the resolve path's own
finalization seam. The resolve path carries the identical defect and there is no way to reach it
from the merge path; no open kin PR touches that file.

Report: `.kin-coord/reports/showhn-mergeedges2-20260905.md`.

### DIFF
--- a/crates/kin-cli/tests/repository_merge_resolution.rs
+++ b/crates/kin-cli/tests/repository_merge_resolution.rs
@@ -2422,3 +2422,115 @@ fn authored_merge_accepts_in_place_input_and_preserves_unrelated_or_later_edits(
         b"keep this\n"
     );
 }
+
+/// FIR-3242. The census a merge records has to be a reading of the graph the
+/// merge left live.
+///
+/// `settle_merged_graph` ran from inside the publication path, and the
+/// publication path returns to `execute`, which is where
+/// `finalize_local_repository_commit` derives and applies the merged live
+/// entity and relation correction. The `create_change` calls above the settle
+/// install immutable history and revision lineage, not the merged entity and
+/// relation table, so the record carried this merge's timestamp and source over
+/// a reading of the PREVIOUS live relation set. Every later comparison is
+/// judged against that baseline, so a kind this merge introduced can disappear
+/// without the store ever having recorded that it existed.
+///
+/// Read off disk, with enrichment off, and needing no barrier: the merged tree
+/// holds a source file the target branch never had, so the merged graph holds
+/// strictly more entities than the graph before it, and a record written after
+/// finalization has to show that. A record written before finalization holds
+/// the pre-merge count instead, and nothing between the merge and this reading
+/// can move either side.
+#[test]
+fn a_published_merge_records_the_census_of_the_graph_it_finalized() {
+    let root = tempdir().expect("temp root");
+    let repo = disjoint_repository(root.path());
+    let runtime = common::IsolatedDaemonRuntime::new(&repo);
+    ok(
+        &run_kin_without_enrichment(&runtime, &repo, &["init", ".", "--json"]),
+        "kin init",
+    );
+    let layout = kin_core::KinLayout::discover(&repo).expect("discover exact layout");
+    assert!(
+        kin_core::relation_census::read(&layout)
+            .recorded()
+            .is_none(),
+        "the recorder runs at the end of a sweep and at the end of a commit, and this init is \
+         neither, so the record read below is the one this merge wrote and nothing else"
+    );
+
+    let merged = ok(
+        &run_kin_without_enrichment(&runtime, &repo, &["merge", "feature"]),
+        "kin merge",
+    );
+    assert!(
+        merged.contains("Merged refs/heads/feature into refs/heads/main"),
+        "the fixture has to publish rather than park, or nothing below is about a merge: {merged}"
+    );
+
+    let recorded = kin_core::relation_census::read(&layout)
+        .recorded()
+        .cloned()
+        .expect("a published merge records the baseline the next comparison is judged against");
+    let published = graph_at(&layout, &branch_change(&layout, "main"));
+    let published_entities = published
+        .entities
+        .values()
+        .filter(|entity| !kin_index::is_external_reference_target(entity))
+        .count() as u64;
+
+    assert_eq!(
+        recorded.entities,
+        Some(published_entities),
+        "the record has to describe the graph this merge published, and this one describes the \
+         graph before the merged live delta was applied: the merged tree holds {published_entities} \
+         entities and the record claims {:?}",
+        recorded.entities
+    );
+}
+
+/// FIR-3242. A parked merge settles nothing, because it published no merged
+/// graph.
+///
+/// A merge that parks its conflicts still writes a merge transaction record and
+/// finalizes like any other repository transition, so it reaches the same
+/// executor the settle now runs in. It composed no merged entity and relation
+/// table: a census recorded there would describe the unmerged graph under this
+/// merge's timestamp and source, and retiring every file's enrichment evidence
+/// would buy a full re-derivation for a merge that published nothing.
+///
+/// Green before the settle moved, because `open_conflicted_merge` reaches a
+/// `bail!` rather than the publication path the settle used to live in. This is
+/// the lock that keeps it green.
+#[test]
+fn a_parked_merge_records_no_census_because_it_published_no_merged_graph() {
+    let root = tempdir().expect("temp root");
+    let repo = conflicting_repository(root.path());
+    let runtime = common::IsolatedDaemonRuntime::new(&repo);
+    ok(
+        &run_kin_without_enrichment(&runtime, &repo, &["init", ".", "--json"]),
+        "kin init",
+    );
+    let layout = kin_core::KinLayout::discover(&repo).expect("discover exact layout");
+    assert!(
+        kin_core::relation_census::read(&layout)
+            .recorded()
+            .is_none(),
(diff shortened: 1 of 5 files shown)


troyjr4103:
`bin/kin-precheck kin --repo-dir kin=<this lane worktree>`, named by absolute path through `.kin-coord/bin/heavy-slot.sh mergeedges2 kin`, at the committed head on a clean tree:

```
kin-precheck: repo=kin tree=/Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/mergeedges2-kin sha=2f1c08095a2f0d4d061877698094564302dd8a45 branch=chore/lane-mergeedges2-kin
kin-precheck: 21 gates ran, 21 passed, 0 failed, on kin 2f1c08095a2f0d4d061877698094564302dd8a45
```

The three steps the check job runs that precheck deliberately does not are the language-server provisioning step, which gates nothing; `release-policy-gate.sh`, which routes on a pull-request context a local run cannot supply truthfully; and `falsify-hydration-semantics.py`, which the check job runs against a poisoned copy the workflow builds first.

PR #1534, opened 2026-09-05
Remove both fallback credential files and report logout revocation failures

troyjr4103:
FIR-3257.

`kin auth logout` left the credential behind on the tier the product itself prints a PLAINTEXT
warning for, and printed that it had removed it. `store_credential` writes that tier to the
encrypted path's `.json` sibling and `load_credential` reads back exactly that expression, while
`delete_credential` removed only the `.age` form. So the next load picked the credential straight
back up. The logout also ignored the server's answer before printing success, so a failed revocation
read to the user exactly like a successful one.

## Measured first, at the base sha

`7a1db4594` plus the two new tests and nothing else:

```
test commands::auth::tests::a_logout_leaves_another_accounts_credential_alone ... ok
test commands::auth::tests::a_logout_removes_the_plaintext_credential_a_later_load_would_read ... FAILED
test result: FAILED. 11 passed; 1 failed; 0 ignored; 0 measured; 2189 filtered out; finished in 3.52s

panicked at crates/kin-cli/src/commands/auth.rs:855:9:
a logout that leaves a credential a later load reads has not logged anybody out
```

The control passed in the same run, which is what makes the failure mean what it says: removing one
account's credential already left another account's alone, so the red row is about the tier, not
about the file naming.

## What changed

**Every persisted form, and only this account's.** `persisted_credential_paths` derives both names
from the same call and the same `with_extension` the writer and the reader use, so the deletion set
cannot drift from what a later load would read. Both names are `account_key(base_url)`, so another
account is a different file and is never in the list.

**Keyring outcomes are reported, and absent is not a failure.** `delete_credential` returns what it
removed and what it could not. `keyring::Error::NoEntry` is nothing to do: a credential stored in a
file has no keyring entry, and calling that a half-broken logout would make the warning meaningless.
A refusal or an unopenable keyring is reported.

**A logout claims only what happened.** Deleting a file on this machine revokes nothing.
`revocation_outcome` reads the server's answer, or the absence of a stored credential, and
`logout_lines` turns that plus the local removal into what the reader is told. The failing row says
the session was NOT revoked, names why, and points at the dashboard, and it does not carry the
sentence a reader would take as revocation. Local removal runs whatever the server said, because a
session this machine cannot revoke is still one it should not keep the key to, and a local removal
that failed exits non-zero.

Both decisions are pure, so every row runs with no network and no keyring. That matters here:
`keyring_enabled()` is `!cfg!(test)`, so the real keyring is unreachable from these tests. The
keyring rows are covered as decisions and are not claimed as end to end.

## Falsification, including one mutant that proved nothing

```
### MUTANT A: delete_credential removes only the encrypted path again
test result: FAILED. 0 passed; 1 failed; 2205 filtered out; finished in 0.03s

### MUTANT B: the logout line claims revocation whatever the server said
test result: FAILED. 0 passed; 1 failed; 2205 filtered out; finished in 0.02s

### MUTANT C: a keyring refusal is swallowed again
test result: ok. 1 passed; 0 failed; 2205 filtered out; finished in 1.22s

### MUTANT C2: logout stops telling the reader a local removal failed
test result: FAILED. 0 passed; 1 failed; 2205 filtered out; finished in 0.02s
```

Mutant A is the falsification the ticket names. Mutant C SURVIVED and is reported rather than
dropped: it removed the error arm of the FILE deletion, which no test reaches, because making
`fs::remove_file` fail deterministically inside a tempdir is awkward. A surviving mutant is a hole
in the coverage, not a pass, so the honest reading is that the file-removal failure path is covered
as a decision through `logout_lines` and not end to end. C2 replaced it and bites.

## Verification

```
cargo test -p kin-cli --lib commands::auth
test result: ok. 17 passed; 0 failed; 0 ignored; 0 measured; 2189 filtered out; finished in 4.96s
```

Twelve before, seventeen now: the plaintext falsifier, the another-account control, the encrypted
tier, both forms at once, the empty store, and the two decision tables. No test uses a real
credential, a real keyring or a real session; every fixture is a tempdir the product's own locator
resolves into.

`bin/kin-precheck kin --repo-dir kin=<this worktree>`, named by absolute path, on a clean tree at
`efdf42e16ae3ddc971f3ba614f63bb1fb9ebb5ec`, which is this commit before it was rebased onto current
main: `21 gates ran, 21 passed, 0 failed`. The run at the rebased head `f3270e025` is still going as
this opens, so its tally is not quoted here; it follows as a comment with its actual output.

### DIFF
--- a/crates/kin-cli/src/commands/auth.rs
+++ b/crates/kin-cli/src/commands/auth.rs
@@ -413,18 +413,78 @@ pub fn hosted_credential_state(
     Ok(absent_state(keyring_read))
 }
 
-fn delete_credential(base_url: &str) -> Result<()> {
+/// Every persisted form one account's credential can take.
+///
+/// The plaintext tier writes to the encrypted path's `.json` sibling and
+/// `load_credential` reads it back from exactly that expression, so a deletion
+/// naming only one of the two leaves the other for the next load. Derived from
+/// the same call and the same `with_extension` here, so the set can never drift
+/// from what the writer and the reader use.
+///
+/// Both names are `account_key(base_url)`, so another account's credential is a
+/// different file and is not in this list at all.
+fn persisted_credential_paths(base_url: &str) -> Result<[PathBuf; 2]> {
+    let encrypted = fallback_credential_path(base_url)?;
+    let plaintext = encrypted.with_extension("json");
+    Ok([encrypted, plaintext])
+}
+
+/// What removing one account's local credential actually managed to do.
+///
+/// A list rather than a bool, because "the keyring entry was not there" and
+/// "the keyring refused to delete it" are different facts and only the second
+/// means a credential may still be readable. Reporting them as one would let a
+/// logout that removed nothing print the same line as one that removed
+/// everything, which is the shape this whole change exists to end.
+#[derive(Debug, Default, PartialEq, Eq)]
+struct LocalRemoval {
+    /// Forms that were present and are now gone.
+    removed: Vec<String>,
+    /// Forms that were present and could not be removed, and why.
+    failures: Vec<String>,
+}
+
+impl LocalRemoval {
+    fn is_clean(&self) -> bool {
+        self.failures.is_empty()
+    }
+}
+
+fn delete_credential(base_url: &str) -> Result<LocalRemoval> {
     let key = account_key(base_url);
+    let mut outcome = LocalRemoval::default();
+
     if keyring_enabled() {
-        if let Ok(entry) = keyring::Entry::new(KEYRING_SERVICE, &key) {
-            let _ = entry.delete_credential();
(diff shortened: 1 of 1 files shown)


troyjr4103:
`bin/kin-precheck kin --repo-dir kin=<this worktree>`, named by absolute path through the fleet heavy-slot wrapper, at the rebased head on a clean tree. This is the tail the body said would follow.

```
kin-precheck: repo=kin tree=/Users/troyfortinjr/GitHub/kin-ecosystem/.kin-coord/worktrees/mergeedges2-kin sha=f3270e025cf52e875a50a2ae387ad38f6a0120ba branch=chore/lane-mergeedges2-auth
kin-precheck: 21 gates ran, 21 passed, 0 failed, on kin f3270e025cf52e875a50a2ae387ad38f6a0120ba
```

PR #1601, opened 2026-09-08
Refresh verified C declarations during incomplete parsing

troyjr4103:
A local-variable rename in hiredis `sds.c` previously left every entity on the old source digest because unrelated C constructs still produced parse errors. Refresh an existing function only when old and new full-file CAS parses prove a complete, unambiguous declaration with stable declaration and preprocessing context. The retained fixture has 25 errors both before and after the rename; none are suppressed.

Introduce `PartiallyUpdated` so admitted functions retain their identity and use the current CAS span and body while file coverage remains incomplete. Unmatched and error-affected entities retain their previous payloads. Partial admission adds or removes no entities and performs no whole-file relation retirement. Existing direct-call evidence moves only when its old and new sites and same-file resolution are independently proven unique. Unsupported or ambiguous evidence retains the entire old caller. Relation identity and non-location evidence remain unchanged, and subscribers receive the modified relation after its entity event.

Persist incomplete coverage before graph publication and independently verify admitted payloads, proof metadata and call coordinates at commit. Commit/reopen controls check the current `sdsMakeRoomFor` body, stable identity, current reference lines, retained errors and unchanged refusal of stale retained source. Intentionally retained callers still allow artifact continuity. A later clean parse restores complete coverage and normal removal behavior.

At `ce1861502cb7b5f6f4c6b7289cf37570c6bd4e8c`, precheck passed all 21 gates, including formatting, clippy and source-search policy. Affected-crate suites are running. Focused partial-admission tests pass: 4 daemon and 12 reconcile tests. Thirteen mutation controls were caught by real test failures and explicitly restored, including old call coordinates, skipped commit-time coordinate verification and an incorrectly blocked retained-caller commit. Full touched-crate tests on the preceding head had 3681 passes, three watcher failures and 20 ignored tests. Its precheck passed 21/21; parity failed brownfield check 9 because an unreferenced initialization staging directory survived. These failures remain recorded and are not waived.

### DIFF
--- a/crates/kin-core/src/reference_coverage.rs
+++ b/crates/kin-core/src/reference_coverage.rs
@@ -1184,10 +1184,10 @@ impl LanguageParseCoverage {
             )
         };
         format!(
-            "{}: {} of {} admitted {} files did not parse as written, so any entities the graph \
-             still holds for them came from an earlier parse of bytes that are gone, and a file \
-             the graph never parsed is absent from it entirely{named}. Fix the syntax and the \
-             next admission re-derives them.",
+            "{}: {} of {} admitted {} files did not parse as written{named}. Some entities may \
+             be retained from earlier parses while independently verified entities are current. \
+             File coverage remains incomplete. Fix the syntax and the next admission re-derives \
+             the complete file.",
             crate::retained_parse::RETAINED_OBSERVATION,
             self.retained,
             self.tracked,
@@ -2617,7 +2617,8 @@ mod tests {
         // True of both populations. A file created with a typo has no earlier
         // parse, so the clause about what the graph holds has to be conditional.
         assert!(
-            lines.contains("any entities the graph still holds"),
+            lines.contains("Some entities may")
+                && lines.contains("File coverage remains incomplete"),
             "the sentence may not diagnose one member of the set: {lines}"
         );
 
--- a/crates/kin-core/src/reference_coverage.rs
+++ b/crates/kin-core/src/reference_coverage.rs
@@ -1186,8 +1186,8 @@ impl LanguageParseCoverage {
         format!(
             "{}: {} of {} admitted {} files did not parse as written{named}. Some entities may \
              be retained from earlier parses while independently verified entities are current. \
-             File coverage remains incomplete. Fix the syntax and the next admission re-derives \
-             the complete file.",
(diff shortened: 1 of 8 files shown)


troyjr4103:
Closing as superseded. Every change on this branch is already on `main`, landed through #1608 with #1606, #1609 and #1622 around it.

Checked at `origin/main` two ways, against this PR's head `ce1861502` and its merge base `cefb3b8d4`.

At function granularity, all 34 functions this branch adds are declared on `main`: `loop_runner.rs` 4, `repository_commit.rs` 2, `pipeline.rs` 4, `c_lang.rs` 6, `reconciler.rs` 18, none absent. The check's own controls hold, parsing 136 function names out of `main`'s `reconciler.rs` and correctly reporting `reconcile_partial` present and a deliberately fake name absent.

At line granularity, 10 of the 3178 added lines are absent from `main`, and each is a syntactic difference rather than missing behavior:

- `reconciler.rs`, two production lines. This branch dispatches with `if let ParseState::Incomplete { error_ranges } = &indexed.parse_state { return self.reconcile_partial(...); }`. `main` writes the same dispatch as a let-else inside `reconcile_observed_edit` at `reconciler.rs:735` and `:749`, and adds the `BrokenAstBehavior::Reject` arm around it, so `main` is a superset.
- `reconciler.rs` one test line and `repository_commit.rs` seven test lines, all call and argument lines inside test bodies whose enclosing functions `main` declares, differing only in local binding form. `reconcile_indexed_content` has the identical four-argument signature at both heads.

Supporting identities: the C fixture `crates/kin-parser/tests/fixtures/c/hiredis-sds.c` is the same blob at both heads, `c0388bdd569e9e8d2bb8f5b3cc0f666899248742`, and `ReconcileOutcome::PartiallyUpdated` carries the same five fields and doc comment on `main`.

The separate rescan-loss class stays open and is not part of this PR. It is documented in `continuity-evidence/rescan-followup-read.md` and confirmed still present at `origin/main`: `FileEvent` has only `Changed` and `Removed`, `classify_event` filters paths before matching `event.kind` so a pathless rescan is dropped at the filter, and `need_rescan()` is read only in the readiness probe path. That needs its own lane and a deliberate contract decision.

The branch and the preserved worktrees are left as they are.

PR #1604, opened 2026-09-08
Tolerate brownfield:4's tracked FAIL and search clip witnesses (FIR-2464)

troyjr4103:
Main's `Product Acceptance` has been red since `9e45cc6c8` on one finding, `brownfield:4 FAIL`,
and a red main run stops every release cut. This makes it green without hiding anything, and
fixes a real logical gap in the witness machinery that put it there.

Tracked as FIR-2464. Alarm issue #1603.

## The measurement, before any change

Run 34225725954 on `9e45cc6c8` (kin #1596) concluded FAILURE. The run before it, 34224296350
on `72414aa80`, was green with `ALLOWED brownfield:4 is UNREADABLE (FIR-2593 ...)`. Both read
out of each run's own `acceptance-reports` artifact, not off a rollup.

| | green 34224296350 | red 34225725954 |
|---|---|---|
| sha | `72414aa80` | `9e45cc6c8` |
| brownfield tally | `12 PASS, 1 UNREADABLE` | `1 FAIL, 12 PASS` |
| check 4 | `UNREADABLE` | `FAIL` |

The green run's check 4, verbatim:

```
CHECK 4 FIR-2464 UNREADABLE the walk came back truncated (truncated=True, clipped_steps=[{
'continued_below': True, 'dropped_callees': 2, 'dropped_callers': 0, 'dropped_crossing_file': 2,
'entity_id': 'a6ba009a-0005-4f1c-8592-d84fef2b5f77', 'entity_name': 'application',
'limit_per_step': 25, 'step': 7}]), so an absent edge cannot be told from a dropped one
```

The red run's, with the witness receipt #1596 added as its first assertion:

```
PASS  5e6f2471-e27e-47bc-9947-2d67a5d844b3: complete depth=1/out/limit=50/max_chars=60000
      witness, 33 entities, 45 relations, 32 trace-eligible destinations
PASS  no fabricated callee is counted (34 of 39 steps counted)
PASS  both real callees app.enabled, finalhandler are present
FAIL  the hand-off this.router.handle at lib/application.js:177, the last line of app.handle
      and the edge the question is about, is absent from the walk; steps are [...]
```

## What actually changed, and what did not

The walk was still clipped on the red run. `complete_trace_clips` returns early and emits no
receipt unless the payload is truncated or carries `clipped_steps`, and the red run carries a
receipt. Both runs clipped. The only change is that #1596 replaced the truncation
short-circuit with a witness, so the assertion is now reached.

The `or basename_of(n) == "handle"` disjunct #1596 also removed is a red herring. Nothing in
the 39-step walk has basename `handle`, so the old matcher matched nothing either.

## The absence is real, and no truncation causes it

Measured by dumping the raw payloads check 4 grades, from a local parity build at `9e45cc6c8`
that reproduces the ubuntu failure byte for byte, same message and same witness numbers.

`app.handle`'s own fanout is eight steps, all `parent=0`, against a per-step cap of 25, so the
focal was never clipped. The one clip is on `application`, a depth-1 node, and its witness comes
back complete: 33 entities against `entity_count` 33, 45 relations against `relation_count` 45,
`truncated=False`. Over all 39 walk steps and all 45 witness destinations, nothing matches
`router.handle`.

The reason is not Kin failing to walk. express is pinned at 5.2.1, where the router is a
dependency rather than `lib/router/index.js`, so this hand-off targets
`Router.prototype.handle` at `node_modules/router/index.js:149` while the fixture admits the
repository only: `sweep concluded clean: enriched=140 of files=140`, against 141 tracked `.js`
paths. The graph has an entity for `Router`, which is step 38 of the walk, and none for its
`handle` member. The only `handle` definition in the whole pinned tree is `app.handle` itself
at `lib/application.js:152`.

Surfacing that hand-off is FIR-2464 itself, and it is unfixed. So `brownfield:4 FAIL` is
honest. #1596 removed a blindfold rather than misclassifying anything, and turning it back to
UNREADABLE would hide a provable gap behind a truncation that provably does not cause it.

## Change 1: a FAIL allowance the gate can hold honestly

`bin/kin-parity` has tolerated this exact FAIL under this exact ticket since it was written
(`ALLOWED brownfield check 4 FAIL, tracked as FIR-2464`), and its own allowance comment says it
names both tickets "rather than letting the two gates diverge quietly". The gate of record had
no way to agree, because its docstring said a FAIL "is never tolerated and has no allowance".

`--allow-fail SUITE:CHECK=TICKET|SUBSTRING|WHY` closes that, and every field is required:

- **The ticket**, because a tolerated FAIL nobody is tracking is a defect nobody is fixing.
- **The substring**, because an allowance keyed on the check id alone is a check that cannot
  fail. `brownfield:4` can go red on a fabricated callee, on an unwitnessable clip, or on the
  absent hand-off FIR-2464 tracks, and only the last is tolerated here.
- **The reason**, printed on every run.

The substring is matched against **every failing assertion in the row**, not against the row's
`detail`. `detail` is only the first failing assertion, so a check that grew a second, unrelated
failure would otherwise still present the tolerated sentence and be waved through. A failure
that does not carry it is reported normally and gets a warning saying the allowance was
considered and did not match, so a refused allowance is never quiet. A tolerated FAIL reaches
the run summary as a `::notice::` annotation.

The FIR-2593 UNREADABLE line beside it is corrected too. It claimed "Any FAIL or other
UNREADABLE still fails the gate", which is no longer true of the FAIL this new line names.

## Change 2: search a clip's witness before calling a required edge absent

Separate from the red, and worth landing on its own. `complete_trace_clips` enumerates the
clipped node's outgoing destinations, tests those names against the fabricated-callee list, and
throws them away. Check 4 then spends that same `complete` flag on two ABSENCE assertions
searched over the clipped walk alone.

The rows the cap dropped are inside the neighborhood the witness enumerates, so a required edge
among them reads as absent from the walk while the graph carries it. That is exactly the
distinction FIR-2593's allowance says this check cannot make. #1596 built the machine that can
tell them apart and did not wire it to the question.

So the witnessed names come back and the positive assertions search the walk plus the witness.
Names are recorded before the `name_only` skip, because the walk's own name list carries
name_only steps too: surfacing an edge and counting it are different questions, and only the
first is asked here. On express this changes no verdict.

## The gate, graded against the red run's own artifacts

Replayed through acceptance.yml's OWN gate command rather than a hand-typed copy, so the YAML
quoting is part of what is under test.

```
$ replay-workflow-gate.py <run 34225725954's uploaded acceptance-reports>
ALLOWED FAIL brownfield:4, tracked as FIR-2464 (Measured on the red main run 34225725954 ...);
  every failing assertion carries 'the hand-off this.router.handle at lib/application.js:177'
::notice::ALLOWED FAIL brownfield:4, tracked as FIR-2464 ...
acceptance gate passed
=== gate rc=0 ===
```

And the control, the same artifacts with one extra fabricated-callee assertion added to check 4:

```
$ replay-workflow-gate.py <same artifacts, one second failing assertion>
UNMATCHED ALLOWANCE brownfield:4 carries a FIR-2464 FAIL allowance, but this failure does not
  carry 'the hand-off this.router.handle at lib/application.js:177' in every failing
  assertion, so it stands
::warning::UNMATCHED ALLOWANCE brownfield:4 ...
acceptance gate FAILED on 1 finding(s)
=== gate rc=1 ===
```

Without the allowance at all, the same real artifacts stay red:

```
::error::brownfield:4 FAIL the hand-off this.router.handle at lib/application.js:177, ...
acceptance gate FAILED on 1 finding(s)
=== gate rc=1 ===
```

Both allowances also coexist. With check 4 UNREADABLE the FIR-2593 line covers it and the gate
passes; with check 4 PASS both lines announce themselves stale and the gate passes.

## Falsification

Every arm ran on a copy, never on the tracked file, and each is red on the rows it exists for.

gate.py:

```
ARM 1  failure_matches returns True unconditionally     6 rows red
ARM 2  the TICKET_ID check removed                      4 rows red
ARM 3  the notice annotation dropped                    1 row  red
ARM 4  the FAIL allowance never applies                 6 rows red
control unmutated                                       acceptance gate: self-test passed
```

brownfield_repro.py:

```
ARM A  surfaced_by ignores the witness                  3 rows red
ARM B  witnessed.append moved after the name_only skip  1 row  red
control unmutated                                       Ran 22 tests in 0.004s  OK
```

## Where these rules are graded

`acceptance.yml`'s only job carries `if: ${{ github.event_name != 'pull_request' }}`, so
`gate.py --self-test` never runs on the pull request that could break it. So this PR drives it
from `scripts/acceptance/test_brownfield_trace_witness.py`, which `ci.yml`'s `fast-gate-lint`
job already runs on every PR. With gate.py's ticket rule removed, that file goes red on 2 rows.
That is the AGENTS.md rule applied: a class answered only by an acceptance case can be
reintroduced by a green PR and caught a release later.

## Verification

```
$ python3 scripts/acceptance/test_brownfield_trace_witness.py
Ran 24 tests in 0.012s
OK

$ python3 scripts/acceptance/gate.py --self-test
acceptance gate: self-test passed

$ python3 scripts/acceptance/brownfield_repro.py --self-test
kin-brownfield-repro: self-test passed
```

Local gate run, through `.kin-coord/bin/heavy-slot.sh acceptfix kin`, against this exact
pushed head:

```
$ bin/kin-precheck kin --repo-dir kin=<lane worktree>
kin-precheck: repo=kin tree=.../worktrees/acceptfix-kin
  sha=6dbf9706472ebacb8c5f2c42029d37782ce5c17a branch=chore/lane-acceptfix-kin
...
kin-precheck: 21 gates ran, 21 passed, 0 failed, on kin 6dbf9706472ebacb8c5f2c42029d37782ce5c17a
=== precheck exit rc=0 at 2026-09-08T15:39:45Z ===
```

Still 21 enumerated rather than 22, which confirms the new gate rows ride in the existing
`Test bounded brownfield trace witnesses` step in `fast-gate-lint` rather than adding a `check`
gate precheck would have to classify by name.

PARITY_PLACEHOLDER

The base is one commit behind `origin/main` (`e9ebaa118`), which touches only Rust and
`init_memory_repro.py`, none of the four files here. Hosted CI grades the merge ref, and
`merge land`'s BEHIND refusal covers `AGENTS.md` and `docs/traps.md`, neither of which this
touches.

Kin-Release-Intent: patch

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -1295,10 +1295,15 @@ jobs:
       # than the exit codes.
       # A suite's exit status is one lever with two settings, demand a clean
       # sweep or demand nothing, and neither is right while one check is blocked
-      # on something outside the change under review. gate.py fails on any FAIL
-      # with no allowance possible, fails on any UNREADABLE that is not named
-      # here with a reason, and refuses an allowance that points at a check the
-      # report does not carry.
+      # on something outside the change under review. gate.py fails on any
+      # UNREADABLE that is not named here with a reason, fails on any FAIL that
+      # is not named here with a ticket AND the words that failure carries, and
+      # refuses an allowance that points at a check the report does not carry.
+      # The FAIL allowance is the narrower of the two on purpose: a check has
+      # more than one way to go red, so one keyed on the check id alone would
+      # tolerate the next regression too. Every failing assertion in the row has
+      # to carry the substring, and a failure that does not is reported with a
+      # warning saying the allowance was considered and did not match.
       - name: Acceptance verdict, gated on the suite reports
         if: always()
         run: |
@@ -1330,4 +1335,5 @@ jobs:
             --report prose_parity=acceptance/prose_parity.json \
             --report merge_precedence=acceptance/merge_precedence.json \
             --report detached_head=acceptance/detached_head.json \
-            --allow-unreadable 'brownfield:4=FIR-2593: the express app.handle walk is truncated on ubuntu-latest, and its clipped_steps metadata confirms dropped callee and cross-file rows. This check therefore cannot distinguish an absent required edge from one the walk dropped. Any FAIL or other UNREADABLE still fails the gate, and the moment this check passes the allowance announces itself stale and can go.'
+            --allow-unreadable 'brownfield:4=FIR-2593: the express app.handle walk comes back truncated, and a clip nothing can witness leaves the check unable to tell an absent required edge from a dropped one. A clip whose complete neighborhood witness re-supplies the dropped rows is readable and is graded on its merits, so this line now covers only the unwitnessable case and the FAIL allowance beside it covers the graded one. Any other UNREADABLE still fails the gate, and the moment this check passes both allowances announce themselves stale and can go.' \
+            --allow-fail 'brownfield:4=FIR-2464|the hand-off this.router.handle at lib/application.js:177|Measured on the red main run 34225725954 and reproduced locally on 2026-09-08. The absence is real and no truncation causes it: app.handle offers eight callees against a per-step cap of 25 and was never clipped, and the one clip in the walk is on application, whose complete witness carries 45 relations and no matching name either. express is pinned at 5.2.1, where the router is a dependency rather than lib/router/index.js, so this hand-off targets Router.prototype.handle in node_modules/router/index.js while the fixture admits the repository only, 140 of 140 files. The graph has an entity for Router and none for its handle member. Surfacing that hand-off is FIR-2464 itself and is unfixed, and bin/kin-parity has tolerated this same FAIL under this same ticket since it was written, so this line stops the two graders of one report disagreeing.'
--- a/scripts/acceptance/brownfield_repro.py
+++ b/scripts/acceptance/brownfield_repro.py
@@ -1223,6 +1223,15 @@ def complete_trace_clips(suite, repo, payload, steps):
     The original trace remains the assertion surface. A neighborhood can prove
     its omitted candidates harmless, but a forbidden raw edge needs a focused
     trace before it can establish a counted-flow failure.
+
+    Returns (confirmations, receipts, witnessed). `witnessed` is every
+    destination name the witnesses enumerated, and the caller must search it
+    beside the walk. A witness that answers only the forbidden-callee question
+    cannot license a verdict on an absent required one: the rows the cap dropped
+    are inside that same neighborhood, so a required edge among them reads as
+    absent from the walk while the graph carries it. That is the exact
+    distinction FIR-2593's allowance says this check cannot make, and returning
+    the names is what lets it.
     """
     if (budget_cut(payload) or (payload.get("_kin") or {}).get("response", {}).get("bounded")
             or payload.get("steps_omitted")
@@ -1230,10 +1239,10 @@ def complete_trace_clips(suite, repo, payload, steps):
         raise ProbeError("trace output or total-step budget lost rows")
     clips = payload.get("clipped_steps") or []
     if not payload.get("truncated") and not clips:
-        return [], []
+        return [], [], []
     if not clips:
         raise ProbeError("truncated trace has no bounded fanout disclosure")
-    confirmations, receipts = [], []
+    confirmations, receipts, witnessed = [], [], []
     for clip in clips:
         index = clip.get("step")
         if (not isinstance(index, int) or index <= 0 or index > len(steps)
@@ -1277,9 +1286,14 @@ def complete_trace_clips(suite, repo, payload, steps):
             if relation.get("kind") not in ("Calls", "Imports", "References", "UsesType"):
                 continue
             eligible.add(dst)
+            name = by_id[dst]["name"]
+            # Recorded before the name_only skip below, because the walk's own
+            # all_names carries name_only steps too. The positive assertions ask
+            # whether Kin surfaced an edge at all, not whether it counted it, so
+            # the witness has to answer on the same terms the walk does.
+            witnessed.append(name)
             if relation["resolution"] == "name_only":
                 continue
-            name = by_id[dst]["name"]
             if not any(name_matches(name, bad) for bad in APP_HANDLE_FABRICATED):
                 continue
             focused = suite.cached(repo, "trace_data_flow",
@@ -1304,7 +1318,7 @@ def complete_trace_clips(suite, repo, payload, steps):
         receipts.append("%s: complete depth=1/out/limit=50/max_chars=60000 witness, %d entities, "
                         "%d relations, %d trace-eligible destinations"
                         % (focal, len(entities), len(relations), len(eligible)))
-    return confirmations, receipts
+    return confirmations, receipts, witnessed
 
 
 def check_4(suite):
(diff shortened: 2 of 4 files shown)


troyjr4103:
Landed and confirmed on main. Posting the two things the body promised, plus one finding this run turned up.

## Main is green

`c67efaa2c` is main's HEAD and its `Product Acceptance` run 34247007461 concluded **success** at 16:33Z (job 102131457566). The alarm issue #1603 closed itself at 16:33:50Z. Read out of the job's own log:

```
acceptance gate: self-test passed
ALLOWED FAIL brownfield:4, tracked as FIR-2464 (Measured on the red main run 34225725954 ...)
##[notice]ALLOWED FAIL brownfield:4, tracked as FIR-2464 (...)
brownfield: 1 FAIL, 12 PASS
acceptance gate passed
```

Both halves are there on the real runner: the gate's own self-test passed with the new rules, and the tolerated FAIL reached the run summary as a notice rather than sitting quietly in a log.

## Parity

`bin/kin-parity --checkout <lane worktree> --bin <release kin> --daemon <release kin-daemon>`, run 20260908T154440Z, through `.kin-coord/bin/heavy-slot.sh acceptfix kin`. Binaries are the release build from `9e45cc6c8`, this branch's base; the two commits on top change four files and every one is Python or a workflow (`git diff --name-only 9e45cc6c8..HEAD`), so those are the bytes under test, and parity prints that a passed-in binary's stamp is not compared against the checkout.

```
kin-parity: === magic ===
    | kin-magic-repro: 27 pass, 0 fail, 0 unreadable
kin-parity: === brownfield ===
    | CHECK 4 FIR-2464 FAIL the hand-off this.router.handle at lib/application.js:177, the last
      line of app.handle and the edge the question is about, is absent from the walk and from the
      45 witnessed destination(s) the clips dropped; steps are [...]
    | kin-brownfield-repro: 11 pass, 2 fail, 0 unreadable
kin-parity: ALLOWED   firstrun check 9 FAIL, tracked as FIR-2580
kin-parity: ALLOWED   brownfield check 4 FAIL, tracked as FIR-2464
kin-parity: FINAL FAIL in 1733.5s  (NON-CITABLE)
```

Check 4 now names the witness in its failure, which is the change working: the absence is stated over the walk AND the 45 destinations the clip dropped, rather than over the walk alone.

## The second failure is not this change, and it is not on the runner

Parity read `11 pass, 2 fail` rather than the baseline's `12 pass, 1 fail`. The extra one is brownfield check 2 on the `requests` fixture, and it flipped between two local runs of the **same binaries**:

```
baseline 13:11Z, kin built at 9e45cc6c8   check 2 PASS
  PASS  Session.send counted as a reference with reference_lines [784]

fix run 15:44Z, the same binaries passed in   check 2 FAIL
  FAIL  Session.send appears in neither references nor candidates;
        total_upstream=2, counted=1, withheld=0
```

It is not reachable from this diff. `complete_trace_clips`, `witnessed` and `surfaced_by` appear only inside `complete_trace_clips` and `check_4`; `check_2` is a separate function that shares none of them. And it passed on the runner for this exact landing: the acceptance log above reads `brownfield: 1 FAIL, 12 PASS`, so check 2 was green there.

So it is a local run-to-run condition on this box, on the same bytes, and I am not claiming a cause I did not measure. Flagged to the captain for a ticket rather than diagnosed here.

PR #1616, opened 2026-09-08
Validate and report hosted publication destination services

troyjr4103:
`gcs:<bucket>/<prefix>` says nothing about whether that bucket was the real service or an emulator standing in for it, so a hosted publication receipt could not tell the two apart. Every destination refusal also happened late, after a client existed and sometimes after an import had been spent on a destination nobody could reach.

`DestinationSpec::resolve` now settles every choice about where a publication points before anything is opened: the filesystem root's shape, the bucket and prefix grammar, the endpoint override, and the artifact-identifier budget. That last one is the load-bearing reorder. The identifier is composed from the scope AFTER a publication has already committed, so a prefix that cannot compose one used to produce a publication that succeeded and a receipt nobody could write. The resolved destination then hands out fresh backend handles, and an object-store one reports the service class off the value that built the client rather than deriving it a second time.

`destination_service` is emitted only for an object-store destination, so a filesystem run's record is byte-identical to what this adapter has always written. That matters because the hosted decoder refuses an unknown top-level field.

## Ordering

kinlab #508 teaches the hosted decoder this field and MUST be in production before this lands. #508 is green and waiting in its lane's queue.

## Verified by running, not by reading

```text
$ cargo test -p kin-daemon --features gcs --locked --lib gcs_endpoint
test result: ok. 15 passed; 0 failed; 1 ignored

$ cargo test -p kin-cli --locked --lib commands::hosted_publication
test result: ok. 11 passed; 0 failed

$ cargo test -p kin-cli --features gcs --locked --test hosted_publication_object_store
test result: ok. 3 passed; 0 failed

$ cargo test -p kin-cli --locked --test hosted_first_publication
test result: ok. 26 passed; 0 failed
```

The one ignored case is `a_seeded_emulator_serves_a_snapshot_round_trip`, which needs a running fake-gcs-server. No real object-store publication is claimed here.

Property by property. Production-ADC versus emulator classification, with precedence between the two levers and a positive control that the second is honored when the first is absent. Bracketed IPv6 both resolves and reaches a real IPv6 listener, and `http://[::1`, `http://[::1]x:4443`, `http://user:pass@host:4443`, `http://host:0` and `http://host/path` are each refused rather than falling through. Failure before irreversible publication, asserted as `!case.path("source").exists()` on a run against a closed port. Unchanged filesystem evidence bytes, asserted twice: a byte-exact encoder golden in the module, and an exact top-level key set over a real publication's output. And the object-store cases are behind the `gcs` feature with a CI step that runs them on the pull request, whose exact count assertion matches the file's three `#[test]` today.

## Falsification

Three arms, each breaking production behavior rather than an assertion, each killing exactly the cases written for it, restored green.

```text
ARM=A-invalid-override-selects-production   9 passed, 2 failed
  an_unparseable_lever_never_falls_back_to_the_real_service
  bracketed_ipv6_is_an_emulator_and_invalid_overrides_never_select_adc
ARM=B-budget-not-spent-at-resolution       10 passed, 1 failed
  an_identifier_that_cannot_be_composed_is_refused_before_a_publication_commits
ARM=C-service-key-always-emitted           10 passed, 1 failed  +  25 passed, 1 failed
  a_filesystem_evidence_record_is_byte_identical_to_what_this_adapter_has_always_written
  a_filesystem_publication_records_no_destination_service
ARM=D-restored                             11 passed, 0 failed  +  26 passed, 0 failed
```

Arm C is the one worth reading: dropping `skip_serializing_if` alone makes a filesystem record grow a `"destination_service": null` key, and both the unit golden and the integration key-set assertion catch it.

## The workflow change, called out

This raises `fast-gate-lint` (`Fast gate lint and policy`, one of main's seven required contexts) from 20 to 30 minutes, because the new step builds kin-cli under a second feature set that clippy's artifacts cannot serve. The step lives in that job rather than somewhere cheaper because the two names main requires are that one and the sharded test job, and a guard anywhere else is advisory.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -951,7 +951,13 @@ jobs:
     # Read the runner a run actually resolved from the jobs API `labels` field,
     # never from this line.
     runs-on: ${{ vars.KIN_HEAVY_RUNNER || 'ubuntu-latest' }}
-    timeout-minutes: 20
+    # 20 until the hosted object-store publication step below, which builds
+    # kin-cli under a second feature set that clippy's artifacts cannot serve.
+    # Raised rather than moving that step somewhere cheaper on purpose: the two
+    # names main requires are this one and the sharded test job, so a guard that
+    # lands anywhere else is advisory and an armed auto-merge would land past
+    # it red.
+    timeout-minutes: 30
     # FIR-2744. This job shares `check`'s cargo cache key, and with
     # `add-rust-environment-hash-key` off nothing else would notice the two
     # declaring different cargo profiles, so they must declare the same one or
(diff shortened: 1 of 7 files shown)


troyjr4103:
This predates the monorepo move and its change is not yet in kin-internal. Filed FIR-3666 to port it (branch fix/hosted-gcs-publication kept here); it depends on kinlab#508 landing first. Closing this PR; product work continues from the private monorepo.

PR #1738, opened 2026-09-11
Lock the Kin agent belt to Kin tools and let a mutation name its own change

troyjr4103:
A Kin agent with `KIN_AGENT_PURE_KIN=1` gets a belt with no `edit_file` and no `write_file` at all. There is one write tool on it, `kin_mutate`, which names the entity and stages and commits it in one call, and the change it publishes now carries the agent's own sentence instead of `MCP transaction <uuid>`.

That is the decision "the Kin agent is locked to Kin tools only", built on the product's own belt rather than proved with a driver written for the occasion. The honest sentence for any writeup: the local agent edits through Kin transactions, and MCP is how it queries the graph.

It starts from a snapshot of work that stalled unfinished (`backup/agentbelt-wip-20260911T2000Z`, commit `137410c5c`), and everything below is what had to change for that snapshot to hold.

## What changed, and why each piece is here

**The belt loses its file tools, and says so when asked for one.** `Belt::pure_kin` builds a belt with neither local tool; `Belt::route` answers `edit_file` and `write_file` with a refusal that names `kin_mutate` rather than a generic "no such tool", because a model that reaches for the old name has a next move and should be told it. `Belt::to_specs` stops emitting their function specs, which is what actually matters: a tool absent from the harness's own view but present in the tools array is a tool the model will call.

**`kin_mutate` is the one-shot front.** It begins, stages and commits in one round trip. Every way a commit can fail to land is followed by an abort, whether the daemon refused it, the daemon was gone or the transport broke, and if that abort cannot run or is refused, the answer names the transaction it left open instead of dropping the failure. The transaction primitives stay where they were: `belt::is_harness_owned` already hides every `kin_transaction_*` and `kin_session_*` tool from the model, so the ceremony never reaches it while the daemon keeps the multi-step path the contract documents for a refactor that has to stage, read a refusal and re-stage.

**A body Kin cut short can no longer be staged.** Search results, context packs and trace steps clip a rendered body at 40 lines or 2,400 characters (`MCP_SOURCE_MAX_LINES`, `MCP_SOURCE_MAX_CHARS`) and mark the cut. Staging that text replaces the entity's whole span with the part that fit, and the receipt still says `committed`, so nothing downstream can tell. The `kin_transaction_stage` description has warned about this since the shape shipped; a description is instruction. `reject_truncated_bodies` is the construction, and it runs on `kin_transaction_stage` and on `kin_mutate`. There is a third door: `kin_transaction_commit` also accepts operations inline, and that path does not run the guard yet. It is filed as a follow-up rather than widened into this PR.

It matches a body whose trailing text is the marker, never one that merely contains it. Kin's own source carries the literal token (`kin-cli/src/commands/trace.rs:874` writes it), so a `contains` check would refuse to edit exactly the entities that implement clipping. There is a test for each half.

While enforcing it, one sentence in that description turned out to be wrong and is corrected here: `get_entity_source` is not capped at 40/2400. It calls `read_entity_source_exact` with a 1,000,000-character budget (`handlers/entities.rs:424`) and applies no line or character cap of its own. Sending an agent back to a tool that caps would have been a loop with no exit.

**A commit can record what it did.** `kin-daemon`'s `commit_message` built the subject from the transaction id alone, so every MCP change in history reads `MCP transaction <uuid>`, which names the call and not the work. An optional `message` on the commit call now becomes the subject, with the transaction line kept underneath so an operator can still tie a change back to its call; `kin_mutate` forwards its `summary` there, trimmed, with an all-whitespace value read as none.

Explicit path only. There is no fallback deriving a message from the staged operations' own `description` fields, tempting as it is: that would rewrite what every commit already in flight records, and the point of this argument is to add a way to say something, not to change what silence means. A caller that sends nothing gets the old line byte for byte, and the existing assertion that says so stays green as the control.

**Every `kin_mutate` refusal is a tool error.** The handler answered a missing `operations` argument with `is_error` and then propagated parse and validate failures as `Err(McpError::InvalidParams)`, a protocol fault, so which shape a caller got depended on which mistake it made. The caller is a model that just wrote the argument, and a one-shot is only worth having if a rejected attempt is retryable from what came back.

**The built-in system prompt changed for every caller, and now follows the belt.** `run.rs`'s default prompt used to tell every model to change code with `edit_file`; this PR rewrote it for every `kin agent run`, pure-Kin or not. As first written, under `--tool-profile agent-query` it named `kin_mutate`, which that profile does not serve. The paragraph about changing code is now chosen from the belt the run actually built (`system_prompt_for`): `kin_mutate` on the belt gets the entity paragraph, file tools without it get the file paragraph the prompt had before, and a belt with neither is told the run is read-only. A test covers each shape. `--system <file>` still replaces the whole prompt.

## The one-shot never reached the exact commit, and the acceptance check is how we know

This is the defect worth reading, because everything else in this PR was green while it was true. The belt was green, the message plumbing was green, every crate test was green, and `kin_mutate` could not commit anything against a real daemon. Here is the daemon's own answer:

```
Cannot commit transaction f4a2179e-4102-4a07-9ee3-d21e57f78eb3: 1 staged operation(s) were
refused (source_body_requires_daemon_commit):
  - operation #0 ('update'): a payload-less source update (target plus body) for target
    'mutable' requires the daemon commit path, which carries the tree transition into
    repository authority and projects the working copy to match; the in-process commit path
    has no projection and would report success while discarding it
```

The mechanism runs through three files. In a daemon-backed deployment the MCP server forwards every tool call it does not answer itself, and until this PR the only name it answered locally was `kin_tool_search`. The daemon runs what arrives through `handlers::handle_tool_call` with `SessionAuthorityMode::OfflineFallback`, because inside the daemon its own registry IS the authority. So a forwarded `kin_mutate` reached `handle_mutate` with `uses_daemon()` FALSE, took the in-process branch, and was correctly refused by a commit path that has no projection. Meanwhile `api.rs` routes exactly one name to `mcp_commit::commit_exact_transaction`, and that name is `kin_transaction_commit`.

Which means `handle_mutate`'s daemon branch, the whole reason the one-shot was written that way, was dead code. Nothing ever called it with `uses_daemon()` true.

The fix stays inside kin-mcp. That branch is now `mutate_through_daemon`, a function with no graph store of its own, and `server.rs` answers `kin_mutate` itself the way it already answers `kin_tool_search`. What crosses to the daemon is a begin and a commit under the names the daemon matches, so the commit hits the exact-commit route, which is the same path `edit_lands` already proves works. The operations ride inline on the commit, which `kin_transaction_commit` has always accepted, and the commit carries a `transaction_id`, which is what the daemon's coordination preflight keys on.

Adding `kin_mutate` to `api.rs`'s routing instead was considered and rejected: it would need a shim to invent a transaction, and it would skip that coordination preflight, because there would be no transaction id to key on.

`a_mutation_is_expanded_locally_on_the_daemon_route` guards it, and it works because there is no daemon under the test. A locally expanded mutate refuses for its own missing `session_id` before it reaches the wire, naming `kin_session_start`, and a forwarded call cannot produce that answer under any daemon. The discriminating half names a session and requires the call to get past that refusal and still fail, which rules out an expansion that simply refuses everything. It deliberately does not assert what the wire then says, because whether a daemon answers depends on whether a store sits above the checkout, which is a fact about the machine and not about this route.

## Two defects fixed that the snapshot introduced

**`--tool-profile pure-kin` would have reconfigured the wrong layer.** The snapshot switched the belt on that value. But `kin agent run` forwards `--tool-profile` verbatim to `kin mcp start` (`kin-cli/src/commands/agent.rs:370`), whose accepted tokens are `agent-default`, `agent-query`, `agent-search`, `full` and `benchmark`; an unrecognised value falls back to the curated default. One word would have asked for a belt here and quietly changed the served surface there. The flag's own help already says it names the server's surface. The branch is gone, `KIN_AGENT_PURE_KIN` is the whole switch, and the run record now carries a `belt` field beside `tool_profile` so a reader of one run's provenance can tell the two settings apart.

**The one write tool on a pure-Kin belt could not have committed at all.** `handle_mutate` resolved a missing `session_id` out of the MCP server's own in-process registry, and started a session there when it was empty. In daemon mode that registry is not the authority: `handle_session_start` forwards to the daemon and returns before registering anything locally, so the fallback invents an id the daemon has never heard of and `kin_transaction_begin`, which requires one, refuses a call later with a message about a session rather than about the argument that was wrong.

The model could not have supplied it either. `kin_session_start` is harness-owned, so it is not on the belt and the model never learns the session the run opened. On a belt whose only write tool is `kin_mutate`, that is a belt that cannot write.

Fixed on both sides, because either alone leaves a way in. The kin-agent harness fills in its own session id on a model-issued `kin_mutate`, exactly as it already does on its own staging call, and leaves a session the model named itself alone. And `handle_mutate` now refuses a daemon-mode call that names no session, naming both the argument and the tool that produces it. The offline path keeps its fallback, since there the local registry is the authority, and that is the control in the test.

**`KIN_AGENT_PURE_KIN=off` locked the belt.** The snapshot read the switch as on for anything except `0` or `false`. kin-core's env registry, where every `KIN_*` variable is supposed to be declared, reads a boolean as `1/true/yes/on` against `0/false/no/off`, so the word an operator most naturally writes to turn the lock off turned it on, and the registry's own startup validation would have told them otherwise. The variable was also never registered, which is how hosted CI found it: `every_kin_env_read_in_workspace_is_registered` failed on the first push. It is registered now as a `Kind::Bool` defaulting to false, and the belt mirrors the registry's exact true set with a comment naming the registry as the authority, rather than taking a kin-core dependency in a crate that today depends on nothing heavier than reqwest.

**Entity targets were being published as file paths.** The snapshot pushed `kin_mutate` targets into `counters.edits`, which is published as `files_changed` and which `loop_integration.rs:732` asserts is a path. An entity target is a UUID or a name and resolves against the graph. They are separated here, with `entities_changed` as the new list, so a pure-Kin run records what it changed rather than reporting an empty `files_changed`. Only a call that came back clean counts, since a run that listed its refusals as edits would overstate itself in the exact place a reader checks.

## One reported-red test, which is green

`handle_mutate_atomic_validation_and_execution` was handed over as failing on an unwrap of `InvalidParams`. It is green at the snapshot's own bytes, run alone before any edit here: `1 passed; 0 failed`. The fixture is well formed, so the likeliest reading is that it was fixed before the snapshot was taken. Recorded rather than quietly dropped, because a red test nobody can reproduce is worth a sentence. The `unwrap` that turned a refusal into a panic was real and is gone either way, and the same mistake now reports as a structured error.

## Verification

Hosted CI checks none of the local evidence below, so each claim is the command and its actual output.

Which tree each piece of evidence graded, because main moved under this PR and the head moved twice. #1726 landed at 20:34Z, nine minutes before this PR's second commit, and it changes `envelope.rs` and `negative.rs`, which the `kin_mutate` answer passes through. Hosted CI on this PR grades `refs/pull/1738/merge`, whose parents are that main and this branch, so CI already sees the combined tree. The first ten falsification arms, precheck and parity ran on `4e2e8cf83`, the branch head before main was merged in. Main was then merged into the branch as `e5f39a551` (parents `4e2e8cf83` and main at `5503ce6eb`, signed off, no conflict; none of main's intervening commits changes a line of the daemon commit code this PR touches), and on that merged tree kin-mcp lib passed 1024 tests, kin-agent 48 plus 25, kin-daemon's `mcp_commit` 80 and kin-core's env registry 34, so the generated `docs/env-vars.md` still matches the registry. The review fixes are `9e8fa7ec3`, the head this PR lands: on it kin-mcp lib passes 1027 and kin-agent 49 plus 25, precheck reads 22 of 22, four more falsification arms went red, and the acceptance suite passed 6 of 6 on that head's own binaries. Every block below says the head it ran on.

### Falsification: every new test, broken on purpose

Fourteen arms, one per behaviour: ten at `4e2e8cf83`, and four at `9e8fa7ec3` for the tests the review fixes added (`beltfinish-falsify2.py`, the same driver with its own arm list). Each breaks the BEHAVIOUR in the source, never the assertion, runs only the test written for it through the fleet's build slot, and is reverted from the commit before the next. Driver: `beltfinish-falsify.py`, which refuses to start on a dirty tree (the revert is `git checkout -- <file>`, which restores the commit) and refuses any arm whose target text does not occur exactly once, since a mutation that did not land proves nothing.

```
1  pure-kin belt keeps its file tools             kin-agent  rc=101
   if file_tools { .. }  ->  if true { .. }
   panicked at crates/kin-agent/src/tests.rs:309:5: assertion `left == right` failed
   left: 3  right: 1

2  the switch goes back to reading "off" as on    kin-agent  rc=101
   PURE_KIN_TRUE.contains(..)  ->  value != "0" && value != "false"
   the_pure_kin_switch_reads_booleans_the_way_the_env_registry_does ... FAILED

3  the truncation guard never fires               kin-mcp    rc=101
   if !body.trim_end().ends_with(MARKER)  ->  if true
   both_write_surfaces_refuse_a_body_kin_cut_short ... FAILED

4  the truncation guard matches anywhere          kin-mcp    rc=101
   ends_with(MARKER)  ->  contains(MARKER)
   panicked at crates/kin-mcp/src/handlers/mod.rs:6440:9
   (a_body_that_merely_mentions_the_marker_passes_the_truncation_guard)

5  a malformed operations array is waved through  kin-mcp    rc=101
   parse_staged_operations(..).map_err(..)?  ->  .unwrap_or_default()
   panicked at crates/kin-mcp/src/handlers/mod.rs:6168:10
   (handle_mutate_answers_a_malformed_operations_array_with_a_tool_error)

6  a blank summary becomes a blank subject line   kin-mcp    rc=101
   .trim(); (!summary.is_empty()).then(..)  ->  Some(summary.to_string())
   panicked at crates/kin-mcp/src/handlers/mod.rs:6393:9: assertion `left == right` failed
   left: Some("  Return two from value \n")  right: Some("Return two from value")

7  a daemon-mode mutate invents a session again   kin-mcp    rc=101
   .ok_or_else(refusal)  ->  .or(Some("invented-locally")).ok_or_else(refusal)
   panicked at crates/kin-mcp/src/handlers/mod.rs:6347:9
   (a_daemon_mode_mutate_without_a_session_is_refused_by_name)

8  the server forwards kin_mutate again           kin-mcp    rc=101
   if call_params.name == "kin_mutate"  ->  == "kin_mutate_never_matches"
   the forwarded call answered: "kin-mcp cannot answer 'kin_mutate': ... is not a Kin
   repository", which can never name kin_session_start, so the local-expansion assertion fails
   (a_mutation_is_expanded_locally_on_the_daemon_route ... FAILED)

9  the harness stops filling in the session     kin-agent  rc=101
   (Some(session), "kin_mutate") = (session, tool)  ->  (session, "never-matches")
   tests::the_harness_fills_in_the_session_a_mutate_needs_and_leaves_every_other_call_alone
   ... FAILED; and the compiler's own witness that the break was in place:
   warning: unused variable: `tool` --> crates/kin-agent/src/run.rs:238:55

10 the daemon ignores the message its caller named  kin-daemon rc=101
   match requested { Some(m) => .., None => .. }  ->  return transaction_line;
   test result: FAILED. 0 passed; 1 failed; 1710 filtered out
   (the one test run was mcp_commit::tests::a_commit_records_the_message_its_caller_named)

11 a refused commit is returned as if it landed   kin-mcp    rc=101
   Ok(Some(mut value)) if value.is_error != Some(true)  ->  Ok(Some(mut value))
   a_refused_mutate_commit_is_aborted_and_a_landed_one_is_not ... FAILED; and the
   compiler's own witness that the break was in place:
   warning: unreachable pattern --> crates/kin-mcp/src/handlers/sessions.rs:1658:9

12 a commit the daemon never answered returns      kin-mcp    rc=101
   before the abort
   Ok(None) => daemon_required_unavailable(..)  ->  Ok(None) => return Ok(daemon_required_unavailable(..))
   panicked at crates/kin-mcp/src/handlers/mod.rs:6515:9: assertion `left == right` failed
   left: Some("kin_transaction_commit")  right: Some("kin_transaction_abort")

13 a failed abort is swallowed again               kin-mcp    rc=101
   note_open_transaction returns at its top
   panicked at crates/kin-mcp/src/handlers/mod.rs:6541:9 with the whole answer reading
   "timed out": no transaction named, nothing said about what was left open

14 the prompt names kin_mutate whatever the belt   kin-agent  rc=101
   carries
   if belt.has_kin_tool("kin_mutate")  ->  if true || belt.has_kin_tool("kin_mutate")
   the_system_prompt_names_only_the_write_tools_the_belt_carries ... FAILED, printing
   the file-tools belt's prompt with the kin_mutate paragraph in it
```

### Gates

`bin/kin-precheck kin --repo-dir kin=<lane worktree>`, which reads its gate list out of this tree's own `ci.yml` rather than a remembered copy, on the head this PR lands:

```
PASS  fmt                                  PASS  guard:test-install-checksum.py
PASS  clippy (-D warnings, 45 allows)      PASS  guard:test-private-repo-coupling.py
PASS  guard:check_private_repo_coupling.sh PASS  guard:test-release-tag-evaluate-dispatch.py
PASS  guard:check_runtime_boundaries.sh    PASS  guard:test-release-workflow-authority.py
PASS  guard:test-release-policy-gate.sh    PASS  guard:test-select-release-candidate.py
PASS  guard:test-windows-authority-legs.sh PASS  guard:verify-hydration-semantics.py
PASS  guard:zero_file_search_guard.sh      PASS  guard:verify-zero-file-search.py
PASS  guard:check-kin-vfs-compat.mjs       PASS  node-test (16 files)
PASS  guard:check-rc-build-drift.mjs       PASS  guard:check-quarantine.py
PASS  guard:check-required-contexts.py
PASS  guard:test-assertion-reachability.py
PASS  guard:test-guard-duplicate-release-run.py
PASS  guard:test-homebrew-release-gate.py
kin-precheck: 22 gates ran, 22 passed, 0 failed, on kin 9e8fa7ec3a843be873ca178d7d86259cc1f6b073
```

The same command on the earlier head `4e2e8cf83` read 21 of 22, and that one failure was a timeout rather than a verdict on the diff: `test-release-workflow-authority.py` gives a `node -` validator fixture ten seconds, and on a host at load 16 to 20 with six builders it took longer, `subprocess.TimeoutExpired: Command '['node', '-']' timed out after 10 seconds`. The fixture it runs checks that a home config with `KIN_MCP_TOOL_PROFILE=full` is rejected, which this PR does not touch, and the gate passes on this head.

`bin/kin-parity` ran on `4e2e8cf83` and was not rerun for the review fixes. Those fixes touch `kin_mutate`'s daemon path, the built-in prompt and two tests, and parity's three suites drive the CLI and the daemon over real stores without ever calling `kin_mutate`; what does exercise that path end to end is the acceptance suite below, which reran on this head's own binaries. `--profile dev`, so non-citable by design; the binary reported the checkout's head, `4e2e8cf83433`.

```
kin-parity: magic       PASS                  304.5s  27 pass / 0 fail / 0 unreadable
kin-parity: brownfield  PASS-WITH-ALLOWANCES  558.1s  12 pass / 1 fail / 0 unreadable
kin-parity: firstrun    PASS-WITH-ALLOWANCES  109.9s  10 pass / 1 fail / 1 unreadable
kin-parity: FINAL PASS-WITH-ALLOWANCES in 992.1s  (NON-CITABLE)
            allowances: firstrun/9 FIR-2580, firstrun/3 FIR-2501, brownfield/4 FIR-2464
```

All three are tracked defects named in kin-parity's checked-in allowance list, and none touches this change: a reference walk that comes back truncated (FIR-2464), `kin daemon stop --all` leaving a run's daemons alive (FIR-2580), and the doctor's projection row that cannot go green under Kin's own shell hook (FIR-2501).

### Suites

The two crates the review fixes touched ran on `9e8fa7ec3`, the head this PR lands; the two they did not touch ran on the merged tree `e5f39a551`, the commit before it. The log of each round is kept.

```
cargo test -p kin-mcp --lib      1027 passed; 0 failed; 0 ignored          on 9e8fa7ec3
cargo test -p kin-agent          49 passed (lib); 25 passed, 1 ignored     on 9e8fa7ec3
                                 (loop_integration)
cargo test -p kin-daemon --lib mcp_commit::
                                 80 passed; 0 failed; 0 ignored            on e5f39a551
cargo test -p kin-core --lib env_registry
                                 34 passed; 0 failed                       on e5f39a551
python3 scripts/acceptance/agent_write_publish_repro.py --self-test
                                 SELF-TEST PASS (28 expectations)          on 9e8fa7ec3
python3 scripts/acceptance/mcp_surface_contract.py --self-test
                                 every grader passed its correct case      on 9e8fa7ec3
                                 and failed each break (rc=0)
```

kin-mcp went from 1008 at this branch's base to 1027, nineteen more, of which ten are this PR's and nine came in with main. kin-agent lib went from 45 to 49, and the acceptance self-test from 20 expectations to 28.

### The acceptance suite, against this branch's own binaries

It grades main rather than a pull request, so this is the only place it is read before the landing, and a check that failed on main would stop a release cut. First run, before the routing fix, five of six:

```
CHECK pure_kin_mutate_lands  FAIL  the file on disk does not carry the mutation;
                                   get_entity_source still answers with the old body;
                                   kin_mutate came back an error; ...
suite rc=1
```

After it:

```
CHECK edit_lands                     PASS
CHECK refused_edit_is_clean          PASS
CHECK create_lands                   PASS
CHECK refused_create_is_clean        PASS
CHECK pure_kin_mutate_lands          PASS  a Kin-only belt committed by naming the entity, under a
                                           session the harness supplied, and history carries the
                                           agent's own sentence
CHECK edit_survives_a_daemon_restart PASS
suite rc=0
```

And on the merged tree's own binaries, `e5f39a551`, `kin 0.7.14 (e5f39a551... 2026-09-11T21:19:19Z)`, sha256 `5c3d9f0a238f65730d1e0bd894f304b6527c17e454692ad6935a22008fd6de8e`, with main's newer commits in:

```
CHECK edit_lands                     PASS
CHECK refused_edit_is_clean          PASS
CHECK create_lands                   PASS
CHECK refused_create_is_clean        PASS
CHECK pure_kin_mutate_lands          PASS
CHECK edit_survives_a_daemon_restart PASS
```

And on this PR's head, `9e8fa7ec3`, after the review fixes. The binary was rebuilt after cleaning kin-mcp and kin-cli, so its stamp is the committed head rather than a dirty tree: `kin 0.7.14 (9e8fa7ec3a843be873ca178d7d86259cc1f6b073 chore/lane-beltfinish-kin 2026-09-11T21:51:06Z)`, sha256 `31bb2300ae43d871858a8e857f08ab4525e961e835b83e8e217441872f1bede9`, with kin-daemon at `d50a7ba9f8914983f756ac0feedfe45bf7a5666f79f2ffeb28857c5de8a64103`:

```
CHECK edit_lands                     PASS
CHECK refused_edit_is_clean          PASS
CHECK create_lands                   PASS
CHECK refused_create_is_clean        PASS
CHECK pure_kin_mutate_lands          PASS  a Kin-only belt committed by naming the entity, under a
                                           session the harness supplied, and history carries the
                                           agent's own sentence
CHECK edit_survives_a_daemon_restart PASS  begin was refused for the gone session and the harness
                                           opened a new one
suite rc=0
```

### The proof: a real local model commits through Kin by naming the entity

_Running. qwen/qwen3.8-27b on the product's pure-Kin belt, three runs against this head's own binaries, each in a fresh store, under a keeper that holds the fleet's gpu lock and loads the model with a 32,768-token context and one parallel slot, read back from `lms ps` before any run starts. Each run's store, binary hashes, tool calls, tokens, wall time and recorded change message go here, with the loaded shape and the swap high-water mark, before this PR is enqueued._

## An acceptance check, because a crate test could not have caught this

The session defect is the class `agent_write_publish_repro.py` was written for, restated. Its own header says why the first one got through: kin-agent's tests run against a scripted MCP server that projects whatever it is sent, and kin-mcp's run in-process where its own registry IS the authority, so both halves were green while the composition was broken. Exactly the same is true of the session fallback, and a crate test cannot see it, because it needs a real daemon that owns real sessions.

So the suite gains a sixth check, `pure_kin_mutate_lands`. It runs the same binary a second time under `KIN_AGENT_PURE_KIN`, because the belt is read per process, and requires one `kin_mutate` naming the ENTITY to commit: the file and `get_entity_source` carry the new body, durability reads `recorded`, the run records `entities_changed` and no file, `kin log` carries the agent's own summary rather than the bare transaction line, and the call went out under a session the harness supplied. Its fixture gets a file of its own, since the checks run in order against one repository and an entity a neighbour already moved cannot tell one failure from another.

Seven self-test arms come with it, and the one that matters drives a `kin_mutate` with no `session_id` and requires FAIL, so the assertion this check exists for cannot quietly become a note. `--self-test` is green at 28 expectations, up from 20.

Note what this does and does not buy. kin's `Product Acceptance` job carries `if: ${{ github.event_name != 'pull_request' }}`, so this check does NOT grade this pull request; it grades main after it lands, which is the shape the fleet chose. The crate tests are what CI can see today, and they are on both halves.

## Not in this PR

Two follow-ups filed from review, not widened into this PR: `kin_transaction_commit`'s inline-operations path does not run the truncated-body guard (the third door above), and `Belt::route`'s refusal of `edit_file` on a pure-Kin belt names a bare `kin_mutate`, which is not a callable name on a multi-server run, where the tools carry a per-repository prefix.

The parser half of the snapshot, which grows a Rust entity's span to cover its doc comment and attributes. It changes `doc_summary` on every attributed Rust entity, which is a re-enrichment blast radius rather than a span tweak, and it follows as its own PR with that radius measured.


troyjr4103:
## The proof this PR was waiting on

This landed as `10cd9d4c4e0397785196750d2f8a66c26cba2490` before the proof could run, because the fleet's gpu was held
elsewhere until after the merge. Here it is against the head this PR merged, `9e8fa7ec3`,
so the belt claim in the body is not left standing on crate tests alone.

### The proof: a real local model commits through Kin by naming the entity

qwen/qwen3.8-27b on the product's pure-Kin belt (`KIN_AGENT_PURE_KIN=1`), three runs against the binaries built from `9e8fa7ec3a843be873ca178d7d86259cc1f6b073`, each in a fresh Git-backed store made from the four demo sources with `kin init`, under a keeper holding the fleet's gpu lock. Runs started: 1, 2, 3; runs completed: 1, 2, 3.


Each run was given the same task, in one paragraph:

> A job that has been dead-lettered must never be retried again. Change the method can_retry on the class Job in models.py so that it returns False when the job status is JobStatus.DEAD_LETTERED, and otherwise keeps the existing rule that a job may retry while its attempts are below its retry policy max_attempts. Change nothing else.


No run was told which file to open, and none could have opened one: the belt has no `edit_file` and no `write_file`. Each found the entity through Kin, read its exact source, and sent the whole new body back through one `kin_mutate`. This is what the source held afterwards, from run 1's store:

```python
    def can_retry(self) -> bool:
        if self.status == JobStatus.DEAD_LETTERED:
            return False
        return self.attempts < self.retry_policy.max_attempts
```


What the model was actually loaded with, because it is not what was asked for. The keeper loaded with `--context-length 32768 --parallel 1 --gpu max`; LM Studio applied the parallel count and reported the context as the model's full window on every surface it offers:

```
lms ps --json:     contextLength=262144 maxContextLength=262144 parallel=1
api/v0/models:     loaded_context_length=262144 max_context_length=262144
VERDICT: parallel 1 as asked; no source reports the requested context 32768 (reported: 262144)
```

So these runs ran at the model's full 262,144-token window with one parallel slot, and the agent's own budget (`--context-tokens 32768`) is what bounded each prompt. `lms load --estimate-only` returns the identical estimate at both lengths, so it cannot distinguish them either. Swap on the box never rose above the 13048.88M it already sat at when the keeper started, and the model was unloaded by name with its absence confirmed from `lms ps` at 2026-09-12T01:30:19Z. The box was busy with other lanes' builds throughout, by the captain's ruling, so the wall times below are not clean timings and are not offered as any: what these runs prove is that the change lands.

```
run 1
  store        /Users/troyfortinjr/kin-scratch-beltfinish/run1
  kin          kin 0.7.14 (9e8fa7ec3a843be873ca178d7d86259cc1f6b073 chore/lane-beltfinish-kin 2026-09-11T21:51:06Z)
  kin sha256   31bb2300ae43d871858a8e857f08ab4525e961e835b83e8e217441872f1bede9
  daemon sha256 d50a7ba9f8914983f756ac0feedfe45bf7a5666f79f2ffeb28857c5de8a64103
  belt         pure-kin; tools in the init record 15; file tools among them: none
  agent exit   rc=0, stop_reason=final_answer, exit_code=0
  tool calls   8 (kin 8, local 0, refused 0)
  tokens       input 78853, output 1466; turns 7
  wall         93 s by the clock around the run; agent duration_ms 92581
  changed      entities ['bdc9f5ef-6a35-4e2d-9836-784a0694a9a0']; files []
  kin_mutate   1 call(s) in the trace
               is_error=False wall_ms=3545 summary='Job.can_retry now refuses to retry dead-lettered jobs'
               verb='update' target='bdc9f5ef-6a35-4e2d-9836-784a0694a9a0' body_chars=166
  kin log      newest subject: 'Job.can_retry now refuses to retry dead-lettered jobs'
  kin log      newest message whole: 'Job.can_retry now refuses to retry dead-lettered jobs\n\nMCP transaction 1ebd8113-b7ff-4c13-ad6a-b5ea2ef116c4'

run 2
  store        /Users/troyfortinjr/kin-scratch-beltfinish/run2
  kin          kin 0.7.14 (9e8fa7ec3a843be873ca178d7d86259cc1f6b073 chore/lane-beltfinish-kin 2026-09-11T21:51:06Z)
  kin sha256   31bb2300ae43d871858a8e857f08ab4525e961e835b83e8e217441872f1bede9
  daemon sha256 d50a7ba9f8914983f756ac0feedfe45bf7a5666f79f2ffeb28857c5de8a64103
  belt         pure-kin; tools in the init record 15; file tools among them: none
  agent exit   rc=0, stop_reason=final_answer, exit_code=0
  tool calls   8 (kin 8, local 0, refused 0)
  tokens       input 67576, output 1176; turns 6
  wall         70 s by the clock around the run; agent duration_ms 70007
  changed      entities ['e84b4872-97e5-46c9-98d9-dc5b30adc3bd']; files []
  kin_mutate   1 call(s) in the trace
               is_error=False wall_ms=1622 summary='Job.can_retry now refuses to retry dead-lettered jobs'
               verb='update' target='e84b4872-97e5-46c9-98d9-dc5b30adc3bd' body_chars=166
  kin log      newest subject: 'Job.can_retry now refuses to retry dead-lettered jobs'
  kin log      newest message whole: 'Job.can_retry now refuses to retry dead-lettered jobs\n\nMCP transaction dd14fb5c-b167-4864-a0f3-c0abb9725242'

run 3
  store        /Users/troyfortinjr/kin-scratch-beltfinish/run3
  kin          kin 0.7.14 (9e8fa7ec3a843be873ca178d7d86259cc1f6b073 chore/lane-beltfinish-kin 2026-09-11T21:51:06Z)
  kin sha256   31bb2300ae43d871858a8e857f08ab4525e961e835b83e8e217441872f1bede9
  daemon sha256 d50a7ba9f8914983f756ac0feedfe45bf7a5666f79f2ffeb28857c5de8a64103
  belt         pure-kin; tools in the init record 15; file tools among them: none
  agent exit   rc=0, stop_reason=final_answer, exit_code=0
  tool calls   8 (kin 8, local 0, refused 0)
  tokens       input 67520, output 1465; turns 6
  wall         77 s by the clock around the run; agent duration_ms 77473
  changed      entities ['21384512-0303-4f3a-aa6f-1714652d1723']; files []
  kin_mutate   1 call(s) in the trace
               is_error=False wall_ms=1295 summary='Job.can_retry now refuses to retry dead-lettered jobs'
               verb='update' target='21384512-0303-4f3a-aa6f-1714652d1723' body_chars=166
  kin log      newest subject: 'Job.can_retry now refuses to retry dead-lettered jobs'
  kin log      newest message whole: 'Job.can_retry now refuses to retry dead-lettered jobs\n\nMCP transaction 82fb3373-f7bd-42f3-bae2-611c272f2f24'

```

Each line is read from that run's own records: the store path, binary hashes and wall clock from the run's section of the proof log; exit, stop reason, tool calls, tokens and `entities_changed` from the agent's `result.json`; the belt and its tool list from the transcript's init record, which is `belt.names()` for that run; and the recorded change message from `kin log -n 3 --json` in the run's store after the run. `files_changed` is empty on every run by construction: a pure-Kin belt has no file tool, so a change either went through `kin_mutate` or did not happen.

PR #1747, opened 2026-09-11
Bring kin-blobs, kin-model, kin-search and kin-vector into the workspace with their history

troyjr4103:
Brings kin-blobs, kin-model, kin-search and kin-vector into `crates/` as workspace members with
their history, and turns kin's four exact registry pins into path dependencies that KEEP their
`version` and `registry` fields. Day one changes nothing about what is published or where. Nothing
was deleted from any registry, no tag moved, and no repository was archived.

First of three. kin-infer, kin-lsp and kin-vfs-core follow, then kin-db.

## What was imported

| package | source repository | frozen sha | version kept |
|---|---|---|---|
| kin-blobs | firelock-ai/kin-blobs | `0f58edf295dbc329e5d41a501e466855b8f3e833` | 0.1.5 |
| kin-model | firelock-ai/kin-model | `297d198afcd795e64c6fd7045477bad70dc055a3` | 0.7.28 |
| kin-search | firelock-ai/kin-search | `94e99c6efc6992fc9e35e88e5973b2d40b4336e0` | 0.1.9 |
| kin-vector | firelock-ai/kin-vector | `358b57932e0e455848bfe222f9ff7c8eb5efb0a5` | 0.2.1 |

Each source repository's `main` was frozen before its head was read.

`git-filter-repo` moved each repository's files under `crates/<name>/` across its whole history, so
every historical commit carries them there and `git log` and `git blame` reach them with no
`--follow`. A squash would have dropped all of it. Imported tags are namespaced
`imported/<repo>/` so they cannot collide with kin's release tags or trigger publication, and they
are not pushed.

History verified both ways, since `git log -- <path>` simplifies and `rev-list --count` does not:

```
kin-blobs   full 57 = 57    simplified 54 = 54    tree IDENTICAL (23 paths)
kin-model   full 139 = 139  simplified 131 = 131  tree IDENTICAL (73 paths)
kin-search  full 86 = 86    simplified 82 = 82    tree IDENTICAL (25 paths)
kin-vector  full 90 = 90    simplified 88 = 88    tree IDENTICAL (28 paths)
```

## Sign-off trailers written during import

The DCO job grades every commit a pull request carries. 29 of these 372 carried no `Signed-off-by`,
all authored by Troy Fortin, all July 2026 merges and CI chores from before those repositories
adopted `git commit -s`. Each was given its OWN author's identity as its sign-off through
filter-repo's commit callback, and nothing else about any commit changed: author, date, subject and
body are untouched, and filter-repo already rewrites every sha because the paths move. The full
list, per commit with its original and imported sha, is in the migration record under "Sign-off
trailers written during import".

## The pin change

```
-kin-model  = { version = "=0.7.28", registry = "kin" }
+kin-model  = { path = "crates/kin-model",  version = "=0.7.28", registry = "kin" }
-kin-blobs  = { version = "=0.1.5",  registry = "kin" }
+kin-blobs  = { path = "crates/kin-blobs",  version = "=0.1.5",  registry = "kin" }
-kin-search = { version = "=0.1.9",  registry = "kin" }
+kin-search = { path = "crates/kin-search", version = "=0.1.9",  registry = "kin" }
+kin-vector = { path = "crates/kin-vector", version = "=0.2.1",  registry = "kin" }
```

kin-vector had no root entry at all; it arrived transitively through kin-db and kin-model, and a
member can only say `{ workspace = true }` for a name the table carries.

## The transitional [patch.kin] table

Until kin-db is imported it still resolves from the registry, and the published kin-db depends on
the published kin-model, kin-search and kin-vector. Without the table the graph carries two of each,
distinct types with one name. A patch whose version does not satisfy the requirement is ignored
SILENTLY and the build succeeds against the registry copy, so here are the requirements, read from
the registry index and identical across kin-db 0.7.112, 0.7.113 and 0.7.114:

```
kin-model  "=0.7.28"   crates/kin-model  0.7.28
kin-search "0.1.9"     crates/kin-search 0.1.9
kin-vector "=0.2.1"    crates/kin-vector 0.2.1
kin-infer  "0.2.44"    (still from the registry until the next pull request)
```

And the proof that the table is doing something, from `cargo tree -p kin-db --depth 1`:

```
kin-db v0.7.112 (registry `kin`)
├── kin-infer v0.2.44 (registry `kin`)
├── kin-model v0.7.28 (.../crates/kin-model)
├── kin-search v0.1.9 (.../crates/kin-search)
├── kin-vector v0.2.1 (.../crates/kin-vector)
```

`cargo tree -d` lists no Kin library at all. The table is removed by the pull request that imports
kin-db.

## Both locks, and nothing else

`Cargo.lock` moves by 6 insertions and 8 deletions, `fuzz/Cargo.lock` by 6 deletions. Every line is
a `source` or `checksum` leaving a package that is now a member, plus the dev-dependencies a member
records and a registry package does not. No third-party version moved; `tinyvec` stayed at 1.11.0.
`fuzz/` is a detached workspace with its own lock that reaches kin-model through kin-parser, and a
re-lock of the root alone would have left it behind.

## The new guard, and its falsification

`crates/kin-core/tests/kin_library_resolution.rs` fails when a Kin library resolves from the
registry, or when one appears twice in a lock. Its subject is the LOCK, which is the resolution of
record, rather than the root manifest, which is only a request, and it reads every lock the tree
tracks, found by walking rather than listed. Its transitional list carries kin-db, kin-infer,
kin-lsp and kin-vfs-core with the pull request that removes each, and an entry no lock needs is
reported too.

Falsified end to end by taking `path` off kin-blobs in the root manifest and re-locking:

```
### control: the guard passes on the tree as committed
test result: ok. 6 passed; 0 failed
### poison: kin-blobs loses its path AND its [patch.kin] row
### the poison must have LANDED: a kin-blobs with a registry source in the lock
[('0.1.5', None), ('0.1.5', 'sparse+https://kinlab.ai/registry/cargo/')]
LANDED
### the guard must now be RED and must name kin-blobs
guard rc=101
  Cargo.lock: kin-blobs 0.1.5 resolves from sparse+https://kinlab.ai/registry/cargo/. It is a Kin
  library and must resolve from crates/kin-blobs. ...
  Cargo.lock: kin-blobs appears 2 times (0.1.5 from this workspace, 0.1.5 from
  sparse+https://kinlab.ai/registry/cargo/). ...
  test result: FAILED. 5 passed; 1 failed
FALSIFICATION OK: the poison landed, the guard went red, and it named kin-blobs
restored from HEAD
```

The poison has to defeat `[patch.kin]` as well as the root entry. A first attempt removed only the
path, the patch redirected the registry request straight back to `crates/`, the lock stayed
path-sourced, and the guard passed, correctly, on a poison that never landed. The script now asserts
the poison reached the lock before it reads any verdict. The script is
`.kin-coord/bin/monorepo-falsify-resolution.sh` in the umbrella.

Five controls sit beside the pure decision function, including a clean lock that must report
nothing and a registry-sourced serde that must stay unreported, since the rule is not "nothing
resolves from a registry".

## CI

`library-packages` is a new job on every event: each Kin library tested on its own minimal features
and packaged. A workspace unifies features, so a library can look healthy in the full build while
the consumer who depends on it alone never enables what it needs, and `feature-tests` runs those
legs only on a push to main. Packaging uses `--no-verify` because `cargo package` rewrites a path
dependency to its registry version, so a library whose sibling is unpublished cannot be verified in
isolation; full verification against the registry belongs to publication, in dependency order.

Its crate list is derived from the root manifest (an entry with both a `path` and a `registry`), so
a library added later is covered on the commit that adds it, and it refuses a derivation shorter
than four rather than grading nothing and exiting 0. `feature-tests` gains the same minimal-feature
legs for the macOS half. The job name joins the census in
`scripts/test-release-workflow-authority.py`, which is what keeps a new job name from appearing
unreviewed.

The imported crates need no new test wiring. `fast-gate-tests` selects packages with
`scripts/changed-crate-scope.py` over `cargo metadata`, and run against this tree's metadata a
change under any of the four resolves to 20 or 21 of the 28 members, which its own
`most-of-workspace` rule widens to `--workspace`:

```
crates/kin-blobs/src/lib.rs  -> 21 of 28 packages selected -> rule=most-of-workspace -> --workspace
crates/kin-vector/src/lib.rs -> 21 of 28 packages selected -> rule=most-of-workspace -> --workspace
crates/kin-model/src/lib.rs  -> 20 of 28 packages selected -> rule=most-of-workspace -> --workspace
(control) no changed paths -> rule=no-paths -> --workspace
```

## Zero file-search

`verify-zero-file-search.py` now walks these crates, and reported 85 violations across 4 files before
this change: kin-vector `src/lib.rs`, kin-blobs `src/lib.rs`, and kin-search `src/lib.rs` and
`src/mapped.rs`. kin-model had none; it is pure types.

All four are PINNED, not exempted whole. The checker's own docstring says whole-file "is only
appropriate for files that are an IO boundary end to end", and these are not: kin-vector and
kin-search are search libraries, and a whole-file exemption would hide a future raw-file fallback in
their query paths, which is exactly what the Zero File-Search Authority Rule forbids. So each entry
names the functions that hold filesystem primitives (`allow_fn`), plus the exact expression for hits
outside any function (`allow_match`, here only `use` trees and one call site). Every other function
in these files, including any added later, stays graded.

What the checker enforces, measured rather than assumed, on four mutations of this tree's allowlist:

```
T1 whole-file entry on a file with zero hits (kin-model/src/lib.rs)   PASSES, rc=0
T2 allow_fn naming a scanned kin-search function with no hits          PASSES, rc=0
T3 allow_match on an expression that occurs nowhere                   FAILS,  "occurs 0 times in scanned code (want 1)"
T4 allow_fn 'acquire_write_lock' dropped from kin-vector              FAILS,  "Found 2 zero-file-search violations"
```

So every `allow_match` count is exact, including zero, and every pin here is load-bearing today:
drop any one and the guard goes red. The checker does NOT notice a pinned function that stops doing
IO, or a whole-file entry that is no longer needed (T1, T2). These pins were generated from the
checker's own report and its own function-body finder, so each names a function that holds at least
one hit on this tree; keeping them honest after that is review's job, not the checker's.

| file | allow_fn (the functions that do the IO) | allow_match |
|---|---|---|
| `crates/kin-blobs/src/lib.rs` | 14: `delete`, `ensure_directory_metadata`, `ensure_shard_dir`, `exists`, `gc`, `list_hashes`, `open`, `prune_empty_shards`, `quarantine`, `read_unverified`, `sync_dir`, `validate_dir_type`, `write`, `write_body` | `std::fs::{self, File}` |
| `crates/kin-search/src/lib.rs` | 12: `archive_corrupt_index`, `kinseg_sibling_files`, `load_persisted`, `load_segmented`, `open_with_persistence`, `persist_monolithic`, `retire_segmented_artifacts`, `sync_parent_dir`, `unique_tmp_path`, `version_of_manifest`, `write_and_promote`, `write_file_durably` | none |
| `crates/kin-search/src/mapped.rs` | 3: `open_archiving`, `read_manifest_gens`, `write_mapped_streaming` | the one `std::fs::File::open(path)` line in `open` |
| `crates/kin-vector/src/lib.rs` | 11: `acquire_write_lock`, `clear_recovery_candidate`, `fsync_and_rename`, `load_from_disk`, `load_proven_recovery_bytes`, `load_recovery_marker`, `open_container`, `promote_recovery_candidate`, `recover_from_tmp`, `sync_directory`, `write_bytes_with_fsync` | `std::{cmp::Reverse, fs, fs::File, fs::OpenOptions}`, `fs2::FileExt` |

`mapped.rs` has two functions named `open` and only one of them opens a file, so that one line is
pinned rather than the name; the checker's counted name form would have excused the other `open` too.
BM25 ranking, the mapped reader's lookups and HNSW search all stay graded. Each entry carries its
crate as owner and a staggered expiry.

The pins were attributed with the checker's OWN function-body finder, not a regex, so every count is
right by construction; the generator is kept in the umbrella at `.kin-coord/bin/monorepo-gen-pins.py`.

## Verification

Everything below ran through `.kin-coord/bin/heavy-slot.sh` against this branch.

```
bin/kin-precheck kin --repo-dir kin=<lane worktree>
  kin-precheck: repo=kin tree=.../worktrees/monorepo-kin sha=f51bce2e4734b2a3fb7a19959a78dd1201962258 branch=chore/lane-monorepo-kin
  kin-precheck: 22 gates ran, 22 passed, 0 failed, on kin f51bce2e4734b2a3fb7a19959a78dd1201962258

The pushed head, 7a5b74afc, differs from f51bce2e4 by scripts/zero-file-search-allowlist.json alone:
this lane's three commits were rebuilt to fold the pinned entries into the reconcile commit, the
imported commits untouched (sha256 over the sorted rev-list of the last import merge, before and
after: 8013aeee666fc51c). At 7a5b74afc:
  verify-zero-file-search.py                          PASSED
  zero_file_search_guard.sh                           rc=0
  falsify-zero-file-search.py over a mktemp copy      every enforced answer module fails when poisoned
  kin-core --test kin_library_resolution              6 passed
  stale-row falsification (a 'kin-bogus' transitional row no lock needs)
                                                      red, "Delete the row", tree restored
  a full precheck at 7a5b74afc is running and lands here as a comment

cargo check --locked --workspace --all-targets                        rc=0
cargo fmt -- --check                                                   clean
cargo clippy --all-targets --all-features -- <45 allows, -D warnings>  rc=0
cargo test --locked -p kin-core --test kin_library_resolution          6 passed
cargo test --locked -p kin-blobs -p kin-model -p kin-search -p kin-vector --lib --tests
  kin-blobs 82 passed, kin-model 340 passed, kin-vector 7 + 2 passed,
  kin-search 66 passed and 1 failed (the liveness test, below)
cargo test --locked -p <each> --no-default-features --lib --tests
  kin-blobs rc=0, kin-model rc=0, kin-vector rc=0, kin-search rc=101 (the same one test)
cargo package -p <each> --no-verify                                    all four rc=0
falsify-hydration-semantics.py and falsify-zero-file-search.py, over a mktemp copy   rc=0
scripts/test-release-workflow-authority.py                             rc=0
actionlint .github/workflows/ci.yml                                    clean
bin/kin-parity --checkout <lane worktree>   release build, binaries report f51bce2e4
  magic       PASS                  27 pass / 0 fail / 0 unreadable
  brownfield  PASS-WITH-ALLOWANCES  12 pass / 1 fail / 0 unreadable   (brownfield/4, FIR-2464)
  firstrun    PASS-WITH-ALLOWANCES  10 pass / 1 fail / 1 unreadable   (firstrun/9 FIR-2580, firstrun/3 FIR-2501)
  FINAL PASS-WITH-ALLOWANCES (NON-CITABLE, local)
  baseline: lane beltfinish's parity on 4e2e8cf83, without this change, same box, same evening:
            identical brownfield and first-run tallies, the same FIR-2580 failure
```

## The registry-release receiver learns kin-vector

The first CI run on this pull request went red in `npm launcher tests`, at
`scripts/test-kin-registry-release-receiver.py`'s
`test_allowed_crates_are_exactly_the_root_registry_boundary`:

```
self.assertEqual(set(validator.ALLOWED_SOURCES), root_registry_dependencies)
AssertionError: Items in the second set but not the first: 'kin-vector'
```

`scripts/validate-kin-registry-release.py` is the receiver that turns a library's registry release
into a pin bump here, and `ALLOWED_SOURCES` is the boundary of what a release may deliver. Its test
pins that boundary to every root `[workspace.dependencies]` entry carrying `registry = "kin"`, and
this pull request gives kin-vector such an entry (it keeps `registry` so that kin-model's packaged
manifest still names the right registry for its kin-vector dependency). So kin-vector joins
`ALLOWED_SOURCES` with the same shape as its siblings, `firelock-ai/kin-vector`, and joins the
unconditional `--crate` list in `.github/workflows/kin-registry-release.yml`, which the receiver's own
tests pin to that set name by name and by count. During the freeze none of these repositories
publishes, so the receiver stays dormant for all of them; retiring it for crates developed here
belongs to the publication work.

Local precheck could not have caught this: it enumerates the gates of `ci.yml`'s `check` job, and
this step runs in the `npm-launchers` job.

## kin-core's workspace scans, and what the first run taught

kin-core carries tests whose subject is every source file in the workspace, which is why `ci.yml`
appends `-p kin-core` to every pull request's scope. The first run on this pull request found two
things they object to in the imported crates, and a third, quieter one surfaced while fixing them.

1. `env_registry::tests::every_kin_env_read_in_workspace_is_registered`: kin-search's soak test
   reads `KIN_SEARCH_SOAK_SEED` directly. The read sits inside `#[cfg(test)] mod tests` in
   `crates/kin-search/src/lib.rs`, so it never reaches a running kin; it joins
   `crates/kin-core/env_var_allowlist.txt` with that reason.
2. `test_env::tests::no_source_outside_the_sanctioned_guard_writes_the_environment`: three
   kin-vector test binaries set `KIN_VECTOR_SIMD` with `env::set_var` before the library's
   `OnceLock` samples it. kin-vector is a standalone publishable crate below kin-core and must not
   take kin-core as a dev-dependency for `EnvVarGuard`, so the three files join
   `crates/kin-core/env_mutation_allowlist.txt` as a stated exception, not a precedent.
   Follow-up: the three tests take the SIMD selection by argument, in a kin-vector pull request
   inside this workspace, and those lines go.
3. `every_pinned_dependency_env_read_is_registered` would have gone quiet. It scans the source of
   kin-db, kin-infer, kin-search and kin-vector for every `KIN_*` literal, because kin-db reads its
   levers through helpers the workspace scan under-counts, and it found that source only in the
   registry cache. When the copy was not there its enumeration arm printed "did not run" and
   passed. A fresh runner never downloads a path dependency, so after this import it would skip
   kin-search and kin-vector, and after kin-db's import, all four. It now scans a workspace
   member's own `crates/<name>/src`, and fails loudly if that tree is missing.

Evidence, run locally on the pushed tree before this push:

```
every_pinned_dependency_env_read_is_registered, the test binary run with an EMPTY CARGO_HOME
  BEFORE the fix  vendored kin-search-0.1.9 source not on disk; enumeration arm did not run
                  vendored kin-vector-0.2.1 source not on disk; enumeration arm did not run
                  (and kin-db, kin-infer the same)                 test result: ok   <- the quiet pass
  AFTER the fix   kin-search-0.1.9: enumeration arm scanning ".../crates/kin-search/src"
                  kin-vector-0.2.1: enumeration arm scanning ".../crates/kin-vector/src"
                  kin-db, kin-infer: still from the registry here, skip legitimately
                                                                   test result: ok
  kin-vector's src moved aside:
                  kin-vector is a workspace member in Cargo.lock, but ".../crates/kin-vector/src"
                  is not a directory                               test result: FAILED  (restored)

cargo test --locked -p kin-core (full), ON THIS HEAD, exit 0:
  ok   895 passed 0 failed 1 ignored   unittests
  ok     2 passed 0 failed 1 ignored   fir2729_bootstrap_seam_probe.rs
  ok     1 passed 0 failed 0 ignored   heap_phase_probe.rs
  ok     4 passed 0 failed 1 ignored   inc2_step_authority_probe.rs
  ok     0 passed 0 failed 1 ignored   init_deep_history_heap_ceiling.rs
  ok     1 passed 0 failed 0 ignored   init_peak_heap_ceiling.rs
  ok     1 passed 0 failed 0 ignored   init_phase_ladder_contiguous.rs
  ok     0 passed 0 failed 1 ignored   init_real_repo_heap_attribution.rs
  ok     6 passed 0 failed 0 ignored   kin_library_resolution.rs
  total 910 passed, 0 failed, 5 ignored
MCP surface contract job, locally            rc=0, 11 of 11 checks, "the served surfaces still
                                             satisfy every main-only assertion"
scripts/test-kin-registry-release-receiver.py   Ran 61 tests, OK
the whole "Test release ordering and npm provenance policy" step   rc=0 (61, 67, 12 and 7 tests OK; node 210 pass, 0 fail)
receiver falsification  workflow array without kin-vector   FAILED (2 contract tests)
                        ALLOWED_SOURCES without kin-vector  FAILED (CI's original assertion)
verify-zero-file-search.py   PASSED        zero_file_search_guard.sh   rc=0
rustfmt --check crates/kin-core/src/env_registry.rs   clean
```

## The four reds from the first run, fixed in three commits

The first CI run on this pull request went red in four checks. All four are this lane's and all four
are fixed here, as three commits on top of the import:

- `33babbd0e` the registry-release receiver accepts kin-vector (npm launcher tests)
- `35318bd7f` the imported crates' test-only environment use (fast gate shards 1 and 3)
- `8e39ff090` the pinned-dependency env scan reads a workspace member's own source

Local precheck could not see any of them: it enumerates the gates of `ci.yml`'s `check` job, and
these live in `npm-launchers` and in the fast-gate shards. That gap is a captain ticket.

CodeQL is red with 14 path-injection alerts, all in `crates/kin-blobs/src/lib.rs`, which is imported
code rather than a change here; the captain is triaging them separately.

## Known, and not a regression

`kin-search::tests::concurrent_writes_with_persistence_do_not_deadlock` is a wall-clock liveness
assertion: a coordinator thread runs the workload and the test does `recv_timeout(45s)` on its
completion channel, then panics naming a deadlock.

```
full suite, shared builder carrying four other lanes' cargo builds
  FAILED  finished in 46.09s   "workload did not complete within 45s"
  FAILED  finished in 46.34s   (the --no-default-features leg, same test)

alone, same box, load averages 37.36 / 28.12 / 19.72, 7 rustc or cargo processes
  ok      finished in 31.52s
  ok      finished in 40.19s
  ok      finished in 32.14s
```

So it is load sensitive rather than a deadlock, and no dependency version behind it moved: the lock
diff carries no third-party change at all. The margin is thin though. Run 2 finished 4.8 seconds
under its own deadline on an otherwise idle slot, and the test has never run on the fleet's shared
builder before, because kin consumed kin-search as a registry dependency and never ran its tests. It is imported source and its deadline is its own design, so nothing about
it was changed, and it was not quarantined: quarantine entry takes a failure at a sha and a
same-code re-run of that job at that sha passing, which only hosted CI can produce.

Migration record, with every dropped path and its reason, the per-crate destination map and the
commit maps: `.kin-coord/reports/monorepo-migration-record-20260911.md`.


troyjr4103:
Local precheck at the pushed head, as promised in the body:

```
kin-precheck: repo=kin tree=.../worktrees/monorepo-kin sha=7a5b74afc3ee9fd66dcba7ec4056f60a5ff5ff33 branch=chore/lane-monorepo-kin
kin-precheck: 22 gates ran, 22 passed, 0 failed, on kin 7a5b74afc3ee9fd66dcba7ec4056f60a5ff5ff33
```

Precheck grades the `check` job's gates only, so it could not see the `npm launcher tests` failure on this run; that one is being fixed on this branch.

troyjr4103:
Flake evidence for `Fast gate test shard (1)`, and the full set is green.

CI run `34658579133` at `8e39ff090de4368f67c024d0acf32994917b5a37`:

- **attempt 1**: shard 1 failed in `kin-core::init_phase_ladder_contiguous::the_admission_ladder_accounts_for_the_whole_of_init` (line 235) with `admit this Git repository: io error: /tmp/.tmpVzZe2f/source/.git/objects/pack/tmp_pack_Y0q8cU: No such file or directory (os error 2)`. The run also cancelled with 3 tests still running, and `init_peak_heap_ceiling` took 183.7s beside it.
- **attempt 2**, a same-code re-run at the same sha: shard 1 **success**, and the run concluded **success**.

`tmp_pack_*` is the name git writes while packing, before it renames, so `init_from_git` enumerated `.git/objects/pack/` and opened an entry git had already moved. That is a time-of-check to time-of-use window in pack enumeration, which needs git packing concurrently to hit, and it is unrelated to this pull request: the three commits here are a validator script, a workflow, two allowlist text files and a `#[cfg(test)]` module, none of them near git admission. The same test passed in the full local `cargo test -p kin-core` on this exact head (910 passed, 0 failed, 5 ignored, exit 0, in the body above).

It is not in `.config/nextest.toml`'s quarantine. Entering quarantine takes a failure at a sha and a same-code re-run of that job at that sha passing, with both run ids on the ticket; the two attempts above are that evidence, should someone want to open one.

PR #1758, opened 2026-09-12
Bring kin-infer, kin-lsp and kin-vfs-core into the workspace with their history

troyjr4103:
Brings kin-infer, kin-lsp and kin-vfs-core into `crates/` as workspace members with their history,
and turns the pins that reached them into path dependencies that keep `version` and `registry`, so
nothing about what is published, or where, changes on day one. Second of three. kin-db follows, and
removes the transitional `[patch.kin]` table.

## What was imported

| package | source repository | frozen sha | imported head | commits | version kept |
|---|---|---|---|---|---|
| kin-infer | firelock-ai/kin-infer | `cd0e44f59877f35dd3cc6cab391317dbd9af757e` | `6c374c09c3c26c3808fc41a658f2ee23a36aea36` | 210 | 0.2.44 |
| kin-lsp | firelock-ai/kin-lsp | `89ca7bcfd4760303610029c8e511cb83b4cd4e68` | `122d4cd557deab9b0749e9fb0070f6c4dd6cfbef` | 134 | 0.7.19 |
| kin-vfs-core | firelock-ai/kin-vfs | `8ac3cfbcd2b3bf354f2a35c1a44f9ae7a17883aa` | `6ab0a64742c5d0ba4ca61b01257b664dc9499086` | 20 of 231 | 0.4.25 |

Only kin-vfs-core comes out of kin-vfs. `--path crates/kin-vfs-core` keeps that crate's history and
leaves the interception, mount and NFS surfaces in that repository, which kin#1748 recorded as
parked in the ecosystem manifest and kin's own docs. Each source repository carried its `<repo>_main_frozen` gate before its head was read, and
all three heads were re-read at 23:39Z and matched.

## History, proved per crate

The comparison is against the import merge's SECOND parent, so kin's own history on the same path is
counted separately rather than inflating it. `full` is
`rev-list --count --full-history <rev> -- <path>`; `simplified` is the same without
`--full-history`, which is the simplification `git log -- <path>` applies.

```
kin-infer     full src=210  dst=210  MATCH   simplified src=200  dst=200  MATCH   kin-own prior on this path=0
kin-infer     tree at the import merge: IDENTICAL (55 paths)
kin-lsp       full src=134  dst=134  MATCH   simplified src=124  dst=124  MATCH   kin-own prior on this path=13
kin-lsp       tree at the import merge: IDENTICAL (48 paths)
kin-vfs-core  full src=21   dst=20   DIFFER  simplified src=18   dst=18   MATCH   kin-own prior on this path=0
kin-vfs-core  tree at the import merge: IDENTICAL (13 paths)
```

Two of those numbers need a sentence each, and neither is a defect.

**`crates/kin-lsp` carries two histories now.** kin-lsp was extracted OUT of kin on 2026-04-03 by
`5ca3505c6`, "Align truth surfaces and depend on standalone kin-lsp", which deleted 20 files and
2,446 lines from that exact path. Thirteen kin-own commits already touch it, so after this import
`git log crates/kin-lsp` reaches the pre-extraction work and the standalone repository's work both.

**kin-vfs-core imports 20 of the 21 source commits that touch its subtree.** The one left out is
`50f01e72c8c8`, a merge whose `crates/kin-vfs-core` subtree is byte identical to its second parent's,
so restricted to that subtree it adds nothing and filter-repo collapsed it. `05b6eea28`, the commit
that carries the content, is an ancestor of that parent and is imported as `e89de58ed`. Separately,
kin-vfs's frozen head changed one workflow file and nothing under the subtree, so the path filter
pruned it and the imported head is the last commit that touched the crate.

Blame at the head reaches the imported history in all three:

```
crates/kin-infer/src/lib.rs        7c731dc6c  Troy Fortin 2026-03-28 feat: initial extraction from kin-db
crates/kin-lsp/src/lib.rs          81936da57  Troy Fortin 2026-04-02 feat: kin-lsp crate scaffold
crates/kin-vfs-core/src/lib.rs     ec2f1370f  Troy Fortin 2026-03-24 feat: initial kin-vfs
```

## Sign-off trailers written during import

Twenty-seven, by the same filter-repo callback PR A used: each commit's own author identity as its
`Signed-off-by`, and nothing else about the commit changed. kin-infer 13 of 210, kin-lsp 14 of 134,
kin-vfs-core 0 of 21. kin-vfs-core had exactly one unsigned source commit and it is the degenerate
merge above, which the path filter dropped, so no trailer was written for that package at all.
Re-read with the DCO job's own loop and regex from `ci.yml:55-66`, falsified both ways first:

```
ee9bc7f7b^1..ee9bc7f7b   commits=211  missing_after=0     (kin-infer)
e9d223491^1..e9d223491   commits=135  missing_after=0     (kin-lsp)
35d849b7f^1..35d849b7f   commits=21   missing_after=0     (kin-vfs-core)
origin/main..HEAD        commits=376  missing_after=0
```

## The pin change, and why the patch table is load-bearing

`kin-lsp` and `kin-vfs-core` gain a `path` beside the `version` and `registry` they already carried.
`kin-infer` gets a root entry for the first time: until now it arrived through kin-db's features and
one direct pin in `crates/kin-migrate` asking for `>=0.2.1`, which becomes `{ workspace = true }`, so
its requirement tightens to `=0.2.44` like every other library. The entry keeps
`default-features = false`, which is what that pin said: a no-op today because kin-infer declares
`default = []`, and the guarantee still standing if it ever gains one.

`[patch.kin]` gains exactly one row, kin-infer's, and it is load-bearing rather than assumed. kin-db
is still a registry crate until PR C, and its own published requirements, read from the sparse index
cache and identical at 0.7.112 and 0.7.114, are:

```
0.7.112  kin-infer '0.2.44' normal optional   kin-model '=0.7.28'   kin-search '0.1.9'   kin-vector '=0.2.1'
0.7.114  kin-infer '0.2.44' normal optional   kin-model '=0.7.28'   kin-search '0.1.9'   kin-vector '=0.2.1'
```

`crates/kin-infer` is 0.2.44, which satisfies `^0.2.44`, so the patch applies. A patch whose version
does not satisfy the requirement is dropped with one warning line and the build succeeds against the
registry copy, which is the silent failure this table exists to prevent, and which `cargo tree -d`
listing no Kin library twice is what actually proves.

kin-lsp and kin-vfs-core are NOT in the table, and that was measured rather than assumed. Rows for
them went in first, on the assumption that the table should list every imported library. The
assumption was wrong: no registry crate depends on either, kin's own crates reach both by path, and
nothing requests them from the registry for a patch to redirect. Removing both and re-locking moved
no line of `Cargo.lock` and produced no cargo output at all, not even an unused-patch warning:

```
control: the tree as committed re-locks with no change and no warning
  cargo said: 0 line(s) of its own
  lock unchanged
arm: drop the kin-lsp and kin-vfs-core rows from [patch.kin]
  removed both rows
  cargo said:
  lock UNCHANGED: the two rows change no resolution
```

So they came out, in a commit of their own, and the comment above the table records what was
measured. The pull request that removes the whole table has one less thing to explain.

`TRANSITIONAL_REGISTRY_LIBRARIES` in `crates/kin-core/tests/kin_library_resolution.rs` drops to
kin-db alone. That guard reports a row no lock needs, so leaving one in fails.

## The re-versioning trap kin-vfs-core would have hit

Its manifest inherited six fields from kin-vfs's root. Two differ from kin's, read off the merged
tree where the v0.7.15 release has already moved the workspace version, and one does not exist
there at all, so a naive move would have changed the artifact silently or failed in a way that reads
like a broken manifest.

| field | kin-vfs root | kin root | written explicitly |
|---|---|---|---|
| version | 0.4.25 | 0.7.15 | `0.4.25` |
| edition | 2021 | 2021 | `2021` |
| rust-version | 1.85 | (none) | `1.85` |
| license | Apache-2.0 | Apache-2.0 | `Apache-2.0` |
| authors | Troy Fortin | Troy Fortin | written out |
| repository | firelock-ai/kin-vfs | firelock-ai/kin | `https://github.com/firelock-ai/kin-vfs` |

Dependencies too: `serde`, `thiserror`, `tracing`, `parking_lot` and `tempfile` are the same string
in both roots and keep `{ workspace = true }`; `sha2` is "0.10" here against kin's "0.11", and `lru`
and `rmp-serde` are not in kin's root at all, so all three are written out at the versions kin's lock
already carried (sha2 0.10.9, lru 0.18.2, rmp-serde 1.3.1). kin-vfs's root also set
`[profile.release] panic = "abort"`, which is moot for a library and is recorded in the migration
record rather than moved.

## The lock

Six packages join, all reached only through kin-infer's dev-dependency on `tokenizers 0.20.4`:
`indicatif 0.17.11`, `itertools 0.11.0`, `itertools 0.12.1`, `number_prefix 0.4.0`,
`rayon-cond 0.3.0` and `tokenizers 0.20.4`. `esaxx-rs` gains a `cc` dependency for the same reason.
kin-db pulls `tokenizers 0.23.1`, so the lock carries two, which `deny.toml` reports as a warning
rather than a failure (`multiple-versions = "warn"`).

`fuzz/Cargo.lock` does not move: it is a detached workspace that reaches kin-model through kin-parser
and carries none of these three.

One third-party edge moved on the first re-lock and was put back by hand. `gix-imara-diff` went from
`hashbrown 0.16.1` to `0.17.1`, both of which the lock already carried, which is index drift and not
something this change asks for. The lock was edited back and re-verified with
`cargo metadata --locked`, so every line that moves here is a line this import causes.

## Zero file-search

The checker scans these crates for the first time. Eleven files get an entry, ten of them PINNED to
the functions and expressions that hold their filesystem primitives rather than exempted whole, so
every other function in each file, including any added later, stays graded.

| file | treatment | pins |
|---|---|---|
| `crates/kin-infer/src/lib.rs` | pinned | allow_fn: `load_sharded_index`, `load_weight_bytes` |
| `crates/kin-infer/src/resource.rs` | pinned | allow_fn: `read_i64_file`, `read_u64_file`; allow_match: `if let Ok(contents) = std::fs::read_to_string("/sys/fs/cgroup/cpu.max") {`, `if let Ok(contents) = std::fs::read_to_string("/sys/fs/cgroup/memory.max") {`, `let contents = std::fs::read_to_string("/proc/meminfo").ok()?`, `let out = std::process::Command::new("ioreg")`, `let out = std::process::Command::new("sysctl") x2` |
| `crates/kin-vfs-core/src/containment.rs` | pinned | allow_fn: `resolve` |
| `crates/kin-infer/src/watchdog.rs` | pinned | allow_match: `std::process::abort()` |
| `crates/kin-lsp/src/discovery.rs` | pinned | allow_match: `std::process::Command::new(path)` |
| `crates/kin-lsp/src/lifecycle.rs` | pinned | allow_match: `std::process::Stdio`, `tokio::process::{Child, ChildStderr, Command}`, `Command::new(command)`, `std::process::id()` |
| `crates/kin-lsp/src/client.rs` | pinned | allow_match: `tokio::process::ChildStdin`, `tokio::process::ChildStdout x2` |
| `crates/kin-lsp/src/enrichment_integrity_tests.rs` | whole-file, not runtime | - |
| `crates/kin-vfs-core/src/provider.rs` | pinned | allow_match: `read_dir( x2`, `read_link( x2` |
| `crates/kin-vfs-core/src/tree.rs` | pinned | allow_match: `read_dir(&rel)`, `read_link( x2` |
| `crates/kin-vfs-core/src/writer.rs` | pinned | allow_match: `read_dir( x2`, `read_link( x2` |

Three are worth naming. kin-infer's `resource.rs` pins are hardware and quota probes (sysctl, cgroup
`cpu.max` and `memory.max`, `/proc/meminfo`) read to size batches to the host, the same class as the
existing `crates/kin-cli/src/capability.rs` entry. kin-vfs-core's `containment.rs` pin is a security
boundary: the function canonicalizes a path and refuses one that escapes the projection root, so it
decides whether a path may be served and never what it holds. And `provider.rs`, `tree.rs` and
`writer.rs` trip on ContentProvider METHOD names, `read_dir` and `read_link`, on a graph-backed
provider, which is a name collision with `std::fs` rather than file search. The one whole-file entry
is kin-lsp's `enrichment_integrity_tests.rs`, which is `#[cfg(test)] mod enrichment_integrity_tests;`
in `lib.rs`, test code the checker's name rule misses because it matches `_test.rs` and not
`_tests.rs`.

Every one of the eleven is load-bearing, proved by dropping each in turn on a fresh copy of the
tree: the checker goes red and names that file, with the unpoisoned control green so a red arm
means something.

```
control: the tree as committed                        rc=0  Verification PASSED
drop crates/kin-infer/src/lib.rs                      rc=1  RED and names it
drop crates/kin-infer/src/resource.rs                 rc=1  RED and names it
drop crates/kin-infer/src/watchdog.rs                 rc=1  RED and names it
drop crates/kin-lsp/src/discovery.rs                  rc=1  RED and names it
drop crates/kin-lsp/src/lifecycle.rs                  rc=1  RED and names it
drop crates/kin-lsp/src/client.rs                     rc=1  RED and names it
drop crates/kin-lsp/src/enrichment_integrity_tests.rs rc=1  RED and names it
drop crates/kin-vfs-core/src/containment.rs           rc=1  RED and names it
drop crates/kin-vfs-core/src/provider.rs              rc=1  RED and names it
drop crates/kin-vfs-core/src/tree.rs                  rc=1  RED and names it
drop crates/kin-vfs-core/src/writer.rs                rc=1  RED and names it
```

The first run of that falsification said all eleven were NOT load-bearing, and it was wrong. Each
arm copied `crates` and `scripts` to a fresh directory, poisoned the COPY's allowlist, and ran the
WORKTREE's checker with the copy as its root. `ALLOWLIST_PATH` is joined to the script's own
directory while `KIN_ROOT` is the argument (`verify-zero-file-search.py:32-34`), so every arm
scanned a poisoned tree against an unpoisoned allowlist. Eleven identical answers was the tell.
Running the copy's own checker is the fix, and the numbers above are from that run. Worth a line in
`docs/traps.md`.

## kin-core's env scans, and a crate-scoped test-only exemption

kin-core's environment-write scan grades test code on purpose. Its failure message says test code
must go through `kin_core::test_env::EnvVarGuard`, because an unguarded `set_var` in a test is the
racy case it exists to catch. The imported libraries are the one class that cannot obey it:
kin-vector, kin-infer and kin-db sit BELOW kin-core and cannot take it as a dev-dependency.

kin-vector landed with three per-file lines, one per test binary, which worked because each is a
small file that does nothing else. kin-infer cannot be handled that way: its writes are in
`#[cfg(test)] mod tests` inside a 7,500-line `src/lib.rs` and in three `tests/` binaries, and a
per-file line on that `lib.rs` would also excuse every future write in its product code, which is
the write the scan exists to catch.

So `env_mutation_allowlist.txt` gains a second line form, `crates/<name> tests-only`, which exempts
that crate's `tests/` binaries and its top-level `#[cfg(test)]` items and nothing else. Product code
in the crate stays graded, every other crate keeps exact-path semantics, and a line shaped like the
new form but malformed is refused rather than read as a path that matches nothing. kin-vector's
three per-file lines convert to it, because the crate-scoped form is the narrower statement here,
not the wider one: it covers test code in three files and stops, where three per-file lines cover
everything those files ever hold.

Finding the test ranges leans on rustfmt, which CI enforces: a top-level item's closing brace is a
`}` alone at column 0. Whatever fools that ends a range early and so grades MORE, never less, and a
`#[cfg(test)] mod x;` declaration is not a block, so the file it names is graded as product code.

Four arms, each on inputs the test builds rather than on the tree, all green in the run below:

```
a_product_write_in_a_tests_only_crate_is_still_flagged     a product write is flagged, and so is a write after the test module closes
a_test_module_write_outside_a_tests_only_crate_is_flagged  a test-module write elsewhere is flagged, and so is one in its tests/ binary
a_test_module_write_in_a_tests_only_crate_passes           a test-module write here passes, and so does one in its tests/ binary
a_malformed_tests_only_line_is_refused                     four malformed spellings are refused
```

Beyond those four unit arms, the exemption is falsified end to end against the REAL scan on the real
tree. The run mutates live allowlist files, so it ran alone and restored from saved copies, and it
ends by diffing both files against those copies rather than trusting the restore:

```
control: both scans on the tree as committed
  read scan      rc=0   1 passed; 0 failed
  mutation scan  rc=0   1 passed; 0 failed

arm 1: drop the twelve kin-infer names from env_var_allowlist.txt
  read scan      rc=101 every_kin_env_read_in_workspace_is_registered panicked, naming
                        crates/kin-infer/tests/embed_scale_hang.rs

arm 2: drop `crates/kin-infer tests-only`
  mutation scan  rc=101 no_source_outside_the_sanctioned_guard_writes_the_environment panicked,
                        naming crates/kin-infer/src/lib.rs

arm 3: drop `crates/kin-vector tests-only`
  mutation scan  rc=101 the same test panicked, naming
                        crates/kin-vector/tests/hnsw_determinism_simd.rs

restored: env_var_allowlist.txt and env_mutation_allowlist.txt both byte identical to the copies
```

Arm 2 is the one that matters most: it shows the crate-scoped form is what keeps
`crates/kin-infer/src/lib.rs` quiet, so the exemption is doing work rather than sitting beside a
scan that would pass anyway.

### The read scan found one name the plan did not

The full `cargo test -p kin-core` on the first tree failed on exactly one thing:
`every_kin_env_read_in_workspace_is_registered`, naming `KIN_EMBED_ORPHAN_CHECK` in
`crates/kin-infer/tests/embed_scale_hang.rs`. Eleven of the twelve names were predicted by a regex
pre-scan; this one was not, which is why the real test is what decides.

It joins `env_var_allowlist.txt` rather than the registry, and the reasoning matters because a
reviewer will ask. It is one of kin-infer's six embed watchdog knobs. All six are already in
`dependency_env_allowlist.txt` with a reachability note, and
`the_unreachable_watchdog_knobs_stay_out_of_the_registry` asserts they are allowlisted AND
unregistered, because neither kin nor the pinned kin-db constructs `EmbedConfig` or `EmbedWatchdog`,
so nothing in a kin process reads any of them. Importing the crate did not change that; it changed
which scan reads the file. Only this one of the six is visible to the workspace scan because
`crates/kin-infer/src/watchdog.rs` reads all six through helpers that take the name as an argument,
while the test binary spells this one out. Registering it would document a lever an operator cannot
reach.

The other eleven are kin-infer's own test-binary reads: the embed scale-hang harness, the
determinism probe and the batch probes.

## Windows

kin-vfs-core joins kin's native Windows job. Its own repository ran `cargo test -p kin-vfs-core` on
`windows-latest` as a REQUIRED check on every pull request, not a nightly cron, and that leg went
with the rest of that CI when the crate moved here. `containment.rs` is the reason it mattered:
`host_relative` carries a `cfg(not(unix))` arm that REFUSES a path whose bytes are not UTF-8 rather
than lossily decoding it into a different file, and that arm is dead code on every other runner.

Windows test execution still moves from pre-merge to post-merge, because kin's native Windows jobs
run on a push to main. The pull-request `windows-cross-check` compiles the crate through kin-cli and
kin-daemon but runs nothing. kin-infer and kin-lsp are not here: neither had Windows CI and neither
carries a `cfg(windows)` arm.

## One commit here is not this lane's, and what it fixes

`fix(infer): gate metal tests to macos, remediate deny and CodeQL alerts` answers the three checks
this import turned red the first time it reached hosted CI, and it was written with run evidence
this lane did not yet have. It is cherry-picked here with a sign-off, and its three claims are
checked against its own diff rather than taken on trust.

**19 cfg gates** move from `#[cfg(feature = "metal")]` to
`#[cfg(all(feature = "metal", target_os = "macos"))]` across 21 test files. That closes a gap this
import created: kin-infer's SOURCE already gated every metal and accelerate site on
`all(feature, target_os = "macos")`, but its TESTS gated on the feature alone, and they had never
been compiled under `--all-targets --all-features` on Linux, because kin-infer's own CI ran
`--all-targets` on default features. kin's `Fast gate lint and policy` does exactly that
combination on ubuntu.

**`deny.toml` gains exactly one line**, RUSTSEC-2025-0119: number_prefix 0.4.0 unmaintained,
transitive via indicatif 0.17. That is this import's doing too, and more than the duplicate-version
warning noted above: kin-infer's dev-dependency on `tokenizers 0.20.4` drags `indicatif 0.17.11`,
which drags `number_prefix 0.4.0`, and the advisories check is an error rather than a warning. The
better fix is not an ignore. Moving that dev-dependency to the `tokenizers 0.23` kin-db already
pulls would remove the duplicate AND the advisory, and it is left out of a relocation pull request
on purpose: changing an imported crate's dependency is a product decision, the SweRank probe is its
only consumer, and it belongs in a kin-infer change inside this workspace, which is the point of the
import.

**CodeQL** is a `salt` to `seed` rename in the synthetic-sequence helpers, which changes no
arithmetic at all (`wrapping_add(x.wrapping_mul(40503))` either way, only the identifier moves), plus
two `forward(&[ids.to_vec()], ...)` temporaries bound to locals first. The ten alerts under the
imported crates are 8 `rust/hard-coded-cryptographic-value` in `embed_stage_split.rs`,
`embed_forward_determinism.rs`, `embed_bucketing_probe.rs` and
`embed_batch_size_invariance_probe.rs`, and 2 `rust/access-invalid-pointer` in
`embed_speed_profile.rs:61` and `swerank_self_retrieval.rs:41`. All six files are in the commit: it
changed code for them and dismissed nothing.

## Other facts a reviewer should have

- **kin-infer's test binaries are mostly inert off `metal`.** Seventeen of its 21 run zero tests
  when the feature is off, which is every ubuntu run and every `-p kin-infer` run, because they are
  feature-gated at the file level and kin-infer declares `default = []`. Same coverage its own CI
  had, which ran plain `cargo test`.
- **kin-lsp's three integration tests skip, here and on CI.** Each guards on `which::which` finding
  `rust-analyzer` AND `rust-analyzer --version` exiting zero; the rustup proxy exists on a hosted
  runner and the component does not, so the guard is false. `scripts/ci-install-language-servers.sh`,
  which the `check` job runs, installs pyright and typescript-language-server only, and no caller of
  `.github/actions/rust-toolchain` asks for the `rust-analyzer` component. Worth knowing for later:
  `rust_analyzer_integration` resolves its workspace by walking two directories up from
  `CARGO_MANIFEST_DIR`, which inside kin is kin's own root, so the day someone adds that component
  it starts rust-analyzer on the whole workspace.
- **`crates/kin-migrate`'s direct pin becomes the inherited one.** `kin-infer = { version =
  ">=0.2.1", registry = "kin", default-features = false }` becomes `{ workspace = true }`, and the
  root entry carries the `default-features = false`.
- **The root manifest's comment that kin-infer/metal "cannot compile" on Linux was stale and is
  corrected.** Every metal and accelerate site is gated `all(feature = ..., target_os = "macos")`,
  including the `pub mod metal_backend;` declaration, and the metal, objc2 and block crates are
  declared only under `[target.'cfg(target_os = "macos")'.dependencies]`, so off macOS the feature
  turns on nothing rather than failing to build. The rule the comment states still holds on its own
  reason, which is that a `[workspace.dependencies]` entry cannot be target-gated.
- **kin-vfs-core is the one imported library with no crate-level LICENSE, NOTICE or README.** The
  kin-vfs repository kept those at its root, outside the filtered subtree, so there was nothing to
  import. Its manifest carries `license = "Apache-2.0"`, which is what the published 0.4.25 was built
  from, so publication is unaffected. Recorded for the publication lane rather than invented here.
- **The imported commit messages carry 124 em dashes**, written before those repositories adopted
  the register. They are left exactly as they are: rewriting an imported commit message is the one
  thing this import exists not to do, and no gate reads them. Everything this pull request itself
  writes, the five commit messages and this body, carries none.
- **Twenty-five paths of standalone-repository scaffolding were dropped**, none of them source: both
  crates' `.github/`, `.kin-release/`, `deny.toml`, `rust-toolchain.toml` (all three pinned the same
  1.96.0 kin's root pins), community files, kin-lsp's `.cargo/config.toml` and kin-infer's tracked
  lock, plus five scripts whose subject was that repository's own workflows and publication.
  kin-infer's `gpu-sweep.sh` and `run-tests.sh` stay: they are crate development tooling and its
  README names the second as the way into the Metal suite. The migration record lists every path
  with its reason.

## Verification

Every tail below is from head `58d1a2142`, which is what this pull request carries. An earlier
pass graded two other trees and they are named here so nothing reads as hidden: `a8127d8a0`, this
lane's five commits before main was merged, and `33a8b70bc`, the first merge of PR A's landing.
That first merged head was then rebased onto main by another session while this lane was stopped on
a usage limit. The rebase linearized the imported repositories' own merge and empty commits, which
is exactly the gap between the full and simplified counts above, so the full-history numbers would
have been false on it. It was restored to the merge shape, verified, and force-pushed once with a
lease; the counts, trees, DCO and commit maps in this body are all re-measured on `58d1a2142`.

| gate | result on `58d1a2142` |
|---|---|
| `cargo test --locked -p kin-core` (the three workspace scans) | rc=101, test result: FAILED. 898 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out; finished in 41.11s |
| `cargo test --locked -p kin-infer -p kin-lsp -p kin-vfs-core` | rc=0, 299 passed across the three crates |
| `cargo test -p <crate> --no-default-features --lib --tests`, all three | kin-infer rc=0, kin-lsp rc=101, kin-vfs-core rc=0 |
| `cargo package -p <crate> --no-verify --locked`, all three | kin-infer rc=0, kin-lsp rc=0; kin-vfs-core through hosted `Library packages`, success |
| `cargo tree --locked -d` | through hosted CI's own resolution, `Library packages` and the fast gate both success; the local run is a post-landing comment |
| `cargo clippy -p kin-infer --all-targets --all-features --target x86_64-unknown-linux-gnu` | `Fast gate lint and policy` success on a real Linux host, which is that exact combination; the local cross-compile is a post-landing comment and redundant |
| `bin/kin-precheck kin --repo-dir kin=<worktree>` | 22 gates enumerated, 22 ran, 20 passed, 2 failed on the pre-merge head; both fixed here, and hosted CI re-ran the same gate list green on `58d1a2142` |
| the MCP surface contract, graded against locally built binaries | hosted `MCP surface contract` success, plus the grader self-test rc=0 locally |
| `scripts/test-kin-registry-release-receiver.py` | rc=0, 61 tests |
| the npm-launchers policy step's `node --test` list, 16 files | rc=0, 323 tests, 0 failed |
| `node scripts/check-kin-vfs-compat.mjs` | rc=0, kin-vfs-core 0.4.25 on both sides, 8 sites in 3 files |
| `python3 scripts/verify-zero-file-search.py` and its guard | rc=0, PASSED, and the guard rc=0 with 73 answer modules graph-clean |
| gitleaks 8.30.1 over the imported range | rc=0, no leaks found |
| the DCO job's own loop and regex over `origin/main..HEAD` | 376 commits, missing_after=0 |
| `bin/kin-parity` | post-landing comment; hosted CI graded the workspace here |

### The two timing-sensitive tests, measured rather than argued about

Two tests failed once each inside a full local suite on a box carrying this lane's gate queue beside
four other lanes' builds. Both pass alone, repeatedly, and neither was changed or quarantined.

```
memory_pressure::tests::concurrent_work_upserts_share_one_read_modify_publish_cycle
  alone: rc=0, rc=0, rc=0, rc=0, rc=0      in the suite: FAILED at 41.11s, where a passing run took 21.94s
  a recv_timeout(1s) thread handoff, from kin#1204 on 2026-08-28, in a file this branch never touches

enrichment_integrity_tests::failed_drop_cleanup_invalidates_later_barriers   (imported kin-lsp)
  alone: rc=0, rc=0, rc=0, rc=0            in the suite: FAILED at 0.82s, so not a timeout
  it asserts a write into a pipe whose reader closed returns EPIPE rather than landing in the kernel
  buffer, which is timing and buffering; from kin-lsp#96 on 2026-09-08
```

The second is the honest cost of importing a crate: its tests run in kin's CI now, and this one can
flake. Where that would bite is worth being precise about, because neither hosted form absorbs it.
`.config/nextest.toml` sets `flaky-result = "pass"` but sets `retries` nowhere outside comments and
carries zero overrides, so the fast-gate shards grade a flake as a plain red; and `Library packages`
runs the exact plain `cargo test` command that flaked here. Both passed on this head on the first
attempt, so the odds are observed rather than guessed. A future red on that test name is this flake,
not a regression, and the fix belongs in a kin-lsp change inside this workspace.

Every cargo command ran through the fleet's builder slots
(`.kin-coord/bin/heavy-slot.sh monorepob kin -- <absolute command>`), and precheck was named by
ABSOLUTE path with `--repo-dir`, so it could not resolve inside the worktree and exit 126 having
graded nothing. Exit codes are recorded beside their logs rather than inferred from output.


troyjr4103:
Post-landing tail 1 of 2, as promised on the body: the kin-infer cross clippy.

It **could not run** on this box, and that is a build-environment limit rather than a result:

```
cargo clippy --locked -p kin-infer --all-targets --all-features --target x86_64-unknown-linux-gnu
error: failed to run custom build command for `onig_sys v69.9.3`
  cc-rs: command did not execute successfully: "x86_64-linux-gnu-gcc" ... regexec.c
```

`--all-targets` pulls kin-infer's dev-dependencies, so `tokenizers 0.20.4` brings `onig` and its C
library `onig_sys`, which needs a Linux C cross-compiler this Mac does not have. It is the same
class lane monorepo recorded for kin-db's `ring`. The earlier `--lib` measurement succeeded because
`--lib` compiles no dev-dependencies.

Nothing is owed here, because hosted CI already answered the question better. `Fast gate lint and
policy` runs `cargo clippy --all-targets --all-features` with kin's own allow list on a real Linux
host, and it passed on `58d1a2142`. It is also the check that went red on `33a8b70bc` before the
metal test attributes were gated to macOS, so its green is direct evidence for that fix rather than
an inference.

Recorded rather than retried: a cross-compile from this host cannot speak for a sys crate it cannot
build, and pretending otherwise would put a number on this pull request that no command produced.

troyjr4103:
Post-landing tail 2 of 2, as promised on the body: `bin/kin-parity`.

```
kin-parity: magic       PASS            172.4s  27 pass / 0 fail / 0 unreadable
kin-parity: brownfield  PASS-WITH-ALLOWANCES  373.6s  12 pass / 1 fail / 0 unreadable
kin-parity: firstrun    PASS-WITH-ALLOWANCES  106.1s  10 pass / 1 fail / 1 unreadable
kin-parity: FINAL PASS-WITH-ALLOWANCES in 1267.6s  (NON-CITABLE)  allowances: firstrun/9 FIR-2580, firstrun/3 FIR-2501, brownfield/4 FIR-2464
```

Read the suite lines, not the wrapper's exit code: the wrapper can exit 0 on a run whose
suites did not all pass, so the lines above are the verdict and the wrapper reported
`rc=0`. A local parity run is NON-CITABLE by design; hosted CI is the gate of record, and
this pull request's own check set was green by name over all 50 checks before it landed.

It ran on `fc19a79ced9020f7576262d4458c52e37b2387f2`, which is main at the merge of this pull request plus the two workflow
files and one test file of kin#1763. No binary compiles any of those three, so the `kin` and
`kin-daemon` this built are the code this pull request landed.

PR #1759, opened 2026-09-12
Anchor a review note on the entity that holds the line it names

troyjr4103:
`kin_review_note_add` and `kin_review_discuss` accept a `line` beside `file_path`, and nothing read it: the anchor stored was the path, so the line a reviewer pointed at was dropped. A caller sending both was told nothing.

A `file_path` with a `line` now resolves to the entity whose span holds that line, the innermost when spans nest, because the class containing a method contains the method's lines too and a reader pointing at one of them means the method. The note anchors on that entity, which survives the declaration moving in the file. A line no entity holds, and a path with no line, fall back to the artifact scope as before, and the `scope` the answer already carries says which happened, so a caller can tell an entity anchor from a file one.

Both parameters keep answering and now say they are deprecated, through the `deprecations` array beside `degradations`, naming `scope: entity:<uuid>` as what replaces them.

Part of FIR-3578.

### DIFF
--- a/crates/kin-mcp/src/handlers/review.rs
+++ b/crates/kin-mcp/src/handlers/review.rs
@@ -734,7 +734,7 @@ fn plan_review_note_add<G: GraphStore>(
 
     let review_id = parse_review_id(args, "review_id")?;
     let body = get_string_param(args, "body")?;
-    let scope = parse_optional_scope_arg(args)?;
+    let scope = parse_optional_scope_arg(args, store)?;
     let author = parse_identity_arg(args, "author", "author_kind", "mcp-client");
     existing_review(store, &review_id)?;
 
@@ -747,12 +747,13 @@ fn plan_review_note_add<G: GraphStore>(
         created_at: Timestamp::now(),
     };
 
-    let result = serde_json::json!({
+    let mut result = serde_json::json!({
         "note_id": note.note_id.to_string(),
         "review_id": review_id.to_string(),
         "scope": scope.map(|s| s.to_string()),
         "author": author.name,
     });
+    record_path_anchor_deprecations(&mut result, args, "kin_review_note_add");
     let json = serde_json::to_string_pretty(&result).map_err(McpError::Json)?;
     Ok(PlannedReviewEvent {
         action: "review.note",
@@ -794,7 +795,7 @@ fn plan_review_discuss<G: GraphStore>(
 
     let review_id = parse_review_id(args, "review_id")?;
     let body = get_string_param(args, "body")?;
-    let scope = parse_optional_scope_arg(args)?;
+    let scope = parse_optional_scope_arg(args, store)?;
     let author = parse_identity_arg(args, "author", "author_kind", "mcp-client");
     existing_review(store, &review_id)?;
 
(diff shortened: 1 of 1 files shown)


troyjr4103:
Local verification for this change, posted after the merge so the evidence sits with the pull request.

Everything below ran on `0b2f22f53a54e30135b5ceef7c937a5922d481f2`, this branch's head before it was rebased onto `10cd9d4c4`. That rebase produced the landed head `23f6c03404172ee27c38dedc5b885ff21e6bb89c` and changed no bytes: the patch-id is `ca4921a1f6a03d472a72f8ee53c2746237ad47b6` on both sides and `crates/kin-mcp/src/handlers/review.rs` hashes to `f4ff6edff2746d26` on both, which is the same sha256 the falsification arms below restored to. Nothing local ran on `23f6c034` itself, and its grade here is hosted CI.

## Falsification

Three arms, each mutating the behaviour one assertion guards, each run against the committed head and each restored by sha256 rather than by `git checkout`. All three ended `restored by hash, tree matches the committed head`, and the run ended `ALL ARMS RUN`.

D1 takes the outermost span instead of the innermost:

```
thread '...::a_path_anchor_resolves_to_the_entity_that_holds_the_line' panicked at
crates/kin-mcp/src/handlers/review.rs:1730:9:
assertion `left == right` failed: the innermost span that holds the line wins
  left: "entity:3d18a008-3079-4f8e-95b2-5bb367bd6e7a"
 right: "entity:56672600-7cb5-44ca-8a92-ae2453c53233"
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1027 filtered out
```

D2 drops containment's upper bound, so a line past every span still matches:

```
panicked at crates/kin-mcp/src/handlers/review.rs:1735:9:
assertion `left == right` failed: a line no entity holds falls back to the artifact
  left: "entity:13c3102f-02cf-422a-bbe2-56f657d4dcb0"
 right: "artifact:src/editor.rs"
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1027 filtered out
```

D4 reads a missing `line` as line 1:

```
panicked at crates/kin-mcp/src/handlers/review.rs:1738:9:
assertion `left == right` failed: a path with no line cannot name an entity
  left: "entity:45e02361-4774-4dcc-8d95-c3f8d57b0973"
 right: "artifact:src/editor.rs"
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1027 filtered out
```

## Gates

```
at 0b2f22f53
fmt rc=0
clippy rc=0
test rc=0
test result: ok. 1028 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 53.94s
kin-precheck: 22 gates ran, 22 passed, 0 failed, on kin 0b2f22f53a54e30135b5ceef7c937a5922d481f2
```

clippy is the form `ci.yml` runs, `RUSTFLAGS="-D warnings" cargo clippy -p kin-mcp --all-targets --all-features` with the 45 allows assembled out of `.github/clippy-allowed-lints.txt`, and it reported zero.

## The caller this was for

kinlab's control plane at `services/control-plane/src/local-kin-gateway.ts:4552-4560` calls `kin_review_discuss` with both `file_path` and `line`, so it is the caller whose line was being dropped, and it gets an entity anchor with no edit on its side. Its sibling at `:4457-4465` reaches `kin_review_note_add` and passes neither. kin-editor calls neither tool. Both parameters keep answering and now carry a deprecation naming `scope: entity:<uuid>`, removed after `0.7.16`, so no paired change is needed in either repo before then.

PR #1774, opened 2026-09-12
docs(claude): add plugin requirements

### LINKED ISSUE #1733: Add a Requirements section to the Claude Code plugin README
The Codex and Cursor plugin READMEs each tell a reader what they need before installing (Node
version, supported platforms). The Claude Code one doesn't, so a reader finds out the hard way
instead of up front.

**Files:** `plugins/kin-claude/README.md` (currently ends after a "Links" section with no
Requirements heading). `plugins/kin-codex/README.md` (around line 50) and
`plugins/kin-cursor/README.md` (around line 51) both already have a "## Requirements" section to
match.

**Expected behavior:** `plugins/kin-claude/README.md` gains its own "## Requirements" section,
consistent with (or corrected for any real difference from) what the other two say.

**How to verify:** Read the three plugin READMEs side by side. A reader following any of the three
should land on a consistent statement of what they need before they start.

**Size:** an evening.

kocaemre:
## Summary
- Add the missing `## Requirements` section to `plugins/kin-claude/README.md`.
- Match the Node, first-run network, and supported-platform wording already used by the Codex and Cursor plugin READMEs.

Closes #1733

## Test Plan
- PASS: `python3 - <<'PY' ... PY` verified the Claude, Codex, and Cursor plugin READMEs contain the same Requirements section text.
- PASS: `git diff --check`
- NOT RUN: full Rust workspace checks (`cargo test`, `cargo clippy`, `cargo fmt`) because this is a docs-only plugin README change.

### DIFF
--- a/plugins/kin-claude/README.md
+++ b/plugins/kin-claude/README.md
@@ -36,6 +36,11 @@ Every response names the graph state that produced it, and an empty result says 
 absence can be trusted. A graph gap is reported as a gap rather than filled in from raw file
 search.
 
+## Requirements
+
+Node 20 or newer for `npx`, and network access on the first run to fetch the Kin release.
+macOS, Linux, and Windows x64 are supported. On Windows, WSL2 is the recommended path.
+
 ## Links
 
 - Repository: https://github.com/firelock-ai/kin


troyjr4103:
Thanks for picking this up, Emre! The prerequisites should be visible before someone tries the plugin. Keeping them aligned with the Codex and Cursor instructions makes that clearer. The focused docs checks are appropriate here; I'll handle the remaining maintainer checks.

troyjr4103:
Merged into main, thanks again, Emre! If you spot another unclear step while setting up the plugin, I'd be glad to hear about it.

PR #1778, opened 2026-09-12
fix(cli): skip profile for top-level display flags

### LINKED ISSUE #1736: Let `kin --help` and `kin --version` skip the repository-boundary check
Running `kin --help` or `kin --version` anywhere under a nested checkout (a Kin repository sitting
inside another one) prints a multi-line warning about refusing to bind a parent store, before the
help or version text a person actually asked for ever shows up. It reads like an error from a
command that hasn't done anything yet.

**Files:** `crates/kin-cli/src/main.rs`, the `run()` function (the call to
`apply_repository_profile_at`, a few lines before argument parsing happens). `crates/kin-core/src/layout.rs`,
the `eprintln!` inside `KinLayout::discover` (roughly lines 52 to 72), which is where the warning
text actually comes from.

**Expected behavior:** `kin --help`, `kin -h`, `kin --version` and `kin -V` print only their own
text, with no store-boundary warning. Every other command keeps exactly today's behavior.

**How to verify:** Run `kin --help` inside a directory nested under another Kin-admitted
repository, which is what triggers the warning today, and confirm only help text prints. Run an
ordinary command in the same spot and confirm the warning still appears exactly as before.

**Size:** an evening, if scoped narrowly to the bare-flag case. Read the comment directly above the
profile-resolution call in `main.rs` first; it explains why that call runs unconditionally today.

kocaemre:
## What

Skip repository profile discovery for the bare top-level display flags `kin --help`, `kin -h`, `kin --version`, and `kin -V`, and add an integration regression for the nested-repository parent-store warning case.

## Why

Closes #1736. In a nested checkout under a parent `.kin/` store, those display-only commands could print the parent-store boundary warning before the help/version text even though the user only asked for display output.

## How

- Detect the exact one-argument top-level display flag case before `apply_repository_profile_at` runs.
- Leave subcommand help and all non-bare invocations on the existing repository-boundary path.
- Cover both sides with a nested git repository fixture: top-level display flags stay warning-free, while `kin resolve --help` still emits the boundary warning.
- Add an Unreleased changelog note for the user-facing CLI behavior.

## Testing

- [ ] `cargo test` passes
- [ ] `cargo clippy` produces no warnings
- [x] `cargo fmt` applied
- [x] Every commit carries a `Signed-off-by` trailer (`git commit -s`); the DCO
      check fails without it, and [CONTRIBUTING.md](https://github.com/firelock-ai/kin/blob/main/CONTRIBUTING.md#dco-sign-off)
      shows how to add it after the fact

Evidence run locally:

```text
export PATH=/root/.cargo/bin:$PATH
cargo fmt --check
# passed

cargo test -p kin-cli --test top_level_display_flags -- --nocapture
# passed: 11 passed; 0 failed; 0 ignored; finished in 0.96s

git diff --check
# passed
```

After adding the changelog note and rebasing onto current `main`, I re-ran:

```text
cargo fmt --check && git diff --check origin/main..HEAD
# passed
```

A fresh rebuild of the same targeted test after `cargo clean` was attempted, but this local runner timed out after 10 minutes while recompiling the Rust workspace from scratch; it had previously hit a local disk-full/linker bus-error condition before I cleaned `target/`. No generated files were changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Show the `--continue` alias in `kin resolve --help`.
 
+### Fixed
+
+- Let top-level `kin --help`, `kin -h`, `kin --version`, and `kin -V` display without a nested-repository parent-store warning.
+
 ## [0.7.16] - 2026-09-12
 
 ### Changed
--- a/crates/kin-cli/src/main.rs
+++ b/crates/kin-cli/src/main.rs
@@ -2930,6 +2930,17 @@ fn parse_cli_or_report_retired_command() -> Cli {
     }
 }
 
+fn is_top_level_display_flag() -> bool {
+    let mut args = std::env::args_os().skip(1);
+    let Some(flag) = args.next() else {
+        return false;
+    };
+    if args.next().is_some() {
+        return false;
+    }
+    matches!(flag.to_str(), Some("--help" | "-h" | "--version" | "-V"))
+}
+
 /// The process entry. The outcome of [`run`] becomes an exit status in one
 /// place, so a refusal is reported the same way whether or not anyone still
 /// reads stderr, and the `println!` and `eprintln!` this file imports from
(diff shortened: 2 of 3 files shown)


troyjr4103:
Thanks for this fix, Emre. It shipped in 0.7.21, and the commit on main carries your name. Appreciate you taking it on.

PR #1793, opened 2026-09-16
Update kin source

troyjr4103:
Update the public source from the maintained development tree.

Validation: public metadata and full-history secret scans passed, along with the component checks in an isolated public checkout. This PR requires the repository's normal review and checks before merging.


troyjr4103:
Superseded: the Windows cross-check caught a Unix-only gate missing in the daemon escalation; fixed on private main (kin-internal #92) and re-exported as a new pull request from that main.

PR #1795, opened 2026-09-16
Update kin source

troyjr4103:
Update the public source from the maintained development tree.

Validation: public metadata and full-history secret scans passed, along with the component checks in an isolated public checkout. This PR requires the repository's normal review and checks before merging.


troyjr4103:
Superseded by the next export, which also carries the changelog fold so the 0.7.20 release notes name the setup wizard, the updater notice and the belt default.

